PR #95, opened 2022-07-31
Move spilling register to stack

lialan:
Since we have a very low-cost stack space, we could lower the general register pressure by allocate some virtual registers to stack slots and use stack addressing mode to access those stack slots. 

A new pass is introduced pre-RA to pre-emptively convert register addressing instructions to stack addressing instructions.


akiramenai (commented):
Thanks for the update.
Let's remove the limitations gradually. And please add tests including ones that test corner cases for each step. At the moment we only check that the transformation works in principle. Let's check that our heuristics works too.

akiramenai (changes requested):
Will check tests after the refactoring

akiramenai:
Should be covered by https://linear.app/matterlabs/issue/CPR-1062/improve-addressing-mode-selection-in-post-ra

PR #106, opened 2022-08-30
[Optimization] Re-enable pre-RA peephole opts

lialan:
This patch revamps the existing peephole pass:
* Turn on previously disabled peephole optimization pass.
* rename it to pre-RA pass so we can have post-RA pass which works on different phases.
* adds a new combination pattern to emit Stack-Reg->Stack instructions.
* combine instructions which work like MOVcrs and MOVsrs because ISEL does not select those instructions (a potential bug in ISEL).


akiramenai:
Will be addressed in https://linear.app/matterlabs/issue/CPR-1062/improve-addressing-mode-selection-in-post-ra

PR #109, opened 2022-09-05
[WIP] ISEL: Use complex pattern for stack/code operand

lialan:
`load_code` and `load_stack` are not working for some reason. And it turns out using complex patterns can work around such issues. 

This should enable stack-operand addressing mode pattern matching in the ISEL phase. Previously we were not able to emit such instructions.


akiramenai:
Will address in https://linear.app/matterlabs/issue/CPR-1062/improve-addressing-mode-selection-in-post-ra

PR #233, opened 2023-05-23
[EVM] Adding integer instructions.

PavelKopyl:
Adding integer instructions.
- Arithmetic
- Logical
- Bitwise
- Shift
- Comparison

TODO: Handle following corner cases of EVM architecture
     - (a / 0)
     - (a % 0)


zksync-admin-bot2:
Benchmark results:
```

```

lialan:
seeing this in the CI test results:
`error: unable to get target for 'evm', see --version and --triple.`
Better fix it?

PavelKopyl:
Yep. It seems we need to fix the CI a bit, otherwise reviewers should trust me, or build and perform ninja check-llvm by themselves...)

llvm/test/CodeGen/EVM/intrinsic.ll
```diff
@@ -0,0 +1,37 @@
+; RUN: llc --mtriple=evm < %s | FileCheck %s
+
+define i256 @addmod(i256 %rs1, i256 %rs2, i256 %rs3) nounwind {
+; CHECK-LABEL: @addmod
+; CHECK: ARGUMENT [[IN3:\$[0-9]+]], 2
+; CHECK: ARGUMENT [[IN2:\$[0-9]+]], 1
+; CHECK: ARGUMENT [[IN1:\$[0-9]+]], 0
+; CHECK: ADDMOD [[TMP:\$[0-9]+]], [[IN1]], [[IN2]], [[IN3]]
+
+  %res = call i256 @llvm.evm.addmod(i256 %rs1, i256 %rs2, i256 %rs3)
```
akiramenai, line 10: We should probably create a ticket to optimise `(rem (add nuw x, y), z)` to `addmod`, same for `mulmod`.

llvm/test/CodeGen/EVM/mul.ll
```diff
@@ -2,10 +2,22 @@
...
 
   %res = mul i256 %rs1, %rs2
   ret i256 %res
 }
+
+define i256 @mulrri(i256 %rs1) nounwind {
+; CHECK-LABEL: @mulrri
+; CHECK: CONST_I256 [[REG1:\$[0-9]+]], 500000
+; CHECK: CONST_I256 [[REG2:\$[0-9]+]], 0
+; CHECK: SUB [[REG3:\$[0-9]+]], [[REG2]], [[REG1]]
```
akiramenai, line 17: It looks like CONST can hold imm256, can't we materialize `-500000` directly?

llvm/test/CodeGen/EVM/sext.ll
```diff
@@ -0,0 +1,46 @@
...
+}
+
+define i256 @sexti128(i128 %rs1) nounwind {
+; CHECK-LABEL: @sexti128
+; CHECK: ARGUMENT [[IN1:\$[0-9]+]], 0
+; CHECK: SIGNEXTEND [[TMP:\$[0-9]+]], 15, [[IN1]]
+
+  %res = sext i128 %rs1 to i256
+  ret i256 %res
+}
```
akiramenai, line 46: What would happen for more exotic types? Are they accessible from Solidity?
P.S. If to handle these types properly we need to add some non-trivial logic to LLVM legalizer, it's better to create follow-up ticket and postpone.

llvm/lib/Target/EVM/EVMAsmPrinter.cpp
```diff
@@ -62,7 +62,7 @@ void EVMAsmPrinter::emitFunctionEntryLabel() {
 }
 
 void EVMAsmPrinter::emitInstruction(const MachineInstr *MI) {
-  EVMMCInstLower MCInstLowering(VRegMapping, MF->getRegInfo());
+  EVMMCInstLower MCInstLowering(OutContext, VRegMapping, MF->getRegInfo());
```
akiramenai, line 65: Could you explain why is it of need? It doesn't seem to be connected with the rest of the patch.

llvm/lib/Target/EVM/EVMInstrInfo.td
```diff
@@ -31,35 +34,170 @@ def EVMreturn   : SDNode<"EVMISD::RETURN", SDT_EVMReturn,
...
+def : Pat<(sext_inreg GPR:$src, i16), (SIGNEXTEND 1, GPR:$src)>;
+def : Pat<(sext_inreg GPR:$src, i32), (SIGNEXTEND 3, GPR:$src)>;
+def : Pat<(sext_inreg GPR:$src, i64), (SIGNEXTEND 7, GPR:$src)>;
+def : Pat<(sext_inreg GPR:$src, i128), (SIGNEXTEND 15, GPR:$src)>;
+
+// Reverse some operations with negative immediate operands to reduce bit-size
+// of the immediate encoding.
+// TODO: It needs to be discussed starting which bit-size this transformation
+// is deems profitable.
+def : Pat<(neg_imm128:$imm),
```
akiramenai, line 132: EVM has `PUSH32` instruction, doesn't it. So these complex materialisation doesn't look good.

akiramenai (changes requested):
Logic around materialising of big constants seems detrimental because `PUSH32` could materialize i256 if I get it right.

PavelKopyl:
> Logic around materialising of big constants seems detrimental because `PUSH32` could materialize i256 if I get it right.

Thank you for reviewing this!
Could you, please clarify a bit what is exactly you mean? The way integers are materialized via CONST_I256 instruction? Anyway, CONST_I256 will be converted to the appropriate PUSH instruction while stackification, depending on the actually size of the immediate.

llvm/test/CodeGen/EVM/intrinsic.ll
PavelKopyl: OK, I'll create the ticket, or add patterns for this.

llvm/test/CodeGen/EVM/mul.ll
PavelKopyl: Indeed, I'd like to discuss materialization of negative immediates.

Yes, CONST_I256 can hold -500000, but that would require all the 32 bytes to encode it and most of them will be non-zero bytes.  As I understand (after reading EVM documentation and other articles), all these 32 bytes need will be a part of the code for which we need to pay "Wei". On the other hand, the following transformation should be beneficial at the point of "Wei":
 
 ```
CONST_I256 (-500000) // to be transformed to: PUSH32 (1 byte) + 32 bytes of the immediate 
  - > 
 CONST_i256 0 // to be transformed to: PUSH0(1 byte)
 CONST_i256 500000 // to be transformed to: PUSH3 (1 byte) + 3 bytes of the immediate
 SUB // to be transformed to: SUB (1 byte)
```
 
 Also there is a difference in Wei price for zero byte and non-zero.
 There are some calculations about this here: https://medium.com/@hayeah/how-to-decipher-a-smart-contract-method-call-8ee980311603 (though it's not official documentation)
 
 I'm not 100% sure in that at the moment, so we could drop this complication and come back to this question later.

File: llvm/test/CodeGen/EVM/sext.ll (hunk 3 above)
Comment (PavelKopyl) [reply llvm/test/CodeGen/EVM/sext.ll:46]: I guess this should work with all the integer types from i8 to i256. The only exotic type I have in mind - is i160 that corresponds to the 'address' of the contract (if I understand it right). But that depends on our Solidity FE, will it generate i160 of expand it to i256?

File: llvm/lib/Target/EVM/EVMInstrInfo.td (hunk 5 above)
Comment (PavelKopyl) [reply llvm/lib/Target/EVM/EVMInstrInfo.td:132]: Please, check above my explanation of the negative integers materialization. This is the same topic.

File: llvm/lib/Target/EVM/EVMAsmPrinter.cpp (hunk 4 above)
Comment (PavelKopyl) [reply llvm/lib/Target/EVM/EVMAsmPrinter.cpp:65]: I added `MCContext &Ctx` field to VMMCInstLower class to be able to print big (> 64bits) immediates:

```
       auto Str = new (Ctx) SmallString<80>();
        CImmVal.toStringUnsigned(*Str);
        MCOp = MCOperand::createExpr(EVMCImmMCExpr::create(*Str, Ctx));
```

So, I'm just passing it to the  EVMMCInstLower ctor.

File: llvm/test/CodeGen/EVM/sext.ll (hunk 3 above)
Comment (hedgar2017) [reply llvm/test/CodeGen/EVM/sext.ll:46]: For now there is no way to pass type information from Solidity, so all cells are i256 by default.

File: llvm/test/CodeGen/EVM/mul.ll (hunk 2 above)
Comment (akiramenai) [reply llvm/test/CodeGen/EVM/mul.ll:17]: Ok, LGTM then

File: llvm/test/CodeGen/EVM/sext.ll (hunk 3 above)
Comment (akiramenai) [reply llvm/test/CodeGen/EVM/sext.ll:46]: Can't we just write a single test rather than guess and check for codegen? Use any iN with N divisible to 8 but not power of 2. If codegen is ok, add CHECK if it's not ticket + todo to not to forget when/if we start to care.

**Review by akiramenai — APPROVED** [review]
LGTM but a single test case to be added.

File: llvm/test/CodeGen/EVM/sext.ll (hunk 3 above)
Comment (PavelKopyl) [reply llvm/test/CodeGen/EVM/sext.ll:46]: My initial intention was to check generation of this instruction for each integer **machine** type available on ISelDAG (MVT:: i8, i16, i32, i64, i128, /*i256, i5128/).  The thing is that the pattern for generation SIGNEXTEND will not work for other types. The following code

`%res = sext i40 %rs1 to i256`

get `expanded` into

```
	SHL $3, %rs1, 216
	SAR $4, $3, 216
```

because i40 is extended EVT (EVT.isExtended() == true). It is hardcoded in getOperationAction() to exactly expand such nodes. Please, note the generated code is valid, but is not optimal. 

The only way to enable generation of SIGNEXTEND with such extended types it to add them to Machine types, the same way as you did it for i256. For me this seems a bit overkill, as FE shouldn't generate these exotic types.
Anyway we can add these types to MVT Ince required.

I can add one more test to this file demonstrating 'sexp' also gets lowered for exotic types, but non-optimally. 
Is it OK?

PR #241, opened 2023-06-05
[CI] Build compiler binaries to use as processes

hedgar2017:
Builds *zksolc* and *zkvyper* before running the tester to run them as processes rather than threads.

### DIFF
--- a/.github/workflows/compiler-integration-tests.yml
+++ b/.github/workflows/compiler-integration-tests.yml
@@ -58,14 +58,14 @@ jobs:
         run: |
           echo "url = \"https://github.com/matter-labs/compiler-llvm\"" >> LLVM.lock
           echo "branch = \"${GITHUB_HEAD_REF}\"" >> LLVM.lock
-          mv LLVM.lock ./compiler-tester/LLVM.lock
+          mv -fv './LLVM.lock' './compiler-tester/LLVM.lock'
 
       - name: Preparing LLVM.lock (workflow dispatch)
         if: github.event_name == 'workflow_dispatch'
         run: |
-          echo "url = \"https://github.com/matter-labs/compiler-llvm\"" >> LLVM.lock
-          echo "branch = \"${{ env.COMPILER_LLVM_BRANCH_NAME }}\"" >> LLVM.lock
-          mv LLVM.lock ./compiler-tester/LLVM.lock
+          echo "url = \"https://github.com/matter-labs/compiler-llvm\"" >> './LLVM.lock'
+          echo "branch = \"${{ env.COMPILER_LLVM_BRANCH_NAME }}\"" >> './LLVM.lock'
+          mv -fv './LLVM.lock' './compiler-tester/LLVM.lock'
 
       - name: Installing the LLVM manager and downloading the LLVM framework
         working-directory: compiler-tester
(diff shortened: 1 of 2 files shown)


.github/workflows/compiler-integration-tests.yml
```diff
@@ -149,13 +149,22 @@ jobs:
...
       - name: Building and running the compiler tester
         id: compiler_tester_run
         working-directory: compiler-tester
-        run: |
-          cargo run --release --bin compiler-tester -- --mode=${{ env.COMPILER_TESTER_MODE }}
+        run: |          
+          export RUST_BACKTRACE='full'
+          export LLVM_SYS_150_PREFIX="$(pwd)/target-llvm/target-final/"
+          cargo build --verbose --release --bin 'compiler-tester'
+          cargo build --verbose --release --manifest-path /usr/local/cargo/git/checkouts/compiler-solidity-*/*/Cargo.toml --target-dir './target-zksolc/'
```
FlashSheridan, line 161: What’s the expected behavior if the wildcards have multiple matches?  Presumably unlikely with the current setup (otherwise [jobs/9337227730](https://github.com/matter-labs/compiler-llvm/actions/runs/5181558700/jobs/9337227730) would have failed already with “error: unexpected argument”), but it might make the build fragile even if it doesn’t build the wrong thing.

.github/workflows/compiler-integration-tests.yml
```diff
@@ -58,14 +58,14 @@ jobs:
         run: |
           echo "url = \"https://github.com/matter-labs/compiler-llvm\"" >> LLVM.lock
           echo "branch = \"${GITHUB_HEAD_REF}\"" >> LLVM.lock
-          mv LLVM.lock ./compiler-tester/LLVM.lock
+          mv --verbose --force './LLVM.lock' './compiler-tester/LLVM.lock'
```
FlashSheridan: `-v` and `-f` seem to be more portable than `--verbose` and `--force`.

.github/workflows/compiler-integration-tests.yml
hedgar2017: On a clean system there is always one copy of each compiler in the registry.
There is no other way unfortunately. If we don't leverage cargo to download the deps for us, we'll have to parse Cargo.toml to get the branch, clone it and only then build.

.github/workflows/compiler-integration-tests.yml
hedgar2017: Are there envs where mv has different flags? I just hate contractions.

FlashSheridan: > Are there envs where mv has different flags? I just hate contractions.

I agree, but MacOS 12 has only the short versions, and warns:

> The -h, -n, and -v options are non-standard and their use in scripts is not recommended.

(I haven’t checked Linux variants DevOps might want to use.)

hedgar2017: Meh, ok, np.

.github/workflows/compiler-integration-tests.yml
FlashSheridan: Fingers crossed; a tiny bit of experimentation suggests that any addition would cause an error rather than building the wrong thing.  (I don’t claim to grok where all 20 hits for those wildcards on my decidedly unclean Mac came from.)

hedgar2017: Those are just different commits pulled every time you update the deps. So it gets really heavy sometimes and it makes sense to nuke it after each update of Rust 😅

zksync-admin-bot2:
Benchmark results:
```

```

lialan (approved):

PR #242, opened 2023-06-07
Instruction mappings for addr modes and argument swap

akiramenai:
The patch implements mapping to access instruction families with different input and output addressing modes, and swap argument flags. It also add missing fat pointer patterns to make mappings complete.

### DIFF
--- a/llvm/lib/Target/SyncVM/SyncVMISelLowering.cpp
+++ b/llvm/lib/Target/SyncVM/SyncVMISelLowering.cpp
@@ -5,6 +5,8 @@
 //===----------------------------------------------------------------------===//
 
 #include "SyncVMISelLowering.h"
+
+#include <unordered_map>
 #include "SyncVM.h"
 #include "SyncVMMachineFunctionInfo.h"
 #include "SyncVMSubtarget.h"
--- a/llvm/lib/Target/SyncVM/SyncVMInstrFormats.td
+++ b/llvm/lib/Target/SyncVM/SyncVMInstrFormats.td
@@ -12,7 +12,9 @@
 class PredRel;
 class PseudoRel; // map: `_p` -> `_s`
 class FlagRel; // map: `_s` -> `_v`
-class AddrModeRel; // generic map of addressing mode
+class AddrModeRel; // generic map of addressing mode of operands
+class RetAddrModeRel; // generic map of addressing mode of results
+class SwapRel; // map in0, in1 swapping operand instruction <-> non swapping one
 
 // The addressing mode can be categorized into operand addressing and destination addressing
 // OperandAddrMode:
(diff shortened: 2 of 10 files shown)


zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.001 ║
║ Best                                4.444 ║
║ Mean                               -0.001 ║
║ Worst                              -3.333 ║
║ Total                              -0.000 ║

```

llvm/lib/Target/SyncVM/SyncVMISelLowering.cpp
```diff
@@ -5,6 +5,8 @@
 //===----------------------------------------------------------------------===//
 
 #include "SyncVMISelLowering.h"
+
+#include <unordered_map>
```
lialan, line 9: @FlashSheridan here

llvm/lib/Target/SyncVM/SyncVMInstrFormats.td
```diff
@@ -236,6 +302,8 @@ class IBinary<bits<8> opcode,
   bit IsPseudoArith = 0;
   bits<3> OperandAddrMode = OpndAddrNotSet.Value;
```
lialan, line 303: Now we can remove `OperandAddrMode` and `DestAddrMode`. `bits` is not as good as `int` in TableGen I think.

llvm/lib/Target/SyncVM/SyncVMInstrFormats.td
```diff
@@ -68,6 +70,70 @@ def getNonFlagSettingOpcode : InstrMapping {
   let ValueCols = [["1"]];
 }
 
+def mapRRInputTo : InstrMapping {
+  let FilterClass = "AddrModeRel";
+  let RowFields = ["BaseOpcode", "DestAddrMode", "FlagSetting"];
+  let ColFields = ["OperandAM"];
+  let KeyCol = ["0"];
+  let ValueCols = [["0"], ["1"], ["2"], ["3"]];
```
lialan, line 78: Maybe some comments about what 0, 1, 2, 3 represents.

or you can try the way `OperandAddrMode` organizes, which uses names instead of integer values for easier to understand.

lialan (changes requested):
Generally LGTM, just a little bit more details.

llvm/lib/Target/SyncVM/SyncVMInstrFormats.td
akiramenai: We might need bits for encoding, don't we? And there is possibility we start debugger work soon which might require binary output.
So, let me create a ticket for this duplication.

akiramenai: https://linear.app/matterlabs/issue/CPR-1154/remove-duplication-between-operandaddrmode-and-operandam-in-td-files

llvm/lib/Target/SyncVM/SyncVMInstrFormats.td
akiramenai: I've actually tried twice. I previously reported that bit vectors aren't good for keys (and indeed I used to produce semantically incorrect `.inc` files), but it seems that Hexagon managed to use them. Now I failed to produce tables for mappings, but otherwise `.inc` file is correct.
So with this particular comment (though I agree magic numbers are evil), I'd prefer to postpone and track as https://linear.app/matterlabs/issue/CPR-1154/remove-duplication-between-operandaddrmode-and-operandam-in-td-files. It seems that we don't need `OperandAM` at all. This patch as it is makes C++ code more concise, while our `td`s are messy anyway. And I'd wait for coq spec to be ready before to clean up `td`s. Even if we aren't going to generate `td`s from the spec, I'd use the exactly same terminology in `td`s as it is in the spec.
@lialan @hedgar2017 @sayon what do you think?

lialan: @akiramenai I tried to use `string` instead of int, and it works in my case. So maybe just change int to string and use something like `rr`, `sr` to represent that.

But I don't object we do that in the ticket.

lialan (approved):
Since we know what to do with the issues we discussed here, it is LGTM now.

lialan:
@akiramenai A friendly reminder that we need to turn on commit signatures. But it seems like it is not enforced yet.

akiramenai:
@lialan thanks. Will learn it and start using with the next patch.

PR #270, opened 2023-07-31
[LLDB] Fix handling of 256-bit types

vladimirradosavljevic:


zksync-admin-bot2:
Benchmark results:
```

```

clang/lib/AST/ASTImporter.cpp
```diff
@@ -1077,6 +1080,13 @@ ExpectedType ASTNodeImporter::VisitAtomicType(const AtomicType *T){
   return Importer.getToContext().getAtomicType(*UnderlyingTypeOrErr);
 }
 
+// SyncVM local begin
+ExpectedType ASTNodeImporter::VisitBitIntType(const BitIntType *T) {
```
lialan, line 1084: @vladimirradosavljevic Curious about this: is this needed to generate bit int dwarf info into the object file so that LLDB can read such info into its type ssytem?

vladimirradosavljevic: This is actually needed to do Expression evaluation (e.g. p var=5). LLDB is using clang to generate LLVM IR that will be ran so it can evaluate expressions. This won't help us, as our source will come from Solidity or Vyper, so we would presumably need to implement our own expression evaluation engine to support it. I fixed this, because I wanted to get more familiar with how LLDB works.

akiramenai (approved):
I wonder are we going to stick with clang type system plugin or do we plan to have our own?

vladimirradosavljevic:
> I wonder are we going to stick with clang type system plugin or do we plan to have our own?

If it can handle 256-bit pointers, I would say we should stick with it. I will try to hack around WASM to check how 256-bit pointers are supported.

PR #281, opened 2023-08-17
[WIP] [SyncVM] Fold select instructions into selectee

lialan:
[CPR-1129](https://linear.app/matterlabs/issue/CPR-1129/fold-op-selectcc-x-y-z-into-op-y-z-opcc-x-z)

A machine pass to:
* detect select instructions, and a folding operand of it
* reschedule the folding operand to after flag setter instruction
* fold the select with the operand
* some test files


akiramenai:
@lialan Do we plan to finish with this work?

lialan:
@akiramenai I plan to do it after current tasks. 

Recently I figured out we should: 1. add `isSelect` to select instructions and 2. use: `TII::optimizeSelect` to do an SSA-level folding. I am planning to re-do it in this patch.

PR #283, opened 2023-08-22
Add implementation of '__(div|sdiv|mod|smod)' functions to the stdlib.

PavelKopyl:
1. Add a range check for the first param of '__signextend' stdlib function.
    Also fixied the constant folding implementation for the case when the first param is out of range.
    Added more LIT tests.

2. Add implementation of '__(div|sdiv|mod|smod)' functions to the stdlib.
    Also added constant folding support for these functions.


zksync-admin-bot2:
Benchmark results:
```
║ Best                                2.410 ║
║ Best                                5.769 ║
║ Best                                2.817 ║

```

hedgar2017 (approved):
The CI passed!
Please merge and rebase keccak256 on top.

llvm/test/CodeGen/SyncVM/constant-folding-earlycse.ll
```diff
@@ -107,7 +111,7 @@ define i256 @test_signextend2() {
 
 define i256 @test_signextend3() {
 ; CHECK-LABEL: @test_signextend3
-; CHECK-NEXT: ret i256 poison
+; CHECK-NEXT: ret i256 -1
```
akiramenai: Sorry, I might have missed, but why are we changing semantics here?

akiramenai (commented):
Are tests copied from somewhere? Should I review them all or just poison, undef and things like that?

llvm/test/CodeGen/SyncVM/constant-folding-earlycse.ll
hedgar2017: This looks strange. `undef` is never generated by FE, so I'd crash on validation here.

PavelKopyl:
> Are tests copied from somewhere? Should I review them all or just poison, undef and things like that?

Yep, I took them form https://github.com/matter-labs/compiler-tests/tree/main/solidity/simple/yul_instructions/{div,sdiv,mod,smod}.yul, but I had to convert hex to signed decimal numbers. You can check undef + some obvious cases (where it's clear what is the answer).

llvm/test/CodeGen/SyncVM/constant-folding-earlycse.ll
PavelKopyl: I decided to change the semantics here after getting acquainted with the code FE emits for '__signextend'.  The code checks if the first parameter (size in bytes minus one of the value being extended) is greater than 30. If so, we just return the value itself. I added this check to the '__signextend' implementation in stdlib (before the path, this check was generated by FE ), As a result there is no need to generate poison here. In case of 'Undef' value of the first param, I refine it to 0 (we can refine it to any other value), so the operations transforms to __signextend(0, value).
@hedgar2017 I understand you concern regarding usage of 'Undef' value. Here I just wanted to implement constant folding in general case (the same way it was done for other functions). 'Undef' value may also appear as a result of transformations, so this is not the place where we can catch all the 'Undef' values. I suppose we need to solve this task (detect usage of 'Undef' values) globally.

akiramenai: @hedgar2017 LLVM could produce undef/poison even if FE doesn't. We've already encountered in the past undef values introduced by the passes which didn't make us any trouble. So we can only guard it in the beginning of the pipeline to assert that FE indeed don't produce any.
We might also detect undefs later, but it's likely to report us a lot of false positives. Nothing is wrong with undef in LLVM unless it turns into the UB (I mean nothing is wrong with UB in LLVM, but it a clear sign that either we have a FE bug, or we a have a bug in the passes). We can also detect some UBs, but if it ins't done in upstream yet, we should first learn why: either nobody cares to catch it that later as any potential diagnostic is bad, and there is a UB santizer which does the work better in FE; or for some reason it's not that easy to implement.

akiramenai (approved):
I checked for undef cases. To be honest, I'm not an expert in the spirit of undef/poison propagation. But as you said we can replace undef with a value we prefer and I don't feel it to be an important choice we can't redo in future (please let me know if I'm wrong on that), so I think it's ok to go with 0s.

llvm/test/CodeGen/SyncVM/constant-folding-earlycse.ll
akiramenai: @hedgar2017 and btw don't we put undef where we don't use all the 12 or 13 arguments of a far call?

hedgar2017: We do, but they are just stubs to avoid rewriting registers IIRC.
For EVM instructions undef doesn't make sense because we don't know what to do with it in terms of equivalence. However, solc's constant folding is even more messy with its roundings.

hedgar2017:
> I checked for undef cases. To be honest, I'm not an expert in the spirit of undef/poison propagation. But as you said we can replace undef with a value we prefer and I don't feel it to be an important choice we can't redo in future (please let me know if I'm wrong on that), so I think it's ok to go with 0s.

It doesn't sound good to me.
1. Why exactly 0s? Can we fold to undef or poison instead?
2. What is 'a value we prefer'? Would folding undef input into a value conceal a bug in the pipeline?

akiramenai:
@hedgar2017 
That's the discussion I tried to avoid because we are going to spend time without changing much.
`2.` Any decision will conceal a bug in the pipeline as much as `a / 0 = 0` conceals a bug in a contract. `a / 0 = 0` only makes a bug reliable, i.e. it's going to reproduce with new versions of compiler exactly the same way, while the worst scenario with UB the behaviour could change one day (and a reminder for those who aren't a C++ programmer anymore: "it works" is just a special case of UB :)).
Again, undef is a legit value and by itself nothing is wrong with undef. In the particular case I referred above it means that you can put to `r3 - r12` whatever and in particular keep the value as it is (so we don't need to change registers' values to make a call). So it's an optimization. Any computations that use undef follow the rules from the spec and again nothing is wrong if the result is not used. I expect LLVM define the semantic here to allow more optimizations to be done.
To really make the pipeline more reliable we can do Alive2 (as @FlashSheridan proposed) or to account for a particular issue with `undef` we can do a pass. The issue with the latter, that it only makes sense if we expect much more density of bugs with `undef`s in LLVM pipeline than anywhere else, otherwise we put effort but the order of reliability remains the same.
As a summary, whatever we decide here, if the change passes our CI, it's good enough and we can't do noticeably better w.o. committing significant effort into it. I don't think the effort should be in scope of this particular work.
`1.` If we want to be pedantic about it , we might replicate the exact value corresponding ll optimization gives, but I expect it to be UB as there could be branching on undef values. Putting undef / poison will make more room for optimizations, putting 0 stops an issue to be propagated if there is an issue. But I'm not sure if I have a preference between Solidity's flavour `undef / 5 = 0` vs replacing all my method's body with `return 42;` in case of ill-formed IR / optimization bug.

hedgar2017:
> @hedgar2017 That's the discussion I tried to avoid because we are going to spend time without changing much. `2.` Any decision will conceal a bug in the pipeline as much as `a / 0 = 0` conceals a bug in a contract. `a / 0 = 0` only makes a bug reliable, i.e. it's going to reproduce with new versions of compiler exactly the same way, while the worst scenario with UB the behaviour could change one day (and a reminder for those who aren't a C++ programmer anymore: "it works" is just a special case of UB :)). Again, undef is a legit value and by itself nothing is wrong with undef. In the particular case I referred above it means that you can put to `r3 - r12` whatever and in particular keep the value as it is (so we don't need to change registers' values to make a call). So it's an optimization. Any computations that use undef follow the rules from the spec and again nothing is wrong if the result is not used. I expect LLVM define the semantic here to allow more optimizations to be done. To really make the pipeline more reliable we can do Alive2 (as @FlashSheridan proposed) or to account for a particular issue with `undef` we can do a pass. The issue with the latter, that it only makes sense if we expect much more density of bugs with `undef`s in LLVM pipeline than anywhere else, otherwise we put effort but the order of reliability remains the same. As a summary, whatever we decide here, if the change passes our CI, it's good enough and we can't do noticeably better w.o. committing significant effort into it. I don't think the effort should be in scope of this particular work. `1.` If we want to be pedantic about it , we might replicate the exact value corresponding ll optimization gives, but I expect it to be UB as there could be branching on undef values. Putting undef / poison will make more room for optimizations, putting 0 stops an issue to be propagated if there is an issue. But I'm not sure if I have a preference between Solidity's flavour `undef / 5 = 0` vs replacing all my method's body with `return 42;` in case of ill-formed IR / optimization bug.

Thanks for the explanation!

1. To me `undef / 5 = 0` isn't much better than `undef / 5 = 42`, but it's hard to reason if we don't know where undefs can be coming from. It looks like with `undef / 5 = undef` we aren't making any unjustified assumptions, and if I'm getting this right it also allows for more optimizations. Can we stick to this, especially in `__signextend`?

2. `a / 0 = 0` is a normal behavior in Solidity even if we don't like it. E.g. someone can be checking if divider is zero this way :) There is no place for UB.
Anyway, if it passes our CI, it's probably fine. How can we check if those undefs and poisons are ever produced? print-after-all?

PavelKopyl:
>It looks like with undef / 5 = undef we aren't making any unjustified assumptions, and if I'm getting this right it also allows for more optimizations. Can we stick to this, especially in __signextend?
Sure, we can do this. Taking into account that constant folding happens at early stages of the pipeline, 'Undef' value anyway will likely to be replaced with zero by InstrCombining pass. So, at the point of catching bugs it will not help significantly. Anyway, I'll make this change.

akiramenai:
@akiramenai 

`2`. I don't know if 0 is similar to undef, but I trust you if you say so. Anyway let's not follow deliberate choices of the EF team :)
`1`. Ok, I got the idea. To summarize, I would not assume `undef == 0` except for the cases where `undef` doesn't affect the result. So `undef / 0 == 0`, but `5 / undef == undef`. I hope you guys know what you're doing and you'll pick the safest approach.

Sorry for editing out your comment :) I didn't know I could do that, and my phone UI got bugged somehow.

llvm/test/CodeGen/SyncVM/constant-folding-earlycse.ll
```diff
@@ -107,7 +111,7 @@ define i256 @test_signextend2() {
...
 
   %res = call i256 @__signextend(i256 undef, i256 32767)
```
hedgar2017: So is this one also `undef` like below?

hedgar2017 (changes requested):

PavelKopyl:
@akiramenai If our goal here is to reveal potential bugs in the incoming LLVM IR, I'm fine with this and we can return undef/position to induce UB at the following pipeline stages.
At the same time it'd like to note the following:
1. I wanted (this seemed natural for me) to keep in sync as much as possible constant folding implementation with the stdlib one. That means I expected identical outputs on the LIT tests for ConstantFolding implementation and the stdlib one. To check stdlib, I disabled our ConstantFolding and forced inlining of stdlib functions which got ConstantFolded on llvm IR instructions level. If we start return Undef/Poision in ConstantFolding implementation we will diverged from stdlib implementation. Probably that's fine.
2. I think we should return only Poison value in case of Undef/Poision input values. That is because according to LLVM specification returning 'Undef' in all cases is not safe.We can return Undef only if all the output bits are affected by the undef bits. For example, this is the case for `add i256 Undef, %x,` but no the case for  `or %X, Undef`, see
[Undefined values](https://llvm.org/docs/LangRef.html#id1798).
As we do ConstantFolding for library calls, I guess, we can impose any semantic, so, for simplicity, let's return Poision for all the functions from the stdlib if at leas one input value is Undef or poison.

llvm/test/CodeGen/SyncVM/constant-folding-earlycse.ll
PavelKopyl: If we suppose that Undef/Poison shouldn't appear as input values, let's return Poision for all the stdlib functions. Poision is more stronger that Undef, will more likely to reveal LLVM IR bugs.

zksync-admin-bot2:
Benchmark results:
```
║ Best                                2.410 ║
║ Best                                5.769 ║
║ Best                                2.817 ║

```

hedgar2017 (approved):

zksync-admin-bot2:
Benchmark results:
```
║ Best                                2.410 ║
║ Best                                5.769 ║
║ Best                                2.817 ║

```

PR #305, opened 2023-09-26
Constant folding implementation for __sha3 stdlib function

PavelKopyl:


zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.701 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.161 ║
║ Worst                              -3.261 ║

║ Mean                                0.703 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.625 ║
║ Mean                                0.516 ║
║ Best                               40.602 ║
║ Worst                              -1.091 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.012 ║
║ Total                               0.007 ║

```

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.701 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.161 ║
║ Worst                              -3.261 ║

║ Mean                                0.703 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.625 ║
║ Mean                                0.516 ║
║ Best                               40.602 ║
║ Worst                              -1.091 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.012 ║
║ Total                               0.007 ║

```

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.701 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.161 ║
║ Worst                              -3.261 ║

║ Mean                                0.703 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.625 ║
║ Mean                                0.516 ║
║ Best                               40.602 ║
║ Worst                              -1.091 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.012 ║
║ Total                               0.007 ║

```

akiramenai:
Rebased

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.701 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.161 ║
║ Worst                              -3.261 ║

║ Mean                                0.703 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.625 ║
║ Mean                                0.516 ║
║ Best                               40.602 ║
║ Worst                              -1.091 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.012 ║
║ Total                               0.007 ║

```

PavelKopyl:
PR has two mostly independent commits:
1. Implementation of keccak256 hash function. This implementation was reviewed here: https://github.com/matter-labs/compiler-llvm/pull/275
2. __sha3 constant folding itself.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  uint64_t Size;
+  ObjectSizeOpts Opts;
+  Opts.NullIsUnknownSize = NullPointerIsDefined(F);
+
+  if (getObjectSize(V, Size, DL, &TLI, Opts))
+    return Size;
+  return MemoryLocation::UnknownSize;
+}
+
+struct MemClobber {
```
akiramenai, line 124: Readability. Documentation is of help.
The meaning of Value is guessable, but a reader still might feel he needs to double check and start looking the context. Perhaps renaming Value to Clobber also helps.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  uint64_t Start;
+  uint64_t Size;
+  APInt Value;
+
+  MemClobber(uint64_t Start, uint64_t Size, APInt Value)
```
akiramenai: POD types are list initializable.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  const LoopInfo &LI;
+
+  // Whether the function contains any irreducible control flow, useful for
+  // being accurately able to detect loops.
+  bool ContainsIrreducibleLoops;
+
+  // All the sha3 calls to be analyzed.
+  SmallVector<MemoryUse *, 16> SHA3MemUses;
+
+  void getUseClobbers(MemoryUse *MU, MemoryLocation Loc,
```
akiramenai: Readability: explicit interfaces. RVO/NRVO makes return by value preferable.

Undefined and unused. `collectUseClobbers` is used instead.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+struct MemClobber {
+  uint64_t Start;
+  uint64_t Size;
+  APInt Value;
+
+  MemClobber(uint64_t Start, uint64_t Size, APInt Value)
+      : Start(Start), Size(Size), Value(Value) {}
+};
+
+struct FoldingState {
```
akiramenai: class + fields encapsulation is preferred.
Also might worth to extract interfaces and move them up.
Documentation would also be of help to understand how the state works without going deep into implementation details

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+      : F(F), AA(AA), AC(AC), MSSA(MSSA),
+        MSSAUpdater(std::make_unique<MemorySSAUpdater>(&MSSA)), DT(DT),
+        TLI(TLI), DL(F.getParent()->getDataLayout()), SQ(DL, &TLI, &DT, &AC),
+        LI(LI) {
+    MSSA.ensureOptimizedUses();
+
+    const ReversePostOrderTraversal<BasicBlock *> RPOT(&*F.begin());
+    for (BasicBlock *BB : RPOT) {
+      for (Instruction &I : *BB) {
+        CallInst *Call = dyn_cast<CallInst>(&I);
```
akiramenai: Readability. Might worth to wrap all checks into a function that checks if the instruction is a call to LibFunc_xvm_sha3. Optionally also use a filtering iterator as it expresses intent better.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+                          SmallVectorImpl<MemoryDef *> &Clobbers) {
+    MemorySSAWalker *Walker = MSSA.getWalker();
+    SmallVector<MemoryAccess *> WorkList{Walker->getClobberingMemoryAccess(MU)};
+    SmallSet<MemoryAccess *, 8> Visited;
+
+    // Start with a nearest dominating clobbering access, it will be either
+    // live on entry (nothing to do, load is not clobbered), MemoryDef, or
+    // MemoryPhi if several MemoryDefs can define this memory state. In that
+    // case stop the search.
+    // FIXME: we can relax this constraint in the future if it turns out we miss
```
akiramenai: TODO + ticket number please

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    // optimization opportunities.
+    while (!WorkList.empty()) {
+      MemoryAccess *MA = WorkList.pop_back_val();
+      if (!Visited.insert(MA).second)
+        continue;
+
+      if (MSSA.isLiveOnEntryDef(MA))
+        continue;
+
+      if (MemoryDef *Def = dyn_cast<MemoryDef>(MA)) {
```
akiramenai: `auto*` as type is in the RHS.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    // Collect whether there is any irreducible control flow in the function.
+    ContainsIrreducibleLoops = mayContainIrreducibleControl(F, &LI);
+  }
+
+  MemoryLocation getLocForSHA3Call(const CallInst *I) const {
+    return MemoryLocation::getForArgument(I, 0, TLI);
+  }
+
+  /// Collect all the potential clobbering \p Loc definitions
+  /// for a given \p Use.
```
akiramenai: the parameter is named MU.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  void collectUseClobbers(MemoryUse *MU, MemoryLocation Loc,
+                          SmallVectorImpl<MemoryDef *> &Clobbers) {
+    MemorySSAWalker *Walker = MSSA.getWalker();
+    SmallVector<MemoryAccess *> WorkList{Walker->getClobberingMemoryAccess(MU)};
+    SmallSet<MemoryAccess *, 8> Visited;
+
+    // Start with a nearest dominating clobbering access, it will be either
+    // live on entry (nothing to do, load is not clobbered), MemoryDef, or
```
akiramenai: Could you explain the logic with MemoryDef here (rephrasing is very welcome here as the sentence is too long) or when adding a value to the working list.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+      if (!Visited.insert(MA).second)
+        continue;
+
+      if (MSSA.isLiveOnEntryDef(MA))
+        continue;
+
+      if (MemoryDef *Def = dyn_cast<MemoryDef>(MA)) {
+        Clobbers.push_back(Def);
+        WorkList.push_back(
+            Walker->getClobberingMemoryAccess(Def->getDefiningAccess(), Loc));
```
akiramenai: Do we need a list at all? It looks like it always contain 1 element at most.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  /// Try to simplify instructions of the functions. Candidates for
+  /// simplification may appear after replacing __sha3 call with the
+  /// calculated hash value.
+  bool simplifyInstructions() {
+    bool Changed = false;
+    for (auto &Inst : make_early_inc_range(instructions(F))) {
+      if (Value *V = simplifyInstruction(&Inst, SQ)) {
+        LLVM_DEBUG(dbgs() << " SyncVMSHA3ConstFolding simplify: " << Inst
+                          << "  to: " << *V << '\n');
+        if (!DebugCounter::shouldExecute(SHA3Counter)) {
```
akiramenai: I apologise, I'm not familiar with debug counters, could you educate my why we skip here?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  }
+
+  /// Exclude the instruction from MSSA if it's going to be removed from a BB.
+  void removeFromMSSA(Instruction *Inst) {
+    if (VerifyMemorySSA)
+      MSSA.verifyMemorySSA();
+    MSSAUpdater->removeMemoryAccess(Inst, true);
+  }
+
+  /// Try to simplify instructions of the functions. Candidates for
```
akiramenai: you mean function F?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  }
+
+  /// Returns true if a dependency between \p Clobber and \p MemUse is
+  /// guaranteed to be loop invariant for the loops that they are in. Either
+  /// because they are known to be in the same block, in the same loop level or
+  /// by guaranteeing that \p MemUseLoc only references a single MemoryLocation
+  /// during execution of the containing function.
+  bool isGuaranteedLoopIndependent(const Instruction *MemUse,
+                                   const Instruction *Clobber,
+                                   const MemoryLocation &MemUseLoc) {
```
akiramenai: const?

Same for other methods where applicable.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    // If the dependency is within the same block or loop level (being careful
```
akiramenai: Could we spread the comment among the code. Otherwise you need first to put into the memory as a whole, and then check the logic. Also it looks like we don't check irreducible loop when in the same BB, etc. So more precise description would be better.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  /// loop. In particular, this guarantees that it only references a single
+  /// MemoryLocation during execution of the containing function.
+  bool isGuaranteedLoopInvariant(const Value *Ptr) {
+    Ptr = Ptr->stripPointerCasts();
+    if (auto *GEP = dyn_cast<GEPOperator>(Ptr))
+      if (GEP->hasAllConstantIndices())
+        Ptr = GEP->getPointerOperand()->stripPointerCasts();
+
+    if (auto *I = dyn_cast<Instruction>(Ptr))
+      return I->getParent()->isEntryBlock();
```
akiramenai: Could you explain why we only consider entry instruction as loop invariant?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    return true;
+  }
+
+  /// Return 'OW_Complete' if a store to the \p ClobberLoc location (by \p
```
akiramenai: Looks like a good documentation for the enumeration itself.

llvm/test/CodeGen/SyncVM/sha3-constant-folding.ll
```diff
@@ -0,0 +1,124 @@
...
+  %next_addr = getelementptr i256, ptr addrspace(1) %addr, i256 1
+  store i256 111, ptr addrspace(1) %addr2, align 1
+  store i256 56457598675863654, ptr addrspace(1) %next_addr, align 1
+  %hash1 = tail call fastcc i256 @__sha3(ptr addrspace(1) %addr, i256 64, i1 true)
+  %hash2 = tail call fastcc i256 @__sha3(ptr addrspace(1) %next_addr, i256 32, i1 true)
+  %hash = add i256 %hash1, %hash2
+  ret i256 %hash
+}
+
+; Memory locations of store instructions do alias with each other, so no
```
vladimirradosavljevic, line 106: As far as I can see, here we have partial aliasing, and our custom Alias Analysis should return that result. I'm not sure how frequently we have this pattern, but I'm wondering whether we can use that information to actually do folding as we should be able to recreate constant? If this makes sense to you, what do you think about creating TODO + ticket to do that in the future? Also, I assume we could use the same optimization and implement it in the InstCombine to remove some loads.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  /// Clobber instruction) is completely within the \p MemUseLoc
+  /// location (by \p MemUseLoc). Return 'OW_MaybePartial' if \p ClobberLoc
+  /// is partially with the \p MemUseLoc they both refer to the same underlying
+  /// object. Returns 'OR_None' if \p Clobber is known to not write to the
+  /// memory referring by \p MemUse. Returns 'OW_Unknown' if nothing can be
+  /// determined.
+  OverlapResult isCompleteOverlap(const Instruction *MemUse,
+                              const Instruction *Clobber,
```
akiramenai: please clang-format

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    const Value *MemUsePtr = MemUseLoc.Ptr->stripPointerCasts();
+    const Value *ClobberUndObj = getUnderlyingObject(ClobberPtr);
+    const Value *MemUseUndObj = getUnderlyingObject(MemUsePtr);
+    const uint64_t MemUseSize = MemUseLoc.Size.getValue();
+    const uint64_t ClobberSize = ClobberLoc.Size.getValue();
+
+    ClobberOff = MemUseOff = 0;
+
+    // Try to convert a pointer to the integer value.
+    auto getPointerAsInt = [](const Value *Ptr) -> std::optional<uint64_t> {
```
akiramenai: Please extract as a utility function. IMO, a named anonymous function is almost always the feature abuse.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+  /// location (by \p MemUseLoc). Return 'OW_MaybePartial' if \p ClobberLoc
+  /// is partially with the \p MemUseLoc they both refer to the same underlying
+  /// object. Returns 'OR_None' if \p Clobber is known to not write to the
+  /// memory referring by \p MemUse. Returns 'OW_Unknown' if nothing can be
+  /// determined.
+  OverlapResult isCompleteOverlap(const Instruction *MemUse,
+                              const Instruction *Clobber,
+                              const MemoryLocation &MemUseLoc,
+                              const MemoryLocation &ClobberLoc,
+                              int64_t &MemUseOff, int64_t &ClobberOff) {
```
akiramenai: Perhaps it make sense to make it a part of the return value.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
akiramenai: Also to the best of my memory @vladimirradosavljevic wrote something alike in the past. So perhaps can be unified to an utility function.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+      }
+      return std::nullopt;
+    };
+
+    // If both clobber and memuse pointers are constant we expect two cases:
+    //  - pointer is just a nullptr
+    //  - pointer is a cast of an integer constant
+    if (isa<Constant>(ClobberPtr) && isa<Constant>(MemUsePtr)) {
+      if (!getPointerAsInt(ClobberPtr) || !getPointerAsInt(MemUsePtr))
+        return OL_Unknown;
```
akiramenai: Why can it happen?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
akiramenai: Also `Offset` is preferred as `Off` is not a widely accepted shortcut for it.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+        return OL_Complete;
+    }
+
+    // Query the alias information
+    const AliasResult AAR = AA.alias(MemUseLoc, ClobberLoc);
+
+    // If the start pointers are the same, we just have to compare sizes to see
+    // if the memuse was larger than the clobber.
+    if (AAR == AliasResult::MustAlias) {
+      // Make sure that the MemUseSize size is >= the ClobberSize size.
```
akiramenai: What if not? Is it expected and just handled below?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    //  - pointer is a cast of an integer constant
+    if (isa<Constant>(ClobberPtr) && isa<Constant>(MemUsePtr)) {
+      if (!getPointerAsInt(ClobberPtr) || !getPointerAsInt(MemUsePtr))
+        return OL_Unknown;
+
+      const uint64_t ClobberPtrInt = *getPointerAsInt(ClobberPtr);
+      const uint64_t MemUsePtrInt = *getPointerAsInt(MemUsePtr);
+      if (MemUsePtrInt <= ClobberPtrInt &&
+          (ClobberPtrInt + ClobberSize) <= (MemUsePtrInt + MemUseSize)) {
+        MemUseOff = static_cast<uint64_t>(MemUsePtrInt);
```
akiramenai: int64_t?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+
+    ClobberOff = MemUseOff = 0;
+
+    // Try to convert a pointer to the integer value.
+    auto getPointerAsInt = [](const Value *Ptr) -> std::optional<uint64_t> {
+      if (auto *CE = dyn_cast<ConstantExpr>(Ptr)) {
+        if (CE->getOpcode() == Instruction::IntToPtr) {
+          const Value *C = CE->op_begin()->get();
+          assert(isa<ConstantInt>(C));
+          return cast<ConstantInt>(C)->getZExtValue();
```
akiramenai: We have i256 pointers. Shouldn't we have some guards in the pass?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+    // both start and end of the dead one is "inside" the killing one:
+    //    |<->|--clobber-|<->|
+    //    |------MemUse------|
+    // We have to be careful here as *Off is signed while *.Size is unsigned.
+
+    // Check if the clobber access starts "not before" the memuse one.
+    if (ClobberOff >= MemUseOff) {
+      // If the clobber access ends "not after" the memuse access then the
+      // clobber one is completely overlapped by the memuse one.
+      if (uint64_t(ClobberOff - MemUseOff) + ClobberSize <= MemUseSize) {
```
akiramenai: `static_cast` instead of C-style one.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+      return OL_MaybePartial;
+    }
+
+    // Can reach here only if accesses are known not to overlap.
+    return OL_None;
+  }
+
+  /// Sort memory clobbers is the ascending order of their starting addresses
+  /// Ensure they do not overlap with each other.
+  bool sortAndCheckMemoryClobbers(SmallVector<MemClobber> &MemClobbers,
```
akiramenai: Can we split sort and check as it violates single responsibility rule? Perhaps the check function can require input to be sorted.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+
+      ++ItPrev;
+    }
+    assert(TotalSize == SHA3MemSize);
+
+    return true;
+  }
+
+  /// Check if we can ignore non store clobber instruction that doesn't actually
+  /// clobber heap memory. For example, a memcpy to AS other than heap.
```
akiramenai: It's the only example. Have we missed a TODO here?

akiramenai: Also what if not an intrinsic but a mere load / store?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,715 @@
...
+      if (!ClobberVal) {
+        LLVM_DEBUG(dbgs() << "\t\tstored value isn't constant" << '\n');
+        return nullptr;
+      }
+
+      const uint64_t ClobberSize = ClobberLoc.Size.getValue();
+      // If we have already seen a clobber with the same start position
+      // and the size, we can ignore the current clobber, as it's killed
+      // by existing one.
+      // TODO: needs to check for flaws in this logic. In fact this is
```
akiramenai: +ticket

akiramenai (changes requested):
Good work, thank you, but require some cleanup.
I finished reading around `getKeccak256Hash` as I think 30 comments is enough for the first iteration. If some stylistic issues apply to it and the code after please apply.

llvm/test/CodeGen/SyncVM/sha3-constant-folding.ll
PavelKopyl: Yep, our AA detects this case, but I decided not to support this it for simplicity, because I haven't seen it in tests (thought I didn't check all of them).
@hedgar2017, probably you know more about __sha3. Can we have the following case, and how frequent it may be in a real life:


```
|---Store1---|
     |---Store2--|
                 |---------Store3----|
   |--------sha3 memory ------| 

```

?

Currently, this pass supports only the following case:

```
|---Store1---|
             |---Store2--|
                         |-Store3----|
|--------sha3 memory ----------------| 
```

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK. Added description and renamed Value -> Clobber.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK. I just wanted to use emplace_back, but let's use push_back.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yep, thank you!

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.701 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.161 ║
║ Worst                              -3.261 ║

║ Mean                                0.703 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.625 ║
║ Mean                                0.516 ║
║ Best                               40.602 ║
║ Worst                              -1.091 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.012 ║
║ Total                               0.007 ║

```

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK. I moved the check to a function.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yes, thank you.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Added their more detailed description, but anyway it's also worth to check MemorySSA documentation.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: You are right, this left from my previous experiments. Thank you, fixed.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: I answered to slack)

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yep, done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done. By the way, I was inspired by DeadStoreElimination pass, where I took this code.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yes, I added it also to the enum.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK, I extracted this code to a wrapper function. Yes, this code is almost identical to what Vladimir implemented in SyncVM AA, so it's reasonable to move this helper into a common code. I've created a ticked for this, CPR-1380.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Agreed, done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: You must be right, I added an assert for this

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yes, this case should be handled below.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yep.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yep, but do you mean it may be the case that the integer offset with require more that 64 bits...? For simplicity I used uint64_t for the offsets, but sure I can switch to APInt.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Then store instructions will be handled by the code below. Stores that were proven by MemorySSA analysis to not alias with __sha3 are excluded from the consideration. This is magic of MemorySSA.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Actually, I removed this TODO, as this check seems to be fine.

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.703 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.162 ║
║ Worst                              -3.261 ║

║ Mean                                0.695 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.624 ║
║ Mean                                0.517 ║
║ Best                               40.602 ║
║ Worst                              -1.031 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.008 ║
║ Total                               0.007 ║

```

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.703 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.162 ║
║ Worst                              -3.261 ║

║ Mean                                0.695 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.624 ║
║ Mean                                0.517 ║
║ Best                               40.602 ║
║ Worst                              -1.031 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.008 ║
║ Total                               0.007 ║

```

llvm/lib/Support/KECCAK.cpp
```diff
@@ -0,0 +1,285 @@
...
+#if defined(BYTE_ORDER) && defined(LITTLE_ENDIAN) && BYTE_ORDER == LITTLE_ENDIAN
+#define KECCAK_LITTLE_ENDIAN
+#endif
+
+#ifndef KECCAK_LITTLE_ENDIAN
+///
+/// Function to load a 64-bit value using the little-endian (LE) convention.
+/// On a LE platform, this could be greatly simplified using a cast.
+///
+static uint64_t load64(const uint8_t *x) {
```
akiramenai: static in an anonymous ns is a bit too much. LLVM [recommends](https://llvm.org/docs/CodingStandards.html#anonymous-namespaces) static.

llvm/test/CodeGen/SyncVM/sha3-constant-folding.ll
akiramenai: @PavelKopyl I believe it's a rare case which doesn't give us much. If it needs effort to implement, I'd go for a todo + a low priority ticket.

@akiramenai You are right, I've even checked this on out test suite and couldn't find such cases. Nevertheless, I mentioned this case in the CPR-1370.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+
+  if (getObjectSize(V, Size, DL, &TLI, Opts))
+    return Size;
+  return MemoryLocation::UnknownSize;
+}
+
+/// This structure holds an information about a single memory clobber.
+/// While walking upwards starting at a __sha3 call (which is a MemUse),
+/// we create a MemClobber instance for each dominating memory clobber
+/// (in current implementation - a store instruction) and put it in a
```
akiramenai, line 121: If a memory intrinsic we skip the transformation?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+///      which, on success, returns a computed keccak256
+///      hash value.
+///      Replace __sha3 values with the calculated hash values,
+///      removing the calls and invoking removeFromMSSA() to keep up to date
+///      the MemorySSA representation.
+///   3. Run simplifyInstructions() to simplify/clean up the F.
+///
+class FoldingState {
+  /// Types of the relative placement of the two memory locations.
+  enum OverlapType {
```
akiramenai: enum class is preferred.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+  }
+
+  /// Return collected MemorySSa uses corresponding to __sha3 calls.
+  const SmallVectorImpl<MemoryUse *> &getSHA3MemUses() const {
+    return SHA3MemUses;
+  }
+
+private:
+  /// Return true if the \p I is a call to the library function \p F.
+  bool isLibFuncCall(const Instruction *I, LibFunc F) const;
```
akiramenai, line 241: If I remember right, we also used something similar in one of your previous patches. If there are other use cases, it might worth to extract as a utility. Just a comment.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
akiramenai: Thank you. We could actually make this one shorter as the enum is documented. Like return overlap status of MemUseLoc and ClobberLoc or something, though it's clear from the signature.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+  /// \p MemUseLoc and they both refer to the same underlying object.
+  /// Return 'OR_None' if \p Clobber is known to not write
+  /// to the memory referring by \p MemUse.
+  /// Returns 'OW_Unknown' if nothing can be determined.
+  OverlapResult isOverlap(const Instruction *MemUse, const Instruction *Clobber,
+                          const MemoryLocation &MemUseLoc,
+                          const MemoryLocation &ClobberLoc) const;
+
+  /// Try to cast the constant pointer value \p Ptr to the integer offset.
+  /// TODO: CPR-1380.
```
akiramenai: Nitpicking.
A brief text describing the issue is of use for those who doesn't want to lookup the ticket.
As we are going to unify it anyway, it's ok to keep as it is.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+      // __sha3() is passed a pointer in the first argument
+      // and the memory size in the second one. It's expected that the
+      // memory size to be a constant expression. In this case
+      // the memory location should be precise.
+      if (MU && getLocForSHA3Call(Call).Size.isPrecise())
+        SHA3MemUses.push_back(MU);
+    }
+  }
+  // Collect whether there is any irreducible control flow in the function.
+  ContainsIrreducibleLoops = mayContainIrreducibleControl(F, &LI);
```
akiramenai: It could be put to the initializer list and made `const`. One less suspicion for a reader.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+  //
+  // We start with a nearest to the __sha3 call dominating clobbering access.
+  // Then we do the same for the just found clobber and so on until we find the
+  // 'live on entry'.
+  // For simplicity, in case of a MemoryPhi we also stop the search. This
+  // constraint can be relaxed. TODO: CPR-1370.
+
+  MemoryAccess *MA =
+      Walker->getClobberingMemoryAccess(MSSA.getMemoryAccess(Call));
+  for (;;) {
```
akiramenai: Why not to move what leads to break to the loop condition and change to while? Isn't `for (;;)` kind of exotic?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+  }
+  return Changed;
+}
+
+bool FoldingState::isGuaranteedLoopIndependent(
+    const Instruction *MemUse, const Instruction *Clobber,
+    const MemoryLocation &MemUseLoc) const {
+  // If the dependency is within the same block or loop level (being careful
+  // of irreducible loops), we know that AA will return a valid result for the
+  // memory dependency. (Both at the function level, outside of any loop,
```
akiramenai: I didn't quite get about compilation time.
We request `getLoopFor` for both `MemUse` and `Clobber` bbs. Shouldn't these calls return `nullptr` in case of outside of any loop. Can't we rely on it?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
akiramenai: We shall not ICE when user writes something to let's say 1234567890123456789. We can just stop optimising because the contract will eat all gas, but we shouldn't fail to compile. So it's up to you guard or switch to APInt.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
akiramenai: Not everywhere.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+
+  // If we can't resolve the same pointers to the same object, then we can't
+  // analyze them at all.
+  if (MemUseUndObj != ClobberUndObj) {
+    if (AAR == AliasResult::NoAlias)
+      return {OL_None, 0};
+    return {OL_Unknown, 0};
+  }
+
+  // Okay, we have stores to two completely different pointers.  Try to
```
akiramenai: Two spaces. Here and below.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+}
+
+std::optional<uint64_t> FoldingState::checkMemoryClobbers(
+    const SmallVector<MemClobber> &MemClobbers) const {
+  uint64_t TotalSize = 0;
+  auto Begin = MemClobbers.begin(), End = MemClobbers.end();
+  auto It = Begin, ItPrev = Begin;
+  for (; It != End; ++It) {
+    TotalSize += It->Size;
+    if (It == Begin)
```
akiramenai: Why can't we initialise TotalSize with Begin->Size, and It with std::next(Begin) thus removing this check?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+      continue;
+
+    if ((ItPrev->Start + ItPrev->Size) != It->Start) {
+      LLVM_DEBUG(dbgs() << "\tclobbers do alias, or there is a gep: ["
```
akiramenai: gap?

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
```diff
@@ -0,0 +1,822 @@
...
+                               DominatorTree &DT, const TargetLibraryInfo &TLI,
+                               const LoopInfo &LI) {
+  LLVM_DEBUG(dbgs() << "********** SyncVM sha3 constant folding **********\n"
+                    << "********** Function: " << F.getName() << '\n');
+
+  FoldingState State(F, AA, AC, MSSA, DT, TLI, LI);
+  SmallSet<MemoryUse *, 16> RemovedSHA3;
+
+  bool Changed = false;
+  for (;;) {
```
akiramenai: do while?

akiramenai (changes requested):
It became much better, thank you.
But there are still some things to address.

llvm/lib/Support/KECCAK.cpp
PavelKopyl: Yep, thank you.

llvm/test/CodeGen/SyncVM/sha3-constant-folding.ll
hedgar2017: > Yep, our AA detects this case, but I decided not to support this it for simplicity, because I haven't seen it in tests (thought I didn't check all of them).
> 
> @hedgar2017, probably you know more about __sha3. Can we have the following case, and how frequent it may be in a real life:
> 
> 
> 
> 
> 
> ```
> 
> |---Store1---|
> 
>      |---Store2--|
> 
>                  |---------Store3----|
> 
>    |--------sha3 memory ------| 
> 
> 
> 
> ```
> 
> 
> 
> ?
> 
> 
> 
> Currently, this pass supports only the following case:
> 
> 
> 
> ```
> 
> |---Store1---|
> 
>              |---Store2--|
> 
>                          |-Store3----|
> 
> |--------sha3 memory ----------------| 
> 
> ```
> 
> 

Would be nice to have actually. Can be done in a follow-up ticket.

PavelKopyl: OK, I've mentioned this case in the [CRP-1370](https://linear.app/matterlabs/issue/CPR-1370/extent-sha3-constant-folding).

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yes, we skip the transformation in this case, because:
 - I haven't found such cases among the tests in our CI (as an experiment, I added llvm_unreachable() for this case, but no falls were observed)
 - This will require additional analysis to be implemented.
 
 An example, of such an example is shown below:
```
 declare void @llvm.memcpy.p1i256.p3i256.i256(ptr addrspace(1) noalias nocapture writeonly, ptr addrspace(3) noalias nocapture readonly, i256, i1 immarg)
 
define i256 @sha3_test_1() nounwind {
entry:
  store i256 304594385234, ptr addrspace(3) null, align 4294967296
  store i256 56457598675863654, ptr addrspace(3) inttoptr (i256 32 to ptr addrspace(3)), align 32
  call void @llvm.memcpy.p1i256.p3i256.i256(ptr addrspace(1) null, ptr addrspace(3) null, i256 64, i1 false)
  %hash = tail call i256 @__sha3(ptr addrspace(1) null, i256 64, i1 true)
  ret i256 %hash
}
declare i256 @__sha3(ptr addrspace(1), i256, i1) #0
attributes #0 = { argmemonly nofree nounwind null_pointer_is_valid readonly}
```

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK, done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Yes, we have similar functionality in the SyncVMLinkRuntime pass. I've created a [ticket](https://linear.app/matterlabs/issue/CPR-1386/move-syncvmsha3constfoldingislibfunccall-function-to-a-common-code) for this.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK, I added a small comment here.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK, moved to the initializer list.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Sorry, re-check again, as I found a mistake in the function declaration. It should be

```
bool isGuaranteedLoopIndependent(const Instruction *Clobber,
                                   const Instruction *MemUse,
                                   const MemoryLocation &ClobberLoc) const;
```

So, the check means that if a Clobber's base address value is within the F's entry block the Clobber considered as a loop invariant. It looks reasonable. By the way, I took this check from DSE pass and the discussion about it is here: https://reviews.llvm.org/D100464.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Fixed other cases of "Off".

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Agreed, done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Fexed.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Fixed.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Sorry, I removed this part regarding the time, as it has no big meaning for this pass.
Yep, getLoopFor can return nullptr. As for the case where both MemUse's Loop and the Clobbers's Loop are both nullptr... it's an interesting case. At first glance, this looks like we are loop invariant, but in general case we have to consider irreducible CFG. Initial author of this check seems to has no clear understanding when to do in this case:
"I was originally more worried about opening the door to any number of unrelated issues coming up from allowing DSE from different blocks were both LI's return nullptr." (See it here: https://reviews.llvm.org/D100464?id=337391#inline-947892). On the other hand, if our front end doesn't generate irreducible CFG we may drop isGuaranteedLoopInvariant, I guess. I'll make an experiment with the removed check isGuaranteedLoopIndependent.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: OK, done.

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Done, thank you.

zksync-admin-bot2:
Benchmark results:
```
║ Mean                                0.413 ║
║ Best                               68.702 ║
║ Worst                              -8.805 ║
║ Total                               0.702 ║
║ Mean                              100.000 ║
║ Best                               96.932 ║
║ Worst                             -26.446 ║
║ Total                               0.282 ║
║ Mean                                0.130 ║
║ Best                               78.162 ║
║ Worst                              -3.261 ║

║ Mean                                0.695 ║
║ Best                                7.815 ║
║ Worst                              -5.922 ║
║ Total                               0.624 ║
║ Mean                                0.517 ║
║ Best                               40.602 ║
║ Worst                              -1.031 ║
║ Total                               0.160 ║
║ Mean                                0.134 ║
║ Best                               40.392 ║
║ Worst                              -0.008 ║
║ Total                               0.007 ║

```

llvm/lib/Target/SyncVM/SyncVMSHA3ConstFolding.cpp
PavelKopyl: Actually, I came back to `for` loop, but I moved this two break conditions into for loop condition.

akiramenai (approved):
LGTM, thanks

akiramenai:
> I'll make an experiment with the removed check isGuaranteedLoopIndependent

Have you tried?

akiramenai:
I see, it's an another PR.

PR #313, opened 2023-10-01
[WIP][EraVM] Add a pass to recognize mem-ops for their indexed version

TerryGuo:
Find mem-ops that can be optimized to indexed version, re-write them into patterns that can be recognized by the subsequent combine pass where the original mem-ops will be replaced by their indexed version.

PR #314, opened 2023-10-01
[WIP][SyncVM] Use indexed load/store to lower memcpy

TerryGuo:
When lower memcpy with known size, do not use loop index to increase the base ptr. Instead, use an explicit ADD to increase the base ptr, the subsequent combine pass can recoginze such pattern and turn them into indexed load/store.


akiramenai:
#313 is a more general approach we'd better to stick. But it worth to fix this one as well to check if we are going to get at least the same performance numbers with #314. Thank you for work!

akiramenai:
@TerryGuo can we close it?

PR #374, opened 2023-12-21
[EraVM] Add EraVM specific CSE pass

vladimirradosavljevic:


zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.025 ║
║ Best                               21.818 ║
║ Total                               0.095 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.035 ║
║ Best                               51.598 ║
║ Worst                              -5.000 ║
║ Total                               0.003 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.005 ║
║ Best                               10.536 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.015 ║
║ Best                               20.339 ║
║ Worst                              -1.990 ║
║ Total                               0.035 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.009 ║
║ Best                               21.429 ║
║ Worst                             -10.487 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.004 ║
║ Best                                9.821 ║
║ Worst                              -7.666 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.676 ║
║ Best                                4.187 ║
║ Total                               1.015 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.121 ║
║ Best                               23.932 ║
║ Total                               0.264 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.014 ║
║ Best                                0.667 ║
║ Total                               0.012 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.028 ║
║ Best                                0.919 ║
║ Total                               0.061 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.007 ║
║ Best                                1.102 ║
║ Total                               0.006 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.002 ║
║ Best                                0.583 ║

```

zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.025 ║
║ Best                               21.818 ║
║ Total                               0.095 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.035 ║
║ Best                               51.598 ║
║ Worst                              -5.000 ║
║ Total                               0.003 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.005 ║
║ Best                               10.536 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.015 ║
║ Best                               20.339 ║
║ Worst                              -1.990 ║
║ Total                               0.035 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.009 ║
║ Best                               21.429 ║
║ Worst                             -10.487 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.004 ║
║ Best                                9.821 ║
║ Worst                              -7.666 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.676 ║
║ Best                                4.187 ║
║ Total                               1.015 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.121 ║
║ Best                               23.932 ║
║ Total                               0.264 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.014 ║
║ Best                                0.667 ║
║ Total                               0.012 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.028 ║
║ Best                                0.919 ║
║ Total                               0.061 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.007 ║
║ Best                                1.102 ║
║ Total                               0.006 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.002 ║
║ Best                                0.583 ║

```

zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.026 ║
║ Best                               21.818 ║
║ Worst                              -1.600 ║
║ Total                               0.098 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.035 ║
║ Best                               51.598 ║
║ Worst                              -5.000 ║
║ Total                               0.003 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.005 ║
║ Best                               10.536 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.016 ║
║ Best                               20.339 ║
║ Worst                              -1.990 ║
║ Total                               0.038 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.009 ║
║ Best                               21.429 ║
║ Worst                             -10.487 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.004 ║
║ Best                                9.821 ║
║ Worst                              -7.666 ║
║ Total                              -0.000 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.689 ║
║ Best                                4.187 ║
║ Total                               1.057 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.121 ║
║ Best                               23.932 ║
║ Total                               0.264 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.014 ║
║ Best                                0.667 ║
║ Total                               0.013 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.038 ║
║ Best                                0.994 ║
║ Total                               0.097 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.006 ║
║ Best                                1.102 ║
║ Worst                              -0.201 ║
║ Total                               0.005 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.002 ║
║ Best                                0.583 ║
║ Worst                              -0.001 ║

```

zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.026 ║
║ Best                               21.818 ║
║ Worst                              -1.600 ║
║ Total                               0.098 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.035 ║
║ Best                               51.598 ║
║ Worst                              -5.000 ║
║ Total                               0.003 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.005 ║
║ Best                               10.536 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.016 ║
║ Best                               20.339 ║
║ Worst                              -1.990 ║
║ Total                               0.038 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.009 ║
║ Best                               21.429 ║
║ Worst                             -10.487 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.004 ║
║ Best                                9.821 ║
║ Worst                              -7.666 ║
║ Total                              -0.000 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.689 ║
║ Best                                4.187 ║
║ Total                               1.057 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.121 ║
║ Best                               23.932 ║
║ Total                               0.264 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.014 ║
║ Best                                0.667 ║
║ Total                               0.013 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.038 ║
║ Best                                0.994 ║
║ Total                               0.097 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.006 ║
║ Best                                1.102 ║
║ Worst                              -0.201 ║
║ Total                               0.005 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.002 ║
║ Best                                0.583 ║
║ Worst                              -0.001 ║

```

zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.032 ║
║ Best                               22.535 ║
║ Worst                              -1.600 ║
║ Total                               0.105 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.073 ║
║ Best                               98.038 ║
║ Worst                              -5.000 ║
║ Total                               0.012 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.030 ║
║ Best                               96.556 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.021 ║
║ Best                               26.087 ║
║ Worst                              -1.990 ║
║ Total                               0.043 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.042 ║
║ Best                               90.576 ║
║ Worst                             -10.487 ║
║ Total                               0.008 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.024 ║
║ Best                               96.541 ║
║ Worst                              -7.666 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.689 ║
║ Best                                4.187 ║
║ Total                               1.057 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.121 ║
║ Best                               23.932 ║
║ Total                               0.264 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.014 ║
║ Best                                0.667 ║
║ Total                               0.013 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.038 ║
║ Best                                0.994 ║
║ Total                               0.097 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.006 ║
║ Best                                1.102 ║
║ Worst                              -0.201 ║
║ Total                               0.005 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.002 ║
║ Best                                0.583 ║
║ Worst                              -0.001 ║

```

zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.268 ║
║ Best                               59.443 ║
║ Worst                              -1.600 ║
║ Total                               1.074 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.269 ║
║ Best                               98.038 ║
║ Worst                              -5.000 ║
║ Total                               0.061 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.233 ║
║ Best                               96.556 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.193 ║
║ Best                               54.545 ║
║ Worst                              -1.990 ║
║ Total                               0.460 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.187 ║
║ Best                               90.576 ║
║ Worst                             -10.487 ║
║ Total                               0.057 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.183 ║
║ Best                               96.541 ║
║ Worst                              -7.666 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.724 ║
║ Best                                4.338 ║
║ Total                               1.167 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.161 ║
║ Best                               23.932 ║
║ Total                               0.323 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.015 ║
║ Best                                0.667 ║
║ Total                               0.016 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.058 ║
║ Best                                1.929 ║
║ Total                               0.171 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.047 ║
║ Best                                8.298 ║
║ Worst                              -0.201 ║
║ Total                               0.063 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.003 ║
║ Best                                0.583 ║
║ Worst                              -0.001 ║
║ Total                               0.003 ║

```

zksync-admin-bot2:
Benchmark results:
```
╔═╡ Size (-%) ╞════════════════╡ All M3B3 ╞═╗
║ Mean                                0.268 ║
║ Best                               59.443 ║
║ Worst                              -1.600 ║
║ Total                               1.074 ║
╠═╡ Cycles (-%) ╞══════════════╡ All M3B3 ╞═╣
║ Mean                                0.269 ║
║ Best                               98.038 ║
║ Worst                              -5.000 ║
║ Total                               0.061 ║
╠═╡ Ergs (-%) ╞════════════════╡ All M3B3 ╞═╣
║ Mean                                0.233 ║
║ Best                               96.556 ║
║ Worst                              -3.797 ║

╔═╡ Size (-%) ╞════════════════╡ All MzB3 ╞═╗
║ Mean                                0.193 ║
║ Best                               54.545 ║
║ Worst                              -1.990 ║
║ Total                               0.460 ║
╠═╡ Cycles (-%) ╞══════════════╡ All MzB3 ╞═╣
║ Mean                                0.187 ║
║ Best                               90.576 ║
║ Worst                             -10.487 ║
║ Total                               0.057 ║
╠═╡ Ergs (-%) ╞════════════════╡ All MzB3 ╞═╣
║ Mean                                0.183 ║
║ Best                               96.541 ║
║ Worst                              -7.666 ║

╔═╡ Size (-%) ╞════════╡ Precompiles M3B3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles M3B3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles M3B3 ╞═╣

╔═╡ Size (-%) ╞════════╡ Precompiles MzB3 ╞═╗
╠═╡ Cycles (-%) ╞══════╡ Precompiles MzB3 ╞═╣
╠═╡ Ergs (-%) ╞════════╡ Precompiles MzB3 ╞═╣

╔═╡ Size (-%) ╞══════════╡ Real life M3B3 ╞═╗
║ Mean                                0.724 ║
║ Best                                4.338 ║
║ Total                               1.167 ║
╠═╡ Cycles (-%) ╞════════╡ Real life M3B3 ╞═╣
║ Mean                                1.161 ║
║ Best                               23.932 ║
║ Total                               0.323 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life M3B3 ╞═╣
║ Mean                                0.015 ║
║ Best                                0.667 ║
║ Total                               0.016 ║

╔═╡ Size (-%) ╞══════════╡ Real life MzB3 ╞═╗
║ Mean                                0.058 ║
║ Best                                1.929 ║
║ Total                               0.171 ║
╠═╡ Cycles (-%) ╞════════╡ Real life MzB3 ╞═╣
║ Mean                                0.047 ║
║ Best                                8.298 ║
║ Worst                              -0.201 ║
║ Total                               0.063 ║
╠═╡ Ergs (-%) ╞══════════╡ Real life MzB3 ╞═╣
║ Mean                                0.003 ║
║ Best                                0.583 ║
║ Worst                              -0.001 ║
║ Total                               0.003 ║

```

llvm/include/llvm/Passes/PassBuilder.h
```diff
@@ -469,6 +469,16 @@ class PassBuilder {
     PipelineEarlySimplificationEPCallbacks.push_back(C);
   }
 
+  // EraVM local begin
+  /// Register a callback for a default optimizer pipeline extension point
+  ///
+  /// This extension point allows adding optimizations before the inliner.
+  void registerPreInlinerOptimizationsEPCallback(
```
akiramenai, line 504: Could you explain why adding a pass _right before_ the inliner this is necessary for our CSE.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+  assert(Call && "Expected call instruction.");
+  auto *Signature = dyn_cast<ConstantInt>(Call->getArgOperand(1));
+
+  // If this isn't valid constant signature, assume we can't eliminate it.
+  if (!Signature || Signature->getBitWidth() != 256)
+    return false;
+
+  const DenseSet<APInt> PreventingSignatures = {
+      // SELFBALANCE, BALANCE
+      {256,
```
akiramenai, line 295: Looks fragile. Can we redesign?

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+  return isa<CallBase>(Inst);
+}
+
+/// Return whether we can handle EraVM specific libcall.
+static bool canHandleStdlibCall(const Instruction *Inst) {
+  // Don't take into account instructions that return void.
+  if (Inst->getType()->isVoidTy())
+    return false;
+
+  const StringSet<> LibFuncNames = {
```
akiramenai, line 373: I expect baking settings like that deep into the implementation could cause all sorts of troubles. It looks like making LLVM-wide EraVM header describing configurations would be a better solution.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+                           Instruction *LaterInst) const;
+
+  /// Exclude the instruction from MSSA if it's going to be removed from a BB.
+  void removeMSSA(Instruction &Inst) const;
+
+  /// Remove dead instruction.
+  void removeDeadInstruction(Instruction &Inst, bool salvageDbgInfo) const;
+
+  /// Traverse CFG and try to find preventers.
+  std::pair<bool, bool> hasPreventersInCFG(BasicBlock *Begin, BasicBlock *End);
```
akiramenai, line 264: A more explicit type would make it less confusing. Or at least documentation on what is returned is expected.
Same below.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+  if (const auto *II = dyn_cast<IntrinsicInst>(Inst))
+    return II->getIntrinsicID() == Intrinsic::eravm_gasleft ||
+           II->getIntrinsicID() == Intrinsic::eravm_getu128;
+
+  const StringSet<> LibFuncNames = {"__sha3", "__system_request"};
+  return isLibFuncCall(Inst, LibFuncNames);
+}
+
+/// Check if instruction prevents elimination of some __system_request calls.
+static bool preventsSReqElimination(const Instruction *Inst) {
```
akiramenai: nitpicking: `const &` looks a better option taking into account the call sites.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+      ScopedHashTableVal<Instruction *, std::pair<Instruction *, unsigned>>>;
+  using InstHTType =
+      ScopedHashTable<Instruction *, std::pair<Instruction *, unsigned>,
+                      DenseMapInfo<Instruction>, InstAType>;
+  InstHTType AvailableCalls;
+
+  /// This is the current generation of the llvm.eravm.getu128 calls.
+  unsigned CurrentGetu128Gen = 0;
+
+  /// This is the current generation of the __system_request calls.
```
akiramenai, line 153: Could we define the term 'generation'?

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+    // If the instruction can be simplified (e.g. X+0 = X) then replace it with
+    // its simpler value.
+    if (Value *V = simplifyInstruction(&Inst, SQ)) {
+      LLVM_DEBUG(dbgs() << "EraVMCSE Simplify: " << Inst << "  to: " << *V
+                        << '\n');
+      if (!Inst.use_empty()) {
+        Inst.replaceAllUsesWith(V);
+        Changed = true;
+      }
+      if (isInstructionDead(&Inst, &TLI)) {
```
akiramenai: Code duplication. Worth extracting an auxiliary function.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,653 @@
...
+
+    // Check if the node needs to be processed.
+    if (!NodeToProcess->isProcessed()) {
+      // Process the node.
+      Changed |= processNode(NodeToProcess->node());
+      NodeToProcess->process();
+    } else if (NodeToProcess->childIter() != NodeToProcess->end()) {
+      // Push the next child onto the stack.
+      DomTreeNode *Child = NodeToProcess->nextChild();
+      nodesToProcess.push_back(new StackNode(AvailableCalls, &CurrentGetu128Gen,
```
akiramenai: I wonder if we can do better than naked `new`, `delete`.

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: Overall, the function is to be refactored.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
```diff
@@ -0,0 +1,802 @@
...
+  %val2 = call i256 @__system_request(i256 %address, i256 %sig, i256 %size, ptr nocapture readonly %ptr1)
+  %ret = add i256 %val1, %val2
+  ret i256 %ret
+}
+
+define i256 @test_noelim2(i256 %address, i256 %sig, i256 %size, ptr nocapture readonly %ptr1, i256 %key, i256 %val) {
+; CHECK-LABEL: @test_noelim2(
+; CHECK-NEXT:  entry:
+; CHECK-NEXT:    [[VAL1:%.*]] = call i256 @__system_request(i256 [[ADDRESS:%.*]], i256 [[SIG:%.*]], i256 [[SIZE:%.*]], ptr nocapture readonly [[PTR1:%.*]])
+; CHECK-NEXT:    call void @llvm.eravm.event(i256 [[KEY:%.*]], i256 [[VAL:%.*]], i256 0)
```
hedgar2017: This one is not a preventer.
Is it just another group of `noelim` due to the unknown signature?

hedgar2017 (changes requested):

llvm/lib/Target/EraVM/EraVMCSE.cpp
hedgar2017: The only way is relying on these constants, unfortunately.
In case you meant anything more high-level than this data structure.

llvm/include/llvm/Passes/PassBuilder.h
vladimirradosavljevic: It is necessary, because code is better optimized than if we add it in `registerPipelineEarlySimplificationEPCallback` where we are running `EraVMSHA3ConstFoldingPass`. Before I came to this conclusion, I had done experiments on one of the benchmarks.
Here is the code after `EraVMSHA3ConstFoldingPass` + `DSE`:
```
  %_118 = load i256, ptr %_1, align 32
  store i256 %_118, ptr addrspace(1) null, align 1
  %_119 = load i256, ptr %_1, align 32
  store i256 %_119, ptr addrspace(1) inttoptr (i256 32 to ptr addrspace(1)), align 1
  %sha3_call = call fastcc i256 @__sha3(ptr addrspace(1) noalias nocapture nofree noundef nonnull readonly align 32 null, i256 4, i1 false)
  %sha3_call20 = call fastcc i256 @__sha3(ptr addrspace(1) noalias nocapture nofree noundef nonnull readonly align 32 inttoptr (i256 32 to ptr addrspace(1)), i256 4, i1 false)
```
In order to inspect clobbers, not only we need to look at stores but we would need to look through the loads, what would make implementation more complex, and shouldn't be part of our CSE.

Just before the Inliner, InstCombine is called, and this code looks like this:
```
  %_118 = load i256, ptr %_1, align 32
  store i256 %_118, ptr addrspace(1) null, align 4294967296
  store i256 %_118, ptr addrspace(1) inttoptr (i256 32 to ptr addrspace(1)), align 32
  %sha3_call = call fastcc i256 @__sha3(ptr addrspace(1) noalias nocapture nofree noundef nonnull readonly align 32 null, i256 4, i1 false)
  %sha3_call20 = call fastcc i256 @__sha3(ptr addrspace(1) noalias nocapture nofree noundef nonnull readonly align 32 inttoptr (i256 32 to ptr addrspace(1)), i256 4, i1 false)
```
In this case, we "only" need to look at stores, and do elimination of the second `__sha3` call. 
To confirm my findings, I [moved](https://github.com/matter-labs/compiler-llvm/pull/373) `EraVMSHA3ConstFolding` just before the Inliner, and saw that we got better numbers. Please note that degradations in that PR are caused by `DeadArgumentElimination` not working as expected.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: I tried to find better solution for this, but taking into account APInt constructors, I didn't manage to find better way of describing this.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Could you please give me an example how would you do this easily, without taking much time?
The solution I can think of is to add all missing functions from `eravm-stdlib.ll` to `TargetLibraryInfo.def` and everything where it is needed, use `isLibFuncCall` from `EraVMSHA3ConstFolding.cpp` and move that implementation to common header. This would take time, and for our use, I decided to do string compare, and refactoring we can do later.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will improve comments for these functions.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will do.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: I took this trick from EarlyCSE, where they called it CurrentGeneration. In that context, it means that we can only do elimination if two calls are the same generation, i.e. there are no preventers between two calls. If you have better idea how to call it, please let me know.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: The purpose why I left this code duplication is because of the `LLVM_DEBUG` prints. If we have 1 function where we would do removal of dead instruction, we won't be able to print that instruction after that function, because we would delete it.
Could you please explain how would you refactor it?

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: I didn't want to change this code, because it is similar to `EarlyCSE` where it is well tested, and if someone sees this code, it would be easier to understand `EarlyCSE` also.
If you have hard opinion to change this, I will do it.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
vladimirradosavljevic: This is not a real world example, and I have added it just for the completeness. The reason why this is preventer is because event intrinsic is marked with sideeffect, and pointer we pass to calls is not marked as noalias, so Alias Analysis can't prove that event won't change this pointer.
Real world example is as follows:
```
  %system_request_calldata_array_pointer20644 = alloca [1 x i256], align 32
  store i256 %address15095, ptr %system_request_calldata_array_pointer20644, align 32
  %system_request_call20646 = call fastcc i256 @__system_request(i256 32778, i256 -44877977326897262784168444354156441158329539312518651612887364914072161059015, i256 36, ptr noalias nocapture nofree noundef nonnull align 32 %system_request_calldata_array_pointer20644)
```

As you can see, we are allocating stack memory, write to it, and pass to the function with the following attributes. In order to eliminate duplicated functions, we would need to inspect input data, since input pointers won't be the same in two calls.
Reason why I'm not using following pattern in tests is because it is easier to write test with one pointer that comes from function + we haven't yet implemented memoization based on input data, where we will add such tests.

vladimirradosavljevic:
@hedgar2017 Since the one of requests is to add option to enable/disable memoization of `__system_request` functions, do you want to have option to completely enable/disable our custom CSE, or just `__system_request` functions? Our custom CSE beside doing CSE and DCE of `__system_request` functions, is doing also:
1. CSE and DCE of `__sha3` calls.
2. CSE of other stdlib calls from `eravm-stdlibs.ll`.
3. CSE and DCE of `llvm.eravm.getu128` intrinsic.
4. DCE of `llvm.eravm.gasleft` intrinsic.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
hedgar2017: I am more concerned about these intrinsics and other types of functions marked with side effects.
Also, this example can be a real world one in system contracts, so it matters whether this kind of intrinsics is a preventer.

Am I getting this right, that we'll exclude such intrinsics from preventers when we implement the analysis of call graph, like the `dummy` functions we have in other tests?

hedgar2017:
> @hedgar2017 Since the one of requests is to add option to enable/disable memoization of `__system_request` functions, do you want to have option to completely enable/disable our custom CSE, or just `__system_request` functions? Our custom CSE beside doing CSE and DCE of `__system_request` functions, is doing also:
> 
> 1. CSE and DCE of `__sha3` calls.
> 2. CSE of other stdlib calls from `eravm-stdlibs.ll`.
> 3. CSE and DCE of `llvm.eravm.getu128` intrinsic.
> 4. DCE of `llvm.eravm.gasleft` intrinsic.

Disable both CSE and DCE for `__sha3` and `__system_request` only.
That is, the number of such calls must not decrease, as if they had side effects like non-static calls.

Intrinsics we don't need to make disableable.

llvm/lib/Target/EraVM/EraVMCSE.cpp
PavelKopyl: I agree, declaring them in TargetLibraryInfo.def seems to be a right way. By the way, they are already there:), please, check declarations with prefix "__xvm" (X - means either Era or EVM, because those functions are used by both targets). I'd like to note that anyway their full definition is scattered via several files: llvm/lib/Analysis/TargetLibraryInfo.cpp, llvm/include/llvm/Analysis/TargetLibraryInfo.def, llvm/lib/Transforms/Utils/BuildLibCalls.cpp, so we need to keep them in sync with llvm/lib/Target/SyncVM/syncvm-stdlib.ll. 
With those declarations, I guess we can avoid using their explicit names and use their IDs - `llvm::LibFunc_xvm_*`.
`isLibFuncCall` can also be simplified to use something like: 

```
LibFunc Func;                                                                
return TLI & TLI->getLibFunc(*F, Func))                    
```

instead of 

`return LibFuncNames.count(Callee->getName());`

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
vladimirradosavljevic: > I am more concerned about these intrinsics and other types of functions marked with side effects.
Also, this example can be a real world one in system contracts, so it matters whether this kind of intrinsics is a preventer.

Here, we have two kinds of preventers:
1. Memory preventers for which we are using complex LLVM machinery to tell us whether something is safe to CSE.
2. CFG preventers that we implemented, based only on instructions, not memory accesses.

For this test, preventer is from 1., and I did a quick investigation how to support it, but unfortunately didn't find a quick solution (I assume I would need to do clobber walk and to skip event intrinsic, but the tricky thing is to test all edge cases).

> Am I getting this right, that we'll exclude such intrinsics from preventers when we implement the analysis of call graph, like the dummy functions we have in other tests?

Unfortunately no, because this is intrinsic that is included in Memory Analysis since we mark it with sideeffects.

What I suggest here is to add a FIXME in the test, to create a ticket with the low priority, since as far as I saw, they want to disable CSE and DCE for `__system_request` and `__sha3` in system contracts as they would like to be in the control. Does this make sense to you?

llvm/include/llvm/Passes/PassBuilder.h
akiramenai: Thanks for explaining.
Could we leave a short version of it as documentation in the code?

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Thanks for the pointers where this support is added 🙏 Since not all libcalls from `eravm-stdlib.ll` are defined in `TargetLibraryInfo.def` (`__system_request`, `__clz`, `__ulongrem`),  this is something we would need to do first, and then create LLVM-wide EraVM header.
@akiramenai @PavelKopyl If you want, I can do that and refactor current implementation.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
hedgar2017: Adding `FIXME` is fine, but low priority is not.
Apparently this `event` intrinsic is not the only one that can prevent memoization. In user contracts you can have others, e.g. `setu128` IIRC is marked with side effects.

What we probably need is different types of side effects, for instance:
1. Default LLVM writes changing internal state or storage, e.g. writing to address spaces
2. Intrinsics changing the external state, like `event`, `inctx`, etc.
3. Our custom preventers from this patch, including not yet analyzed `dummy` function
Each type has different requirements for elimination and memoization, but without distinguising them we won't squeeze the easy performance gain.
For `__system_request`, only type 3 can be a preventer.

Is it correct that we're now done with type 1 and type 3 minus `dummy` function, but not yet with type 2?

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: Could we at least
1. Reference to where the thing is defined (better is to add notes both sides that the definition is to be kept in sync).
2. Not necessary now (the follow up ticket is ok) refactor it to a header. I'd prefer to have configuration like this move visible.

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: @vladimirradosavljevic it's ok to postpone the refactoring task as long as it's tracked in Linear.

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: I don't have a better on at the moment, but could we add your explanation to the code?

llvm/lib/Target/EraVM/EraVMCSE.cpp
hedgar2017: These are keccak's of function signatures defined in system contracts.
The signatures won't change because all existing EraVM contracts depend on them already.

I have them hardcoded on the FE:
https://github.com/matter-labs/compiler-llvm-context/blob/main/src/eravm/evm/ext_code.rs#L21
https://github.com/matter-labs/compiler-llvm-context/blob/main/src/eravm/evm/ext_code.rs#L39
https://github.com/matter-labs/compiler-llvm-context/blob/main/src/eravm/evm/ether_gas.rs#L49
Exact hashes won't be ever documented, and we don't have that kind of docs for SC, so you can either recompute them here from those preimages, or simply add these links.
Let me know if you'd like to have SC links rather than FE ones. Though keep in mind that any of them can be invalidated by occasional shifts in the source code.

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: In the similar block below
```
        if (!Inst.use_empty())
          Inst.replaceAllUsesWith(InVal.first);
```
I expect the check is not necessary. And perhaps I put this comment too early and we should start refactor from the next block. It looks that we have 3 or 4 identical up to the debug messages pieces of code. I also expect it's more or less apparent from Inst if it's a stdlib call, system request or getu128. If it's not, this part of the message can be passed as an argument of a function to be extracted. Does it make sense or you?

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: It's not that strong.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
vladimirradosavljevic: I think it is more about the question of how Memory Analysis works in the LLVM, and where it stops w.r.t. intrinsics. For EraVM backend, for all intrinsics that are marked with `IntrNoMem` or `IntrInaccessibleMemOnly` [here](https://github.com/matter-labs/compiler-llvm/blob/dev-15/llvm/include/llvm/IR/IntrinsicsEraVM.td), Memory Analysis won't stop when it finds these intrinsics, but for other will and we will face the same thing as here, it will say that something is not safe to simplify or delete if it finds them between two instructions.
With that said, since we are using common LLVM's MemoryAnalysis to check memory preventers, it will return that it is not the safe to do the `__system_request` CSE if there is `event`, `tol1`, `precompile` or `decommit` between identical calls. Since we are creating our custom CSE pass, we can skip these queries from MemoryAnalysis and say it is safe to do for `__system_request` CSE.
Please note that `setu128` and `inctx` are not preventers for optimizations that depends on MemoryAnalysis, thus they are not preventers in our custom CSE.
Will increase priority for that ticket.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: That makes sense, will refactor it that way. Thanks!

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will do.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will update [this](https://linear.app/matterlabs/issue/CPR-1386/move-syncvmsha3constfoldingislibfunccall-function-to-a-common-code) ticket with how we agreed here, and add reference to it.

llvm/include/llvm/Passes/PassBuilder.h
vladimirradosavljevic: Yes, will do it.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
hedgar2017: Is it correct that these 4 intrinsics are marked as they can mutate the memory passed to them?

hedgar2017: In fact, those attributes look extremely weird to me at the moment.
Why `event` is marked with side effects, and `inctx` is not?

vladimirradosavljevic: > Is it correct that these 4 intrinsics are marked as they can mutate the memory passed to them?

I'm not sure, @akiramenai could you please answer to this question?

> In fact, those attributes look extremely weird to me at the moment.
Why event is marked with side effects, and inctx is not?

When we got ~3% size and cycle improvements, I only focused on context intrinsics, so I marked some of them with `IntrInaccessibleMemOnly`.
For Context and Persistent memory access groups, is there any intrinsic that can change memory state of where current contract memory points to, i.e. where following optimization is not safe:
```
Store Value to certain location in address space
Intrinsic
Load same Value where it is stored
```
to
```
Store Value to certain location in address space
Intrinsic
Use Value from Store, and remove Load
```

In case there is no such intrinsic, we can easily mark Persistent memory access intrinsics with `IntrInaccessibleMemOnly`, and gain more performance, as they won't be preventers anymore for Memory Analysis.

hedgar2017: Intrinsics can never change memory. The CI will show if I'm wrong by any chance.
Their side effects are only about other VM state data.

@vladimirradosavljevic could you create a quick patch that would mark all intrinsics as read only w.r.t. all ASes?

vladimirradosavljevic: Of course, [here](https://github.com/matter-labs/compiler-llvm/pull/377) it is. I marked them with `InaccessibleMemOnly`, so they don't get CSEd.

llvm/include/llvm/Passes/PassBuilder.h
```diff
@@ -469,6 +469,19 @@ class PassBuilder {
     PipelineEarlySimplificationEPCallbacks.push_back(C);
   }
 
+  // EraVM local begin
+  /// Register a callback for a default optimizer pipeline extension point
+  ///
+  /// This extension point allows adding optimizations before the inliner
+  /// and it differs from PipelineEarlySimplificationEPCallbacks, because some
+  /// passes are called inbetween (e.g. InstCombinePass, SimplifyCFGOptions)
```
akiramenai: in between

llvm/include/llvm/Passes/PassBuilder.h
```diff
@@ -469,6 +469,19 @@ class PassBuilder {
...
+  /// that can in some cases create opportunities for some optimizations.
```
akiramenai: You've mentioned that it's simpler to analyze if CSE right before the inliner if I remember it right. Could we express it more explicitly? E.g. any passes in between create opportunities for some optimizations, but we added this EP to enable our CSE, didn't we?

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,710 @@
...
+    return II->getIntrinsicID() == Intrinsic::eravm_setu128 ||
+           II->getIntrinsicID() == Intrinsic::eravm_nearcall ||
+           II->getIntrinsicID() == Intrinsic::eravm_precompile ||
+           II->getIntrinsicID() == Intrinsic::eravm_decommit;
+  return isa<CallBase>(&Inst);
+}
+
+/// Return whether we can handle EraVM specific libcall.
+static bool canHandleStdlibCall(const Instruction *Inst) {
+  // Don't take into account instructions that return void.
```
akiramenai: You literally repeat the code with this comment. It would be better to explain why instead.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,710 @@
...
+}
+
+/// Return whether we can handle __system_request call.
+static bool canHandleSystemRequestCall(Instruction *Inst) {
+  // Don't take into account instructions that return void.
+  if (Inst->getType()->isVoidTy())
+    return false;
+
+  const StringSet<> SystemRequestName = {"__system_request"};
+  return isLibFuncCall(Inst, SystemRequestName);
```
akiramenai: Can be rewritten as a oneliner
`return Inst->getType()->isVoidTy() && isLibFuncCall(Inst, {"__system_request"});`
Same below.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,710 @@
...
+  // read/write memory, then we can safely return true here.
+  auto *EarlierMA = MSSA.getMemoryAccess(EarlierInst);
+  if (!EarlierMA)
+    return true;
+  auto *LaterMA = MSSA.getMemoryAccess(LaterInst);
+  if (!LaterMA)
+    return true;
+
+  // Since we know LaterDef dominates LaterInst and EarlierInst dominates
+  // LaterInst, if LaterDef dominates EarlierInst then it can't occur between
```
akiramenai: What is `it`?

llvm/lib/Transforms/Scalar/EarlyCSE.cpp
```diff
@@ -457,6 +469,13 @@ struct CallValue {
     CallInst *CI = dyn_cast<CallInst>(Inst);
     if (!CI || !CI->onlyReadsMemory())
       return false;
+
+    // EraVM local begin
```
akiramenai, line 473: Git blame would explain, but better is to document that we handle it in a separate pass.

llvm/test/CodeGen/EraVM/early-cse-no-cse-calls.ll
```diff
@@ -0,0 +1,73 @@
+; RUN: opt -passes='early-cse<memssa>' -S < %s | FileCheck %s
+
+target datalayout = "E-p:256:256-i256:256:256-S32-a:256:256"
+target triple = "eravm"
+
+define i256 @test_srequest_noelim1(i256 %address, i256 %sig, i256 %size, ptr noalias nocapture nofree noundef nonnull align 32 %ptr1) {
+; CHECK-LABEL: @test_srequest_noelim1(
+; CHECK-NEXT:  entry:
```
akiramenai: If we only want to check that both calls persist, perhaps we should do just that rather than checking the exact ir for match.
Same applies below.

llvm/test/CodeGen/EraVM/eravm-cse-intrinsics.ll
```diff
@@ -0,0 +1,282 @@
...
+; CHECK-LABEL: @test_getu128_noelim1(
+; CHECK-NEXT:    [[VAL1:%.*]] = call i256 @llvm.eravm.getu128()
+; CHECK-NEXT:    br i1 [[COND:%.*]], label [[THEN:%.*]], label [[ELSE:%.*]]
+; CHECK:       then:
+; CHECK-NEXT:    call void @llvm.eravm.setu128(i256 1)
+; CHECK-NEXT:    br label [[RET:%.*]]
+; CHECK:       else:
+; CHECK-NEXT:    br label [[RET]]
+; CHECK:       ret:
+; CHECK-NEXT:    [[VAL2:%.*]] = call i256 @llvm.eravm.getu128()
```
akiramenai: A follow up. This one can be replaced with 1, can't it?

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
```diff
@@ -0,0 +1,803 @@
...
+;
+entry:
+  %val1 = call i256 @__system_request(i256 %address, i256 %sig, i256 %size, ptr nocapture readonly %ptr1)
+  store i256 1, ptr %ptr2, align 1
+  %val2 = call i256 @__system_request(i256 %address, i256 %sig, i256 %size, ptr nocapture readonly %ptr1)
+  %ret = add i256 %val1, %val2
+  ret i256 %ret
+}
+
+; TODO: llvm.eravm.event shoudn't prevent CSE of __system_request calls.
```
akiramenai: Ticket please

akiramenai (changes requested):
Could we minimise the checks in the tests? Checking full IR output creates line noice so tests become harder to follow.

llvm/include/llvm/Passes/PassBuilder.h
vladimirradosavljevic: Will change it.

llvm/include/llvm/Passes/PassBuilder.h
vladimirradosavljevic: Not just only CSE, but also we would like to move `EraVMSHA3ConstFolding` here. I will add comment which type of optimizations can benefit from this extension point.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will add comment.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will do.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: Will fix the comment. It is taken from EarlyCSE.

llvm/lib/Transforms/Scalar/EarlyCSE.cpp
vladimirradosavljevic: Will add comment.

llvm/test/CodeGen/EraVM/early-cse-no-cse-calls.ll
vladimirradosavljevic: Will remove unnecessary checks.

llvm/test/CodeGen/EraVM/eravm-cse-intrinsics.ll
vladimirradosavljevic: No, since if we are coming from `then`, we are running into a `setu128` preventer.
Please note that #377 is changing attribute for `getu128`, so handle of this intrinsic in EraVMCSE and this test will be removed.

llvm/test/CodeGen/EraVM/eravm-cse-system-request.ll
vladimirradosavljevic: #377 is changing `event` attribute, so it won't be preventer and we don't need to do special handling.

hedgar2017 (approved):

llvm/include/llvm/Passes/PassBuilder.h
```diff
@@ -469,6 +469,20 @@ class PassBuilder {
     PipelineEarlySimplificationEPCallbacks.push_back(C);
   }
 
+  // EraVM local begin
+  /// Register a callback for a default optimizer pipeline extension point
+  ///
+  /// This extension point allows adding optimizations before the inliner
+  /// and it differs from PipelineEarlySimplificationEPCallbacks, because some
```
akiramenai, line 476: I'm picky about documentation. So if I'm wrong please just say no, it doesn't worth the time to polish, etc. I really mean it.
What I anticipate is that in time when some other person want to add a pass he might be confused if he needs to add a pre-inliner pass (Probably LLVM didn't bother to document EPCallbacks though). One might also want to recheck if we need this local change at all. So time would be wasted.
It's expected that later in pipeline, after some optimizations, that IR is better optimised. But we don't explain why we need it.
Maybe a good explanation should look like this 
EraVM adds passes to remove repetitive computation of read only library function calls (examples). These passes must work before the inliner, but to simplify inner pattern matching logic and make it work more frequently, we found it better to put it right before the inliner. [If we have some theoretical explanations, not just empirical, we should also put it here]

llvm/lib/Transforms/Scalar/EarlyCSE.cpp
akiramenai: Still here

llvm/lib/Target/EraVM/EraVMCSE.cpp
akiramenai: @vladimirradosavljevic Could we add a ticket to move logic that recognises a system request kind out of this pass? Technical debt, medium. Or maybe we should add it to CPR-1386 scope.

llvm/lib/Target/EraVM/EraVMCSE.cpp
```diff
@@ -0,0 +1,718 @@
...
+  // that can prevent elimination of some instructions.
+  if (BB->hasNPredecessorsOrMore(2)) {
+    assert(Node->getIDom() && "Expected parent dominator node.");
+
+    auto Preventers = hasPreventersInCFG(Node->getIDom()->getBlock(), BB);
+    CurrentSReqGen += Preventers.first;
+    CurrentMetaGen += Preventers.second;
+  }
+
+  for (auto &Inst : make_early_inc_range(*BB)) {
```
akiramenai, line 527: I still think this loop could be simplified.
We call lookup for the same instruction 4 times, we insert there 4 times, and it feels like early continue logic will help to unify. I don't think precision in diagnostic justifies the repetition.
Unless we have strong reasons to postpone, I'd suggest to refactor.

akiramenai (changes requested):

llvm/include/llvm/Passes/PassBuilder.h
vladimirradosavljevic: This is a good point that someone might be confused where to put pass, and I didn't think about it. I will add your explanation. Thanks!

llvm/lib/Transforms/Scalar/EarlyCSE.cpp
vladimirradosavljevic: I added explanation above `isEraVMNoCSECall` function, but I will move it here.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: I will add it in [CPR-1386](https://linear.app/matterlabs/issue/CPR-1386), since it makes sense to do global refactoring.

llvm/lib/Target/EraVM/EraVMCSE.cpp
vladimirradosavljevic: I have added repetition because different conditions must be met in order to remove duplicated calls:
1. For stdlib calls we just need to find identical call.
2. For `__sha3`, identical calls and memory preventers. Please note that we will add there CSE based on input data, so we will add there more code.
3. For `__system request`, identical calls, memory preventers and instruction preventers. Please note that we will add there CSE based on input data, so we will add there more code.
4. For `meta`, identical calls and instruction preventers.

Beside all of this, since these calls are not preventers and can do continue, we can in the end check other instructions and whether they are preventers for some CSE. As far as I understand, early continue would make to check this in a more complex way.
What I suggest is to think about more refactoring when we add CSE based on input data for `__sha3` and `__system_request`. What do you think?

akiramenai: I got the part about different conditions and a check for continue will indeed be more complex by itself but could be extracted away into a `can...` method thus simplifying the loop.
We can postpone if you prefer, please create ticket in that case. I can't anticipate what the additions you mention would change.

vladimirradosavljevic: Added the ticket for simplification.

akiramenai (approved):

PR #377, opened 2024-01-15
[EraVM] Update intrinsic attributes

vladimirradosavljevic:


llvm/include/llvm/IR/IntrinsicsEraVM.td
```diff
@@ -33,8 +33,8 @@ def int_eravm_inctx: Intrinsic<[], [], [IntrInaccessibleMemOnly]>;
 def int_eravm_setpubdataprice: Intrinsic<[], [llvm_i256_ty], [IntrInaccessibleMemOnly]>;
 
 // Persistent memory access.
-def int_eravm_event: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
-def int_eravm_tol1: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
+def int_eravm_event: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrInaccessibleMemOnly]>;
```
hedgar2017, line 38: Are you sure this semantic reflects what I described?
I tried to lookup the semantic but couldn't find a complete description.

In fact, this intrinsic does not even read any memory, but only works with registers.
The only one that **reads** memory (AS 1) is the `decommit` below.

But all of them have side effects unrelated to memory, so we probably need `IntrHasSideEffects` to make sure they are **never** eliminated.

hedgar2017 (changes requested):

llvm/include/llvm/IR/IntrinsicsEraVM.td
vladimirradosavljevic: Yes, `IntrInaccessibleMemOnly` (description is [here](https://github.com/matter-labs/compiler-llvm/blob/be006c1119cd9405ee0a3c8d4b6d1add85ca6807/llvm/docs/LangRef.rst#L1701)) is removing sideeffects w.r.t. memory accesses around this intrinsic, i.e. MemoryAnalysis is not stopping when it reaches this intrinsic. This attribute is weaker than `IntrNoMem` since it doesn't enable DCE and CSE for this intrinsic, whereas `IntrNoMem` does. This means that this intrinsic is never eliminated.
Here are the tests for intrinsics with this attribute:
1. [Test](https://github.com/matter-labs/compiler-llvm/blob/be006c1119cd9405ee0a3c8d4b6d1add85ca6807/llvm/test/CodeGen/EraVM/cse-context.ll#L79) for no CSE.
2. [Test](https://github.com/matter-labs/compiler-llvm/blob/be006c1119cd9405ee0a3c8d4b6d1add85ca6807/llvm/test/CodeGen/EraVM/aa-memory-opt.ll#L58) for memory optimizations.

It seems that our CI is currently broken, but if we mark intrinsic with `IntrInaccessibleMemOnly`, and there is a test failure, I would say that means there is actually sideeffect for that intrinsic w.r.t. Memory Accesses.

vladimirradosavljevic: Btw, I was reading our ISA specification, and I found something interesting:
`Precompiles are able to change data_pages.`
From what I understand is that we can't remove sideeffects from precompile and decommit since they actually can change current contract data.
So this version where we marked event and tol1 as `IntrInaccessibleMemOnly`, should be fine. Let's wait for CI to get fixed, and run again this version of the patch.
What do you think?

hedgar2017: @vladimirradosavljevic thanks for the explanation!
I need to verify everything, especially the changing data_pages thing.
Will get back to you soon!

sayon: Precompiles indeed take data from memory pages and write them back to memory pages. But this happens only for a couple of system contracts; in other cases log.precompile just burns gas.

hedgar2017: Hey @sayon, could you remind which intrinsics never mutate memory?

hedgar2017: @sayon I think only `precompile` and `decommit` can do that.
All others are memory read-only and some of them mutate non-memory data, like `inctx` and `set_pubdata_price`, `event`, `to_l1`, etc.

vladimirradosavljevic:
@sayon @hedgar2017 @akiramenai
Could you please check the list of the intrinsics and see if we marked them with the correct attributes? Please note that `IntrNoMem` means no side effects w.r.t. memory accesses + DCE + CSE, and `IntrInaccessibleMemOnly` means no side effects w.r.t. memory accesses and no DCE and CSE.

llvm/include/llvm/IR/IntrinsicsEraVM.td
hedgar2017: @vladimirradosavljevic I've updated the attributes with the following assumptions:
1. `IntrNoMem`: no side effects, and can be eliminated if the result is unused (e.g. `gasleft`).
2. `IntrInaccessibleMemOnly`: side effects, no memory access, never eliminated.
3. `IntrHasSideEffects`: side effects including memory access, never eliminated.

Question:
1. Can we memoize and eliminate `getu128` with only `setu128` as the preventer?

@sayon please say your final word on `decommit` and `precompile`. We can benefit if they cannot write to their input memory.

vladimirradosavljevic: 1. And also means enabling LLVM's CSE. This could lead to issues for `gasleft`. Please note that we are doing DCE of `gasleft` in our custom CSE. Could you please revert it back to the `IntrInaccessibleMemOnly`?
2. Correct.
3. Correct.

> Can we memoize and eliminate getu128 with only setu128 as the preventer?

Yes, we are doing that in our custom CSE. I have also added preventers, as calls to another functions (this will be improved when we add function attributes) and `precompile` and `decommit`.

hedgar2017: @vladimirradosavljevic ok understood the CSE thing. Does it mean that `gasleft` can be reordered and potentially put in a place unexpected by the user?

Let me just see the benchmarks out of curiosity, and then I'll revert.

vladimirradosavljevic: I would assume it could do anything. Please look [here](https://github.com/matter-labs/compiler-llvm/blob/be006c1119cd9405ee0a3c8d4b6d1add85ca6807/llvm/lib/Transforms/Scalar/GVNHoist.cpp#L231) in GVNHoist pass that we are using in our backend in class that says:
```
// Records all call instructions candidate for code hoisting.
```
So yes, `IntrNoMem` means no side effects w.r.t memory accesses, DCE, CSE, reordering.

hedgar2017: Thanks, reverted!

hedgar2017 (approved):

llvm/include/llvm/IR/IntrinsicsEraVM.td
```diff
@@ -20,21 +20,27 @@ def int_eravm_staticcall: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty]
...
+
+// Write access of inaccessible EraVM context data.
+def int_eravm_inctx: Intrinsic<[], [], [IntrInaccessibleMemOnly]>;
+def int_eravm_setpubdataprice: Intrinsic<[], [llvm_i256_ty], [IntrInaccessibleMemOnly]>;
+def int_eravm_event: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrInaccessibleMemOnly]>;
+def int_eravm_tol1: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrInaccessibleMemOnly]>;
+
+// Read-only access of accessible EraVM context data. 
 def int_eravm_gasleft: Intrinsic<[llvm_i256_ty], [], [IntrInaccessibleMemOnly]>;
 def int_eravm_getu128: Intrinsic<[llvm_i256_ty], [], [IntrInaccessibleMemOnly]>;
```
sayon: # About getu128 and setu128

The names are misleading as get/set are not referring to the same memory cell!
They refer to two values: set operates with the *write-only context register*, get retrieves a *captured read-only context value*.

Captured value is immutable, and stored in the stack frame. When we perform a (far) call, *captured context value* is assigned the current value of *context register* and stays for the duration of the contract.
Therefore we can memoize the result of `getu128` until the contract end, `setu128` does not affect it.

I argue that it makes it akin to a pure function, as long as we do not perform "multi-contract" analysis. If so, the right attribute for `int_eravm_getu128` is `IntrNoMem`.

llvm/include/llvm/IR/IntrinsicsEraVM.td
```diff
@@ -20,21 +20,27 @@ def int_eravm_staticcall: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty]
...
+// Write access of accessible EraVM context data. 
 def int_eravm_setu128: Intrinsic<[], [llvm_i256_ty], [IntrInaccessibleMemOnly]>;
-def int_eravm_inctx: Intrinsic<[], [], [IntrInaccessibleMemOnly]>;
-def int_eravm_setpubdataprice: Intrinsic<[], [llvm_i256_ty], [IntrInaccessibleMemOnly]>;
 
-// Persistent memory access.
-def int_eravm_event: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
-def int_eravm_tol1: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
+// Can have generic side effects, including writing to the input memory a.k.a. calldata.
 def int_eravm_precompile: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
```
sayon, line 42: Precompiles can read and write the heap page of the current contract (but not auxheap). So I guess `IntrHasSideEffects`

Note: kernel-only.

vladimirradosavljevic:
@hedgar2017 @sayon I think we should also add `IntrNoMerge` attribute for `gasleft`, since SimplifyCFG pass is doing following optimization:
```
define i256 @test(i1 %cond) {
  br i1 %cond, label %then, label %else

then:
  %val1 = call i256 @llvm.eravm.gasleft()
  %res1 = add i256 %val1, 5
  br label %ret

else:
  %val2 = call i256 @llvm.eravm.gasleft()
  %res2 = add i256 %val2, 6
  br label %ret

ret:
  %phi = phi i256 [ %res1, %then ], [ %res2, %else ]
  ret i256 %phi
}
```
to
```
define i256 @test(i1 %cond) {
ret:
  %val1 = tail call i256 @llvm.eravm.gasleft()
  %. = select i1 %cond, i256 5, i256 6
  %res2 = add i256 %val1, %.
  ret i256 %res2
}
```
I recall that in the past I had some problems with SimplifyCFG pass doing this, and `IntrNoMerge` helped me to prevent this behaviour.
Please note that same thing would happen if `gasleft` has `IntrHasSideEffects` attribute. SimplifyCFG doesn't respect that...
WDYT?

vladimirradosavljevic:
I took liberty and squashed commits and added mentioned attribute. Let's see what the numbers will be, since I expect to see some degradations.

hedgar2017:
> @hedgar2017 @sayon I think we should also add `IntrNoMerge` attribute for `gasleft`, since SimplifyCFG pass is doing following optimization:
> 
> ```
> 
> define i256 @test(i1 %cond) {
> 
>   br i1 %cond, label %then, label %else
> 
> 
> 
> then:
> 
>   %val1 = call i256 @llvm.eravm.gasleft()
> 
>   %res1 = add i256 %val1, 5
> 
>   br label %ret
> 
> 
> 
> else:
> 
>   %val2 = call i256 @llvm.eravm.gasleft()
> 
>   %res2 = add i256 %val2, 6
> 
>   br label %ret
> 
> 
> 
> ret:
> 
>   %phi = phi i256 [ %res1, %then ], [ %res2, %else ]
> 
>   ret i256 %phi
> 
> }
> 
> ```
> 
> to
> 
> ```
> 
> define i256 @test(i1 %cond) {
> 
> ret:
> 
>   %val1 = tail call i256 @llvm.eravm.gasleft()
> 
>   %. = select i1 %cond, i256 5, i256 6
> 
>   %res2 = add i256 %val1, %.
> 
>   ret i256 %res2
> 
> }
> 
> ```
> 
> I recall that in the past I had some problems with SimplifyCFG pass doing this, and `IntrNoMerge` helped me to prevent this behaviour.
> 
> Please note that same thing would happen if `gasleft` has `IntrHasSideEffects` attribute. SimplifyCFG doesn't respect that...
> 
> WDYT?

To summarize, we must make `gasleft` totally immovable relatively to adjacent operations, but it must be eliminated if the result is unused.
Is it achievable with such attributes?

vladimirradosavljevic:
> > @hedgar2017 @sayon I think we should also add `IntrNoMerge` attribute for `gasleft`, since SimplifyCFG pass is doing following optimization:
> > ```
> > 
> > define i256 @test(i1 %cond) {
> > 
> >   br i1 %cond, label %then, label %else
> > 
> > 
> > 
> > then:
> > 
> >   %val1 = call i256 @llvm.eravm.gasleft()
> > 
> >   %res1 = add i256 %val1, 5
> > 
> >   br label %ret
> > 
> > 
> > 
> > else:
> > 
> >   %val2 = call i256 @llvm.eravm.gasleft()
> > 
> >   %res2 = add i256 %val2, 6
> > 
> >   br label %ret
> > 
> > 
> > 
> > ret:
> > 
> >   %phi = phi i256 [ %res1, %then ], [ %res2, %else ]
> > 
> >   ret i256 %phi
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > to
> > ```
> > 
> > define i256 @test(i1 %cond) {
> > 
> > ret:
> > 
> >   %val1 = tail call i256 @llvm.eravm.gasleft()
> > 
> >   %. = select i1 %cond, i256 5, i256 6
> > 
> >   %res2 = add i256 %val1, %.
> > 
> >   ret i256 %res2
> > 
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I recall that in the past I had some problems with SimplifyCFG pass doing this, and `IntrNoMerge` helped me to prevent this behaviour.
> > Please note that same thing would happen if `gasleft` has `IntrHasSideEffects` attribute. SimplifyCFG doesn't respect that...
> > WDYT?
> 
> To summarize, we must make `gasleft` totally immovable relatively to adjacent operations, but it must be eliminated if the result is unused. Is it achievable with such attributes?

These attributes should make `gasleft` totally immovable, and I will do some experiments in order to prove this. Regarding removing unused results, we will be doing that in our custom CSE.

llvm/include/llvm/IR/IntrinsicsEraVM.td
sayon: I had some concerns about `gasleft`. 

It seems to me that we need to satisfy two sets of constraints:

1. `gasleft` should be a two-way compiler barrier, preventing moving prior instructions past `gasleft`, and preventing moving following instruction before `gasleft`.
2. `gasleft` should not interfere with memory analysis i.e. it should not invalidate the knowledge about memory values. 
For example, if I stored `x -> loc`, then issued `gasleft`, I do not need to read from `loc` and may just use the value `x` directly.


If we are positive that `gasleft` will not be reordered with prior and following instructions, then `IntrInaccessibleMemOnly` should be fine. This **requires all intrinsic calls to act as two-way compiler barriers by default**. If this claim is false, I speculate that `HasSideEffects` will force the barrier semantics on `gasleft` (but will also invalidate all potentially cached values, which is bad).

llvm/include/llvm/IR/IntrinsicsEraVM.td
```diff
@@ -20,21 +20,27 @@ def int_eravm_staticcall: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty]
...
+
+// Write access of accessible EraVM context data.
+def int_eravm_setu128: Intrinsic<[], [llvm_i256_ty], [IntrInaccessibleMemOnly]>;
 
-// Persistent memory access.
-def int_eravm_event: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
-def int_eravm_tol1: Intrinsic<[], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
+// Can have generic side effects, including writing to the input memory a.k.a. calldata.
 def int_eravm_precompile: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
 def int_eravm_decommit: Intrinsic<[LLVMQualPointerType<llvm_i8_ty, 3>], [llvm_i256_ty, llvm_i256_ty], [IntrHasSideEffects]>;
```
sayon, line 43: @shamatar said that `decommit` writes to heap memory, does not read, does not interact with auxheap. So maybe `IntrWriteMem`? 
Also, can we use `WriteOnly` on the pointer somehow, or it can only be applied to parameters, not return values?

llvm/include/llvm/IR/IntrinsicsEraVM.td
vladimirradosavljevic: Should this be vice versa?
* `sys_precompile` with `IntrHasSideEffects`, since they can change heap page of the current contract
* `usr_precompile` with `IntrInaccessibleMemOnly`, since it just burn gas
I think these two cases are distinguishable and we can detect burn gas if arguments are constant/known.

shamatar: No opcode can write to calldata..

llvm/include/llvm/IR/IntrinsicsEraVM.td
shamatar: Just in case, decommit has internal optimizations, so it may overwrite heap (and give pointer to a slice of it), or may just give pointer to some older page used for the decommitment of the same hash. It's impossible to know statically, but it's fyi in case if there are special memory regions/address spaces involved

llvm/include/llvm/IR/IntrinsicsEraVM.td
akiramenai: @sayon it would increase precision of gasleft but perhaps beyond reasonable. LLVM anyway creates a mishmash from your code, so any cardinal (in contrast with ordinal) expectations on `gasleft()` are futile. LLVM will not schedule `gasleft()` after its use. And it will try to move `gasleft()` closer to a use which is the only sensible way to use it with O3 anyway. And if you have multiple uses of the same `gasleft()` result so hoisting could happen and you need precision... well it looks like a misuse.

akiramenai:
> These attributes should make gasleft totally immovable, and I will do some experiments in order to prove this. Regarding removing unused results, we will be doing that in our custom CSE.

@vladimirradosavljevic But why? The transformation you shown don't make `gasleft()` misuse any worse. If you start doing arithmetic on it and expect to get some sensible result... with O3 you will be dissapointed anyway. IMO, `gasleft()` are for branches and far calls, outside of that contexts it doesn't make much sense.
We can't ever guarantee that
```
x = gasleft()
// some_yul
y = gasleft()
assert(x - y == costof(some_yul) + costof(gasleft));
```
So what's the point to avoid changing the exact cost of instructions executed in between `gasleft()` call and use in a particular pass? If it's sufficiently complex logic, we can't provide any guarantees anyway no matter if it's in Solidity or YUL (and we don't expose our assembler).

I'd say if one have
```
if (expr(gasleft()) cmp Value)
```
`expr(gasleft())` must be computed right before the comparison and branch. If you put `expr(gasleft())` somewhere else it's either UB or must be moved right before.
More precisely it should even become `if (gasleft() cmp expr'(Value))` where `expr'(Value)` is to be computed first.

vladimirradosavljevic:
> > These attributes should make gasleft totally immovable, and I will do some experiments in order to prove this. Regarding removing unused results, we will be doing that in our custom CSE.
> 
> @vladimirradosavljevic But why? The transformation you shown don't make `gasleft()` misuse any worse. If you start doing arithmetic on it and expect to get some sensible result... with O3 you will be dissapointed anyway. IMO, `gasleft()` are for branches and far calls, outside of that contexts it doesn't make much sense. We can't ever guarantee that
> 
> ```
> x = gasleft()
> // some_yul
> y = gasleft()
> assert(x - y == costof(some_yul) + costof(gasleft));
> ```
> 
> So what's the point to avoid changing the exact cost of instructions executed in between `gasleft()` call and use in a particular pass? If it's sufficiently complex logic, we can't provide any guarantees anyway no matter if it's in Solidity or YUL (and we don't expose our assembler).
> 
> I'd say if one have
> 
> ```
> if (expr(gasleft()) cmp Value)
> ```
> 
> `expr(gasleft())` must be computed right before the comparison and branch. If you put `expr(gasleft())` somewhere else it's either UB or must be moved right before. More precisely it should even become `if (gasleft() cmp expr'(Value))` where `expr'(Value)` is to be computed first.

This is a good point you made, and also in that sense one more example what user can't expect is stack spills and reloads that are needed to get correct execution.
Please note that I used arithmetics to just create a test case, as it was easier to do so.
Regarding `if (expr(gasleft()) cmp Value)`, have you seen so far misuse of it?

hedgar2017:
@akiramenai more often `gasleft` is used to measure expensive calls rather than single arithmetic instructions, but making an effort towards the latter would be a nice contribution to equivalence.

It is not required to have it totally immovable, but at least not hoisting it to the beginning of a function is a must if it's not guaranteed.
Ultimately, it would be nice if we could tell LLVM that *any* operation can invalidate `gasleft` w.r.t. to M0 LLVM IR.

By the way, what set of attributes do you prefer?

sayon:
I agree that O3 is quite limiting when it comes to gas predictability. But I fear that O3 can render `gasleft` completely useless.

Here is a quick example; I checked the assembly and our compiler  moves the check of `gas_left` prior to any function call. Are we alright with this?

```solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

interface IExternalContract {
    function f() external;

    function g() external returns (int256) ;
    function h(int256) external ;
  }


contract GasChecker {
    IExternalContract C;

    constructor(address _externalContractAddress) {
        C = IExternalContract(_externalContractAddress);
    }

    function checkGasAndCall() external {

        int256 x = C.g();
        if (gasleft() > 1000) {
            C.f();
            }
        C.h(x);
    }
}

```


Assembly:

```
__entry:
.func_begin0:
	nop	stack+=[2]
	add	r1, r0, r3
	shr.s	96, r3, r3
	and	@CPI0_0[0], r3, r3
	and!	1, r2, r2
	jump.ne	@.BB0_1
	add	128, r0, r2
	st.1	64, r2
	sub.s!	4, r3, r2
	jump.lt	@.BB0_2
	ld	r1, r1
	and	@CPI0_5[0], r1, r1
	sub.s!	@CPI0_6[0], r1, r1
	jump.ne	@.BB0_2
	context.get_context_u128	r1
	sub!	r1, r0, r1
	jump.ne	@.BB0_2
	sload	r0, r2
	add	@CPI0_7[0], r0, r1
	st.1	128, r1
	context.gas_left	r1
	and	@CPI0_2[0], r2, r2
	sub.s!	4, r2, r3
	jump.ne	@.BB0_15
	add	stack[@returndatasize], r0, r3
	sub.s!	32, r3, r1
	add	r3, r0, r4
	add.ge	32, r0, r4
	jump	@.BB0_25
 
...
 
.BB0_25:
	add	159, r4, r1
	and	224, r1, r1
	st.1	64, r1
	sub.s!	32, r3, r1
	jump.ge	@.BB0_26
.BB0_2:
	add	r0, r0, r1
	ret.revert.to_label	r1, @DEFAULT_FAR_REVERT
.BB0_26:
	ld.1	128, r1
	add	r1, r0, stack-[2]
	context.gas_left	r1
	sub.s!	1001, r1, r1
	jump.lt	@.BB0_30
	 ...

```

sayon:
It also seems that with O3 the call to `C.g` is moved before the first invokation of `gasleft`.
This is, as far as I understand, a rather typical usage of `gasleft`.  
```
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

interface IExternalContract {
    function f() external;

    function g() external returns (int256) ;
    function h(int256, uint256) external ;
}


contract GasChecker {
    IExternalContract C;

    constructor(address _externalContractAddress) {
        C = IExternalContract(_externalContractAddress);
    }

    function checkGasAndCall() external {

        uint256 start = gasleft();
        int256 x = C.g();
        uint256 end = gasleft();

        C.h(x, end-start);
    }
}
``` 


[test.txt](https://github.com/matter-labs/compiler-llvm/files/13976944/test.txt)
The full assembly is attached.

sayon:
If we chose to accept that `gasleft` is totally unpredictable in O3, maybe compiler should reject programs with `gasleft` when launched with `O3`?
Something like:

```sh
zksolc -O3 test.sol
# rejected because of gasleft usage
zksolc -O3 --force-accept-gasleft test.sol
# compiles but gasleft is a poor man's biased RNG
```

akiramenai:
@hedgar2017, @sayon 
I agree that `gasleft` should not be hoisted to predecessors. I see the following possible solutions for `gasleft`:
1. Sink to use or UB if impossible w/o crossing a side-effect border (but people in blockchain are terrified by 'UB' word).
2. gasleft == optnone for the function containing it (but it would make gasleft quite expensive and perhaps unexpectedly so)
3. Maybe `hasSideEffects` would prevent most of the transformations while keeping the instruction approximately where it was. So it wouldn't be entirely noice or RNG, but it's a signal + some noice (this one is probably the hardest to explain / document).

vladimirradosavljevic:
> @hedgar2017, @sayon I agree that `gasleft` should not be hoisted to predecessors. I see the following possible solutions for `gasleft`:
> 
> 1. Sink to use or UB if impossible w/o crossing a side-effect border (but people in blockchain are terrified by 'UB' word).
> 2. gasleft == optnone for the function containing it (but it would make gasleft quite expensive and perhaps unexpectedly so)
> 3. Maybe `hasSideEffects` would prevent most of the transformations while keeping the instruction approximately where it was. So it wouldn't be entirely noice or RNG, but it's a signal + some noice (this one is probably the hardest to explain / document).

For the 3rd option, `hasSideEffects` is not actually a flag that you can retrieve in the LLVM IR, it is just to prevent tablegen from generating other memory attributes, so this attribute, won't prevent sinking and hoisting. On the other hand, `IntrNoMerge` will prevent hoisting and sinking in the SimplifyCFG, and it is propagated down to the LLVM MIR, so backend is also aware of this (although we would need to check newest versions of the LLVM and cherry-pick bug fixes w.r.t handling this attribute in backend). If we want this option, I would vote for `IntrInaccessibleMemOnly` + `IntrNoMerge`. We can add these, so @sayon can check again his tests to see the results.

akiramenai:
Ok, then for this particular patch, I'd go with your suggestion @vladimirradosavljevic (and we might want to add @sayon tests as regression ones). I'd also recommend to add a note on that or a ticket for our future Solidity compiler or if we want something like 1 or 2 as an option or as a default.

akiramenai (approved):

vladimirradosavljevic:
@akiramenai Could you please do the review, since LLVM IR NoMerge is not propagated to the LLVM MIR for intrinsics, just for calls, so I needed to add NoMerge MIR flag explicitly.
@sayon Could you please take this version of the patch and try your tests? It would be good to check the impact of this.

akiramenai:
@vladimirradosavljevic @sayon 
May we actually attach @sayon's tests to the patch? If @sayon you don't mind to turn them into lit format.

vladimirradosavljevic:
> @vladimirradosavljevic @sayon May we actually attach @sayon's tests to the patch? If @sayon you don't mind to turn them into lit format.

That would be awesome!

akiramenai (approved):
LGTM

llvm/include/llvm/IR/IntrinsicsEraVM.td
```diff
@@ -20,21 +20,23 @@ def int_eravm_staticcall: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty]
 def int_eravm_mimiccall: Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty, llvm_i256_ty], []>;
 def int_eravm_nearcall: Intrinsic<[llvm_i256_ty], [LLVMPointerType<llvm_i256_ty>, llvm_i256_ty, llvm_vararg_ty], [Throws]>;
 
-// Get value from the context.
+// Read-only access of inaccessible EraVM context data.
 def int_eravm_this: Intrinsic<[llvm_i256_ty], [], [IntrNoMem]>;
 def int_eravm_caller: Intrinsic<[llvm_i256_ty], [], [IntrNoMem]>;
 def int_eravm_codesource: Intrinsic<[llvm_i256_ty], [], [IntrNoMem]>;
 def int_eravm_meta: Intrinsic<[llvm_i256_ty], [], [IntrNoMem]>;
```
sayon: Sorry I missed this one.
`context.meta` returns a packed structure containing:

``` 
      ergs_per_pubdata_byte: ergs;
      heap_size: mem_address;
      aux_heap_size: mem_address;
      this_shard_id: shard_id;
      caller_shard_id: shard_id;
      code_shard_id: shard_id; 
```

Therefore `IntrInaccessibleMemOnly` seems to be a better fit here: heap/auxheap sizes may increase after a load/store from heap/auxheap, or by a farcall (if we forward a bigger memory chunk).

sayon (changes requested):

sayon:
Sure, I'll figure out the lit test format and add tests here, either tonight or during the weekend [CPR-1502].

llvm/include/llvm/IR/IntrinsicsEraVM.td
hedgar2017: @sayon please resolve if no objections.

llvm/include/llvm/IR/IntrinsicsEraVM.td
hedgar2017: @sayon please resolve if no objections.

hedgar2017 (approved):

llvm/include/llvm/IR/IntrinsicsEraVM.td
sayon: OK let's keep it `IntrHasSideEffects` for now, and revise it when `decommit` is implemented.
[[CPR-1503]](https://linear.app/matterlabs/issue/CPR-1503/revise-decommit-intrinsic-once-it-is-fully-implemented-in-vm)

vladimirradosavljevic:
Added one TODO and the ticket (CPR-1505), since there is a relationship between LLVM IR intrinsic attributes and backend instruction attributes.
Thanks @hedgar2017 for improving comments. I have squashed your commit.

llvm/include/llvm/IR/IntrinsicsEraVM.td
vladimirradosavljevic: It looks like that order of meta instruction matters, so we should consider adding `IntrNoMerge` also. Let's wait for your experiments, so we can decide about it.

hedgar2017: @vladimirradosavljevic I think only heap and aux heap operations can mutate it, so unless we can detect those, we must treat it just like `gasleft`.

vladimirradosavljevic: Let's then include memoization of meta intrinsic in our custom CSE with these preventers (heap/auxheap and farcalls).

sayon:
# On gasleft

Seems correct now and I also called a false alarm on it -- the tests that I have posted actually do not trigger an invalid behavior. We figured it out thanks to @vladimirradosavljevic yesterday.

# On meta

I confirm that `meta` was being optimized too aggressively with `IntrNoMem` and is treated correctly with `IntrInaccessibleMemOnly`.

```yul
                let before := verbatim_0i_1o("meta")
                mstore(999,0)
                let after := verbatim_0i_1o("meta")
                mstore(1,0)
                mstore(42,before)
                mstore(43,after)
                return(0, 0)

```
## With `IntrNoMem`

```ll
runtime_code_call_block:                          ; preds = %entry
  %contract_call_simulation_meta.i = tail call i256 @llvm.eravm.meta()
  store i256 0, ptr addrspace(1) inttoptr (i256 999 to ptr addrspace(1)), align 1
  store i256 0, ptr addrspace(1) inttoptr (i256 1 to ptr addrspace(1)), align 1
  store i256 %contract_call_simulation_meta.i, ptr addrspace(1) inttoptr (i256 42 to ptr addrspace(1)), align 2
  store i256 %contract_call_simulation_meta.i, ptr addrspace(1) inttoptr (i256 43 to ptr addrspace(1)), align 1
  tail call void @llvm.eravm.return(i256 0)
  unreachable
}

```

```nasm
	st.1	999, r0
	st.1	1, r0
	context.meta	r1
	st.1	42, r1
	st.1	43, r1
```

## with `IntrInaccessibleMemOnly`

```ll
runtime_code_call_block:                          ; preds = %entry
  %contract_call_simulation_meta.i = tail call i256 @llvm.eravm.meta()
  store i256 0, ptr addrspace(1) inttoptr (i256 999 to ptr addrspace(1)), align 1
  %contract_call_simulation_meta1.i = tail call i256 @llvm.eravm.meta()
  store i256 0, ptr addrspace(1) inttoptr (i256 1 to ptr addrspace(1)), align 1
  store i256 %contract_call_simulation_meta.i, ptr addrspace(1) inttoptr (i256 42 to ptr addrspace(1)), align 2
  store i256 %contract_call_simulation_meta1.i, ptr addrspace(1) inttoptr (i256 43 to ptr addrspace(1)), align 1
  tail call void @llvm.eravm.return(i256 0)
  unreachable
```

```nasm
	context.meta	r1
	st.1	999, r0
	context.meta	r2
	st.1	1, r0
	st.1	42, r1
	st.1	43, r2
```

I figure that we probably don't need a test for `gasleft` but we do for `meta`. Is it correct?
In any case I would approve this PR and add tests in a separate one to unblock the release, if this is alright.

sayon (approved):

vladimirradosavljevic:
@sayon Tests for meta will be part of our custom CSE, where we will remove duplicated calls where it is safe.

hedgar2017:
@sayon tests for `gasleft` should be in CSE as we cannot do this kind of prevention with the mere intrinsic attributes, if I understood @vladimirradosavljevic correctly.

UPD: @vladimirradosavljevic hehe, you were first.

PR #380, opened 2024-01-17
[InstCombine] Simplify -X > ~X to X != 0

vladimirradosavljevic:
# Code Review Checklist

## Requirements
- [x] Have the requirements been met?
- [x] Have stakeholder(s) approved the change?

## Implementation
- [x] Does this code change accomplish what it is supposed to do?
- [x] Can this solution be simplified?
- [x] Does this change add unwanted compile-time or run-time dependencies?
- [x] Could an additional framework, API, library, or service improve the solution?
- [x] Could we reuse part of LLVM instead of implementing the patch or a part of it?
- [x] Is the code at the right abstraction level?
- [x] Is the code modular enough?
- [x] Can a better solution be found in terms of maintainability, readability, performance, or security?
- [x] Does similar functionality already exist in the codebase? If yes, why isn’t it reused?
- [x] Are there any best practices, design patterns or language-specific patterns that could substantially improve this code? 

## Logic Errors and Bugs
- [x] Can you think of any use case in which the
- [x] Can you think of any inputs or external events
that could break the code?

## Error Handling and Logging
- [x] Is error handling done the correct way?
- [x] Should any logging or debugging information
be added or removed?
- [x] Are error messages user-friendly?
- [x] Are there enough log events and are they
debugging?

## Maintainability
- [x] Is the code easy to read?
- [x] Is the code not repeated (DRY Principle)?
- [x] Is the code method/class not too long?

## Dependencies
- [x] Were updates to documentation, configuration, or readme files made as required by this change?
- [x] Are there any potential impacts on other parts of the system or backward compatibility?

## Security
- [x] Does the code introduce any security vulnerabilities?

## Performance
- [x] Do you think this code change decreases
system performance?
- [x] Do you see any potential to improve the

## Testing and Testability
- [x] Is the code testable?
- [x] Have automated tests been added, or have related ones been updated to cover the change?
	- [x] For changes to mutable state
- [x] Do tests reasonably cover the code change (unit/integration/system tests)? 
	- [x] Line Coverage
	- [x] Region Coverage
	- [x] Branch Coverage
- [x] Are there some test cases, input or edge cases

## Readability
- [x] Is the code easy to understand?
- [x] Which parts were confusing to you and why?
- [x] Can the readability of the code be improved by
smaller methods?
- [x] Can the readability of the code be improved by
- [x] Is the code located in the right
file/folder/package?
- [x] Do you think certain methods should be
flow?
- [x] Is the data flow understandable?
- [x] Are there redundant or outdated comments?
- [x] Could some comments convey the message
better?
- [x] Would more comments make the code more
understandable?
- [x] Could some comments be removed by making the code itself more readable?
- [x] Is there any commented-out code?

## Documentation
- [x] Is there sufficient documentation?
- [x] Is the ReadMe.md file up to date?

## Best Practices
- [x] Follow Single Responsibility principle?
- [x] Are different errors handled correctly?
- [x] Are errors and warnings logged?
- [x] Magic values avoided?
- [x] No unnecessary comments?
- [x] Minimal nesting used?

## Experts' Opinion
- [x] Do you think a specific expert, like a security
- [x] Will this code change impact different teams, and should they review the change as well?


vladimirradosavljevic:
[Here](https://godbolt.org/z/r154e8efK) is the proof for this optimization. Only difference is in the test3, where I on purpose left to do this optimization even though there are uses, since we can do more stuff with `ICMP_NE` in our backend.

akiramenai:
Can we upstream?

vladimirradosavljevic:
@akiramenai No, since they are already doing this transformation, but in a different way. [Here](https://linear.app/matterlabs/issue/CPR-1488/optimize-sub-xor-ugt-pattern) is the explanation why I decided to do this pattern matching manually, instead of cherry picking upstream patches.

akiramenai:
So is it the case that the patch is supposed to go off when LLVM update (either the upcoming one or one of subsequent)? If so can we reflect that in the commit message?

akiramenai:
Also please add the reference to the proof to the commit message.

vladimirradosavljevic:
We should revert this patch when we update to LLVM-18, because I checked it, and it is not a part of LLVM-17. Added comments regarding proof and when should we revert it.

akiramenai (approved):
LGTM, thanks

PR #384, opened 2024-01-26
[EraVM] Change __addmod implementation

vladimirradosavljevic:


akiramenai (approved):
LGTM,
But if we can add a minimal test that shows perf improvement it would be better.

vladimirradosavljevic:
> LGTM, But if we can add a minimal test that shows perf improvement it would be better.

Creating test from `p256verify` precompile will be hard, so my suggestion is to wait for the results and maybe we will be lucky to find a small test that shows performance improvement.

akiramenai:
@vladimirradosavljevic please create a follow up ticket to dig into perf regressions on real life group.

vladimirradosavljevic:
@akiramenai This should fix regressions in real life and numbers in p256verify should stay the same.

vladimirradosavljevic:
@akiramenai I managed to create the minimal test to show perf improvement. It turns out that with the new implementation, ISel optimizations are able to remove one select instruction.

PR #385, opened 2024-01-30
[EraVM] Add freeze before use of first argument in __addmod

vladimirradosavljevic:


vladimirradosavljevic:
Alive2 before freeze: https://alive2.llvm.org/ce/z/73AsFT
Alive2 after freeze: https://alive2.llvm.org/ce/z/G9nKaA

FlashSheridan:
The freeze seems the correct way to avoid the problem of optimizing one use of %arg1m into multiple uses.  So LGTM, but I don’t know enough of the context for a formal approval.

akiramenai (approved):
LGTM

PavelKopyl (approved):
LGTM.

hedgar2017 (approved):

hedgar2017:
@vladimirradosavljevic can it be done in other places, `__mulmod` as the closest one?

vladimirradosavljevic:
@hedgar2017 In case we changed implementations from using overflow intrinsics to other non-overflow instructions, we should add them in alive2 to verify. Currently, I'm not sure whether we have to add freeze instruction in other implementations.

PR #387, opened 2024-02-06
[EraVM] Optimize patterns in precompiles

vladimirradosavljevic:


vladimirradosavljevic:
@akiramenai FYI, in precompiles there are only patterns of instructions that uses constant pools, other cases are added to generalize implementation.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -6,25 +6,13 @@
...
-//
-//   x = SEL non-zero-val, 0, cc
-//   ...
-//   z = ADDrrr_s non-zero-val, y
-//
-// can be folded into:
-//   z = ADDirr_s.cc non-zero-val, y
-//
-// if z and y can be allocated to same reg. The tie is used to ensure this.
+// This file implements a pre-ra pass to optimize SELECT and CMP instructions.
```
akiramenai: More explicit documentation is welcome here.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -79,12 +70,88 @@ class EraVMOptimizeSelectCmpPreRA : public MachineFunctionPass {
...
+  /// I.e. an instruction of form `add imm, rIn1, rOut`.
+  bool isImmAdd(const MachineInstr &MI) const;
+
+  /// Return whether MI is an add register.
+  /// I.e. an instruction of form `add rIn0, rIn1, rOut`.
+  bool isRegAdd(const MachineInstr &MI) const;
+
+  /// Return whether MI is a sub immediate that sets flags.
+  /// I.e. an instruction of form `sub.s! imm, rIn1, rOut`.
+  bool isImmSubWithFlags(const MachineInstr &MI) const;
```
akiramenai: Maybe `SubS` / `Subs`/`SwappedSub`? Swap is essential here. And the doc should reflect it too.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -79,12 +70,88 @@ class EraVMOptimizeSelectCmpPreRA : public MachineFunctionPass {
...
+  ///   sub rIN - const
+  bool isConstAddIdenticalToConstSub(const MachineInstr &Add,
+                                     const MachineInstr &Sub) const;
+
+  /// In case first instruction is a sub imm that doesn't set flags and the
+  /// second instruction is a sub imm that sets flags, return whether these
+  /// two instructions are identical.
+  /// They are identical in case they have the same immediate and the same
+  /// input register.
+  bool areSubsIdentical(const MachineInstr &FirstSub,
```
akiramenai: Shouldn't we generalise and consider other cases of identical `sub`s?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -79,12 +70,88 @@ class EraVMOptimizeSelectCmpPreRA : public MachineFunctionPass {
...
+  /// can be folded into:
+  ///   outSEL = ADD.cc imm/const/x, y
+  ///
+  /// if outSEL and x/y can be allocated to same reg.
+  /// The tie is used to ensure this.
+  bool tryFoldAddToSelect(MachineBasicBlock &MBB);
+
+  /// Find consecutive identical instructions, so we can remove the first and
+  /// replace it with the second.
+  bool tryReplaceOpWithCmp(MachineBasicBlock &MBB);
```
akiramenai: This one looks very generic in potential. We narrow it to subs, but it principle given no side effects any two consecutive identical instructions are to be removed. Doesn't LLVM's peephole optimizer do so? Can we use it?

akiramenai (changes requested):
I haven't finished reading (and I'm continuing) but as my comment requires investigation, I post it now. I suspect LLVM must have functionality to remove identical instructions. @vladimirradosavljevic can we use it instead of our own logic to detect identical `sub.s`es?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+bool EraVMOptimizeSelectCmpPreRA::isRegAdd(const MachineInstr &MI) const {
+  return MI.getOpcode() == EraVM::ADDrrr_s &&
+         TII->getCCCode(MI) == EraVMCC::COND_NONE;
+}
+
+bool EraVMOptimizeSelectCmpPreRA::isConstAdd(const MachineInstr &MI) const {
+  return MI.getOpcode() == EraVM::ADDcrr_s && EraVM::in0Iterator(MI)->isCPI() &&
+         (EraVM::in0Iterator(MI) + 1)->isImm() &&
+         !(EraVM::in0Iterator(MI) + 1)->getImm() &&
```
akiramenai: Why don't we support static offsets here?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -0,0 +1,513 @@
...
+        (!InAddRegs.count(EraVM::in0Iterator(UseMI)->getReg()) &&
+         !InAddRegs.count(EraVM::in1Iterator(UseMI)->getReg())))
+      continue;
+
+    UsesToUpdate.insert(&UseMI);
+    Deleted.emplace_back(&MI, &UseMI);
+  }
+
+  // 2. Combine.
+  for (auto [Add, Use] : Deleted) {
```
akiramenai: This is largely duplicated code. Wouldn't you mind to create a ticket to handle it?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -0,0 +1,513 @@
...
+    bool OutAddIsIn1Use = EraVM::out0Iterator(*Add)->getReg() ==
+                          EraVM::in1Iterator(*Use)->getReg();
+    auto CC = getImmOrCImm(*EraVM::ccIterator(*Use));
+    auto CCNewMI = OutAddIsIn1Use ? InverseCond[CC] : CC;
+    Register TieReg = OutAddIsIn1Use ? EraVM::in0Iterator(*Use)->getReg()
+                                     : EraVM::in1Iterator(*Use)->getReg();
+
+    // Create new instruction.
+    auto NewMI = BuildMI(*Use->getParent(), Use, Use->getDebugLoc(),
```
akiramenai: Could you try to lift `replace...` auxiliary functions from `EraVMCombineAddressingMode` to our `InstrInfo` header and apply it here (and to @TerryGuo part as well)?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
```diff
@@ -0,0 +1,513 @@
...
+    // It's expected that if there are more uses of add, it's very unlikely that
+    // all of them are select instruction where folding is feasible.
+    Register OutAddReg = EraVM::out0Iterator(MI)->getReg();
+    if (!RegInfo->hasOneNonDBGUser(OutAddReg))
+      continue;
+
+    MachineInstr &UseMI = *RegInfo->use_instr_nodbg_begin(OutAddReg);
+
+    // Skip use that is not in the same BB as add instruction. It's very
+    // unlikely that this instructions is select, where folding is feasible.
```
akiramenai: What makes us to restrict it anyway? We are doing the check below and the comment just says that we are to bail out because this check is likely to fail. Would it work if they are in different BBs?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: Will add it.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: Will do it.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: Yes, I missed `EraVMcmp` cases where we can match regs, `load_code` and `load_stack`. Will add them.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: We are doing only SUB instructions that were generated from CMP, since other identical instructions should be CSEd during LLVM IR optimizations and `ISel` phase.
Before I decided to implement it like this, I did investigate whether we can do use other LLVM passes, and here are my findings:
1. We can do this during `ISel` phase, but this can be bug prone since we are changing `cmp` with `add` instruction that can have more uses. I don't think this is a good place to do it, since for this to be profitable, we will need to call `ReplaceAllUsesOfValueWith` for add, and we can face the same issue as before.
2. In `MachineCSE` pass, but in this case, we would need to replace `cmp` instruction with `add`. This also could be bug prone, since add would need to set flags, and between `cmp` and `add` can be another instruction that sets flag.
3. In `PeepholeOptimizer::optimizeCmpInstr`, but this optimization is removing `cmp` in case it is not needed.

For these kind of optimizations where we are optimizing CMP or SELECT, in case there is no common pass that suits our needs (to take into account all cases of manipulating flags), I would prefer creating our own pass to do it.
Did I miss something?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: We are only generating memop with constant pool entries with offset 0, so I limited this check to zero offset.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: I will refactor out replacing code in this pass, and will check how it matches with `replace...` auxiliary functions from `EraVMCombineAddressingMode`. In case we need to spend more time in matching these functions, I will create a ticket for it. Wdyt?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: If you mean it is identical with code from `tryFoldSelectZero`, yes I will refactor out. In case you didn't mean that, could you please clarify it more?

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: I expect that LLVM will move code closer to it's use, so I added this restriction. I managed to create example where use is in another BB, but if I put it through the opt pipeline, it will move it just before it's use.
I can remove this restriction as this will work if use is in the different BB. Wdyt?

vladimirradosavljevic:
> I haven't finished reading (and I'm continuing) but as my comment requires investigation, I post it now. I suspect LLVM must have functionality to remove identical instructions. @vladimirradosavljevic can we use it instead of our own logic to detect identical `sub.s`es?

Thanks for the review! I added explanation why I chose to implement it this way in the comment.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
akiramenai: I think it's better to remove. It doesn't save us tons of compilation time while it increases maintenance burden.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
akiramenai: sgtm

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
akiramenai: This structure of transformation originated from post-RA RDA-based passes. If to grep `for (auto \[.*Deleted` or similar I expect you to find other places where this code is replicated. We probably can do better, though I don't insist doing it in the scope of this work.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
akiramenai: Ok, I found the cause of my confusion. It's the word "identical" in the doc and the methods names. They aren't. So it would be great to replace it with "foldable" or something because if one instruction sets flags and another one is not, they aren't identical or even equivalent (we shouldn't claim that `add` is identical to `sub`).

But as a general consideration
> since other identical instructions should be CSEd during LLVM IR optimizations and ISel phase

This is also imprecise. After isel we have pseudos that can be even if by coincidence become _foldable_ with an adjacent instruction. If one can specify in generic manner where to fold, the combining logic might be extracted away and generalised to work on any MIR (pre RA or post RA and any target). If it's missing in upstream, it can be upstreamed too.
It indeed sounds like more work than profit, but probably worth tracking. Wdyt?

akiramenai (approved):
Assuming "identical" will become less misleading, LGTM.

vladimirradosavljevic:
@akiramenai This is not the final version yet, I just want to check the CI. Will update you when it is ready for the review.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: As we agreed offline, I moved this optimization into the separate `EraVMFoldSimilarInstructions` pass and generalize implementation for the sub instructions. I have created the ticket (CPR-1544) to move this pass into the common code.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: I chose this approach because it is the safest solution where we won't ever run into issues with accessing invalidated iterator since here we are removing 2 and adding 1 instruction. Yes, this can't happen in this situation, but I wanted to make code more uniform throughout our backend.
Please note that I changed this approach to be similar with the `tryFoldSelectZero` implementation.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: I have refactored out replacing code, but it is different than in `replace...` auxiliary functions from `EraVMCombineAddressingMode`, since in one place we are mixing operands of 2 instructions, and in other we are using same operands.

llvm/lib/Target/EraVM/EraVMOptimizeSelectCmpPreRA.cpp
vladimirradosavljevic: Done.

vladimirradosavljevic:
@akiramenai PTAL.

llvm/lib/Target/EraVM/EraVMFoldSimilarInstructions.cpp
```diff
@@ -0,0 +1,231 @@
...
+  /// Return whether these two sub instructions are similar. Instructions
+  /// are similar iff their operands are identical, expect in the case
+  /// where we are trying to fold first sub with his flag setting pair
+  /// sub, and we are allowing for the second sub to have one more
+  /// implicit def of the flags register.
+  bool areSubsSimilar(const MachineInstr &FirstSub,
+                      const MachineInstr &SecondSub) const;
+
+  /// Return whether operands of two instructions are identical. In case
+  /// SkipSecondLastImplicitDef is true, we expect that the last implicit
```
akiramenai: Wouldn't it be better to name `IgnoreFlagSetting`.
In the implementation we can also check all implicit operands I believe instead of assuming the exact position.

llvm/lib/Target/EraVM/EraVMFoldSimilarInstructions.cpp
```diff
@@ -0,0 +1,231 @@
...
+      !EraVM::hasRROutAddressingMode(FirstSub) ||
+      !EraVM::hasRROutAddressingMode(SecondSub))
+    return false;
+
+  bool isSimilarToFlagSettingPair =
+      EraVM::getFlagSettingOpcode(FirstSub.getOpcode()) ==
+      static_cast<int>(SecondSub.getOpcode());
+  return (isSimilarToFlagSettingPair ||
+          FirstSub.getOpcode() == SecondSub.getOpcode()) &&
+         areOperandsIdentical(FirstSub, SecondSub, isSimilarToFlagSettingPair);
```
akiramenai, line 177: Looks like a formatting issue.

llvm/lib/Target/EraVM/EraVMFoldSimilarInstructions.cpp
```diff
@@ -0,0 +1,231 @@
...
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This file implements a pass that folds similar instructions.
+// Currently, it is implemented to fold similar consecutive instructions,
+// but this search can be expanded.
```
akiramenai: Also worth mentioning that it only work with subs and adds.

llvm/test/CodeGen/EraVM/fold-add-to-select.ll
```diff
@@ -0,0 +1,190 @@
...
+; CHECK-NEXT:  ret
+  %add = add i256 %a, %b
+  %cmp = icmp ult i256 %a, %b
+  %select = select i1 %cmp, i256 %b, i256 %add
+  ret i256 %select
+}
+
+define i256 @test_stack1(i256 %a) {
+; CHECK-LABEL: test_stack1
+; CHECK:       nop stack+=[1]
```
akiramenai: Redundant check as the test is not supposed to check frame structure and we already check it elsewhere.

akiramenai (approved):
LGTM.
My initial thoughts around `EraVMFoldSimilarInstructions` were to make it less idiosyncratic and more abstract. `TII` is to tell when to fold, and the pass is to fold in otherwise target agnostic way.
But the more I think about it the more I understand that our design on flags in many cases is incompatible with mainstream architectures and the aforementioned pass is about optimising flag setting sequence.
Yet name of the pass now implies some generality while it's not here and it's unlikely to be needed. `FoldFlagSettingSequence` would be a good alternative but similar is already used.

llvm/lib/Target/EraVM/EraVMFoldSimilarInstructions.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMFoldSimilarInstructions.cpp
vladimirradosavljevic: Changed the name to `IgnoreFlagSetting` and removed the assumption that implicit flags reg is the last operand.

llvm/lib/Target/EraVM/EraVMFoldSimilarInstructions.cpp
vladimirradosavljevic: clang-format did this change, and it looks good to me because indentation in brackets looks good and first line after the bracket is aligned to the open bracket.

llvm/test/CodeGen/EraVM/fold-add-to-select.ll
vladimirradosavljevic: Removed frame structure checks.

vladimirradosavljevic:
> LGTM. My initial thoughts around `EraVMFoldSimilarInstructions` were to make it less idiosyncratic and more abstract. `TII` is to tell when to fold, and the pass is to fold in otherwise target agnostic way. But the more I think about it the more I understand that our design on flags in many cases is incompatible with mainstream architectures and the aforementioned pass is about optimising flag setting sequence. Yet name of the pass now implies some generality while it's not here and it's unlikely to be needed. `FoldFlagSettingSequence` would be a good alternative but similar is already used.

Let's rethink about this pass when we start working on the overflow reverts.

PR #389, opened 2024-02-12
[EraVM] Add optimized memmove

vladimirradosavljevic:


llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
```diff
@@ -36,15 +37,300 @@ class EraVMLowerIntrinsics : public ModulePass {
...
+
+char EraVMLowerIntrinsics::ID = 0;
+
+INITIALIZE_PASS(EraVMLowerIntrinsics, DEBUG_TYPE, "Lower intrinsics", false,
+                false)
+
+// Lower memmove to IR. memmove is required to correctly copy overlapping memory
+// regions; therefore, it has to check the relative positions of the source and
+// destination pointers and choose the copy direction accordingly.
+static void createEraVMMemMoveLoop(Instruction *InsertBefore, Value *SrcAddr,
```
akiramenai, line 57: Shouldn't we put it to `Utils`? EVM can probably use it as well.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
```diff
@@ -36,15 +37,300 @@ class EraVMLowerIntrinsics : public ModulePass {
...
+                                         ResidualBits, "dst-mask");
+  Value *DstMasked = ResBuilder.CreateAnd(DstLoad, DstMask, "dst-masked");
+  Value *StoreElement =
+      ResBuilder.CreateOr(SrcMasked, DstMasked, "store-element");
+  ResBuilder.CreateAlignedStore(StoreElement, DstResPhi, PartDstAlign,
+                                DstIsVolatile);
+  ResBuilder.CreateBr(ExitBB);
+}
+
+// TODO: CPR-1380 Move this function to a common code.
```
akiramenai: 'a' is not needed.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
```diff
@@ -36,15 +37,300 @@ class EraVMLowerIntrinsics : public ModulePass {
...
+  // Create an empty preheader BB for copy-forward-loop since in this loop
+  // we can use indexed instructions and EraVMIndexedMemOpsPrepare
+  // requires loops to be in a simplify form.
+  auto *CopyFwdLoopPreheaderBB =
+      BasicBlock::Create(Ctx, "copy-forward-loop-preheader", F, ExitBB);
+  auto *CopyFwdLoopBB = BasicBlock::Create(Ctx, "copy-forward-loop", F, ExitBB);
+  // Create an empty cleanup BB for copy-forward-loop since in this loop
+  // we can use indexed instructions and EraVMIndexedMemOpsPrepare
+  // requires loops to be in a simplify form.
+  auto *CopyFwdLoopCleanupBB =
```
akiramenai: I suggest to use `LoopExitingBB` as it describes the loop structure and set expectations.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
```diff
@@ -36,15 +37,300 @@ class EraVMLowerIntrinsics : public ModulePass {
...
+      BasicBlock::Create(Ctx, "copy-forward-loop-cleanup", F, ExitBB);
+  auto *CopyFwdRBB =
+      BasicBlock::Create(Ctx, "copy-forward-residual", F, ExitBB);
+  auto *ResBB = BasicBlock::Create(Ctx, "memmove-residual", F, ExitBB);
+
+  // Remove terminator after splitBasicBlock.
+  OrigBB->getTerminator()->eraseFromParent();
+
+  // Create conditional branch in the original BB based on the ptr compare.
```
akiramenai: Nitpicking. 
Despite these comments describe IR-level entity names they are largely obvious while create significant line noice. If to make names like `CompareLC`, `CompareRB` more explicit referencing `loop-bytes-count`, `residual-bytes` become redundant.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
```diff
@@ -0,0 +1,177 @@
+; RUN: opt -passes=eravm-lower-intrinsics -S < %s | FileCheck %s
+; RUN: llc -O3 < %s | FileCheck --check-prefix=CHECK-INSTRS %s
+
+target datalayout = "E-p:256:256-i256:256:256-S32-a:256:256"
+target triple = "eravm"
+
+declare void @llvm.memmove.p1.p1.i256(ptr addrspace(1), ptr addrspace(1), i256, i1 immarg)
+
+define i256 @test_unknown(ptr addrspace(1) %dst, ptr addrspace(1) %src, i256 %size) {
```
akiramenai, line 9: We IRGen memmove expansion and then we write a test that checks that we IRGened it as expected. Why didn't we write it in IR in the first place. There aren't probably good APIs, but I wonder if it's a concern to somebody in mainline?
Another thought is that user-defined literals might be of help when somebody needs to write an IR generating code.
Writing to discuss it only.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
```diff
@@ -0,0 +1,177 @@
...
+; CHECK-NEXT:    [[DST_MASKED:%.*]] = and i256 [[DST_LOAD]], [[DST_MASK]]
+; CHECK-NEXT:    [[STORE_ELEMENT:%.*]] = or i256 [[SRC_MASKED]], [[DST_MASKED]]
+; CHECK-NEXT:    store i256 [[STORE_ELEMENT]], ptr addrspace(1) [[DST_RES_ADDR]], align 1
+; CHECK-NEXT:    br label [[MEMMOVE_DONE]]
+; CHECK:       memmove-done:
+; CHECK-NEXT:    ret i256 0
+
+; CHECK-INSTRS-LABEL: test_unknown:
+
+; Preheader and backwards loop.
```
akiramenai, line 77: I don't feel it's necessary to thoroughly check asm if we've already made sure that IR is correct. It's also very fragile checks.
We might want to check we indeed use indexed operations in the forward loop. And I'd add `p3` tests for fat pointers as well. But only specific trait to be checked, imo.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
```diff
@@ -0,0 +1,177 @@
...
+; CHECK-INSTRS-NEXT:  shl r2, r3, r2
+; CHECK-INSTRS-NEXT:  or r2, r4, r2
+; CHECK-INSTRS-NEXT:  st.1 r1, r2
+;
+entry:
+  tail call void @llvm.memmove.p1.p1.i256(ptr addrspace(1) %dst, ptr addrspace(1) %src, i256 %size, i1 false)
+  ret i256 0
+}
+
+; Test that we called memcpy implementation.
```
akiramenai, line 121: Don't we have tests that check memov -> memcpy in another file. We also have tests for memcpy expansion if I remember it right. So do we need to check asm here?

akiramenai (approved):
LGTM.
Improvements possible here.
I'd also prefer to avoid having redundancy in tests. It's a generaliration over several last reviews. Ideally if something goes wrong one tests fails and indicates what's wrong, so ideally a test is to check only what it intends to check and nothing more. It's especially true if the checks rely on fragile things like specific registers allocation that might change.
But it looks like you prefer to be very explicit in tests so do you think otherwise?

vladimirradosavljevic:
> LGTM. Improvements possible here. I'd also prefer to avoid having redundancy in tests. It's a generaliration over several last reviews. Ideally if something goes wrong one tests fails and indicates what's wrong, so ideally a test is to check only what it intends to check and nothing more. It's especially true if the checks rely on fragile things like specific registers allocation that might change. But it looks like you prefer to be very explicit in tests so do you think otherwise?

Yes, I prefer adding all checks, as from my experience they can catch issues much earlier in some cases. Example is when you are making optimization that doesn't have anything to do with this test, but it turns out it changed something, so you need to investigate why that change happened and to see if that change is correct or not. With this approach, we are actually expanding our testing coverage. This is something that I saw from the RISCV developers, and made me think that it is a correct approach to spot issues as fast as possible. This comes with an overhead in size of the tests, because they will be huge, but I think it is a small overhead we need to pay in the end. Also, creating and changing these kind of tests is not an overhead to a developer (it will actually speed up creating tests), since there are 2 python scripts that are used for this: update_test_checks.py (for opt tests) and update_llc_test_checks.py (for llc tests). For the latter, we would need to add support for our backends, and it is not a big problem to do so.
Does this make sense to you? If you think it is not important, I will start making tests that are adding only intended checks.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
vladimirradosavljevic: EVM has MCOPY instruction and should generate it for memcpy and memmove. Am I missing something?

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
vladimirradosavljevic: Will remove.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
vladimirradosavljevic: Will change the name.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
vladimirradosavljevic: Will change the names and remove comments.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
vladimirradosavljevic: Since there is a script to generate these checks (update_test_checks.py), you just need to call it for this test and to check what is generated.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
vladimirradosavljevic: I have added asm checks to see generated code, since there are a lot of passes that we are running after this, and I found an issue in `SeparateConstOffsetFromGEP` pass, so it is handy to check the generated code for key BBs. Also, it is good to check cases where src and dst are known, to see whether LLVM is able to remove unnecessary BBs and to optimize the code. 
Could you please clarify why we need `p3` tests? Since we can't store to fat pointers, it is not possible to call memmove for p3.p3, and if src and dst are in different address spaces, memcpy will be called.

llvm/lib/Target/EraVM/EraVMLowerIntrinsics.cpp
akiramenai: It's me who missed. Thank you.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
vladimirradosavljevic: This is different than memmove -> memcpy tests, since that optimization will be done if src and dst don't overlap. Here, we are testing `isCopyForward` function, since memmove implementation for forward copy is essentially memcpy implementation.

llvm/test/CodeGen/EraVM/memmove-expansion.ll
akiramenai: Thanks for explaination.

akiramenai:
Perhaps we are to highlight intendend checks then to make tests more readable. I still find it valuable even if for the purpose of documentation and review to have concise tests that check one thing only (e.g. incremental operations). While in the same time as you have practical examples and we have scripts to reduce the maintenance burden, I agree that we should try keeping thorough tests as well. Just maybe it's better to not to combine them together.
Another idea came from Kevin Henney is to name tests as propositions e.g. `forward_memmov_must_use_incremental_ldst`. It might look a bit exuberant but it also helps to check the test itself against the stated intent.
Wdyt?

vladimirradosavljevic:
Your suggestion is to start using scripts to add all checks, and to separate tests into more files with meaningful names for the future PRs? I totally agree if this is what you meant, and will add support for our backend in update_llc_test_checks.py next week.

akiramenai:
My suggestions were
1. To have 2 types of tests: auto generated and for specific functionality only.
2. Specific tests should be minimal in terms of number of checks and should have clear names that express intent of a test so a reviewer can check what is stated in the name with what `CHECK`s really do.

I didn't ask to not to mix them it in a single file more to at least to not to mix them in a single test function, but keeping them separate on files level is a good idea too.

vladimirradosavljevic:
> My suggestions were
> 
> 1. To have 2 types of tests: auto generated and for specific functionality only.
> 2. Specific tests should be minimal in terms of number of checks and should have clear names that express intent of a test so a reviewer can check what is stated in the name with what `CHECK`s really do.
> 
> I didn't ask to not to mix them it in a single file more to at least to not to mix them in a single test function, but keeping them separate on files level is a good idea too.

I agree with this approach with one adjustment as in some cases where we are doing bug fixes or optimizations and tests are not that big, we can only go with auto generated approach with pre-committing tests, so reviewer can check what exactly has changed. Does this make sense to you?

PR #390, opened 2024-02-13
[EVM] Back End

PavelKopyl:
# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


llvm/include/llvm/IR/IntrinsicsEVM.td
```diff
@@ -0,0 +1,210 @@
...
+def int_evm_byte
+  : Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty],
+              [IntrNoMem]>;
+
+// Memory operations.
+def int_evm_mstore8
+  : Intrinsic<[], [LLVMQualPointerType<llvm_i8_ty, AS.HEAP>, llvm_i256_ty],
+              [IntrWriteMem, IntrHasSideEffects]>;
+
+def int_evm_msize : Intrinsic<[llvm_i256_ty], [], []>;
```
akiramenai, line 78: Similar to what we have done in EraVM we should consider IntrInaccessibleMemOnly [+ IntrNoMerge]. Could we create a ticket for that and optionally add a TODO?

llvm/include/llvm/IR/IntrinsicsEVM.td
```diff
@@ -0,0 +1,210 @@
...
+
+def int_evm_calldatasize : Intrinsic<[llvm_i256_ty], [], [IntrNoMem]>;
+
+def int_evm_calldataload
+  : Intrinsic<[llvm_i256_ty], [LLVMQualPointerType<llvm_i256_ty, AS.CALL_DATA>],
+              []>;
+
+def int_evm_codesize : Intrinsic<[llvm_i256_ty], [], [IntrNoMem]>;
+
+// TODO: Ensure the gasprice is fixed while a function execution.
```
akiramenai, line 104: @hedgar2017 do you know if it is the case?

llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
```diff
@@ -7188,7 +7188,9 @@ SDValue SelectionDAG::getMemcpy(SDValue Chain, const SDLoc &dl, SDValue Dst,
   // Check to see if we should lower the memcpy to loads and stores first.
   // For cases within the target-specified limits, this is the best choice.
   ConstantSDNode *ConstantSize = dyn_cast<ConstantSDNode>(Size);
-  if (ConstantSize) {
+  // EVM local begin
```
akiramenai: Could we explain this local change? It looks suspicious.

llvm/lib/Support/Triple.cpp
```diff
@@ -173,11 +176,14 @@ StringRef Triple::getArchTypePrefix(ArchType Kind) {
 
   case loongarch32:
   case loongarch64: return "loongarch";
-  
+
```
akiramenai: Probably you reformatted too much or something.

llvm/lib/Target/EVM/EVM.h
```diff
@@ -0,0 +1,69 @@
...
+void initializeEVMAllocaHoistingPass(PassRegistry &);
+void initializeEVMLowerIntrinsicsPass(PassRegistry &);
+void initializeEVMArgumentMovePass(PassRegistry &);
+void initializeEVMLinkRuntimePass(PassRegistry &);
+void initializeEVMOptimizeLiveIntervalsPass(PassRegistry &);
+void initializeEVMRegColoringPass(PassRegistry &);
+void initializeEVMSingleUseExpressionPass(PassRegistry &);
+void initializeEVMStackifyPass(PassRegistry &);
+
+struct EVMLinkRuntimePass : PassInfoMixin<EVMLinkRuntimePass> {
```
akiramenai, line 66: Could we create a ticket to get rid of this duplication?

akiramenai:
@vladimirradosavljevic we need to put license to every file. Am I right?

llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
PavelKopyl: Sure.
For EVM all the memcpy, memmove intrinsics should be lowered into appropriate machine instructions. There is one-to-one correspondence (unlike of EraVM where some of them are expanded into loops). For this, I overrided SelectionDAGTargetInfo::EmitTargetCodeForMemcpy, EmitTargetCodeForMemmove. I need these methods to always be invoked no matter if the memory size of constant or not, see line 72098. But the logic of SelectionDAG::getMemcpy (getMemmove) first checks if the size is constant and if it is, tries to expand the intrinsic into a sequence of plain LD/ST. This change disables this behavior for EVM.

vladimirradosavljevic:
> @vladimirradosavljevic we need to put license to every file. Am I right?

Yes, we should always put the licence at the beginning of new files.

llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
vladimirradosavljevic: Sorry for jumping in, but can we avoid these local changes by creating pre-isel pass (e.g. EVMCodeGenPrepare) to convert memcpy and memmove calls to the EVM specific intrinsics? Just an idea to think about.

vladimirradosavljevic: I checked the implementations of `getMemcpyLoadsAndStores` and `getMemmoveLoadsAndStores` and it seems to me that if we override `findOptimalMemOpLowering` and return false for memcpy and memmove, it will always end up calling EmitTargetCodeForMemcpy or EmitTargetCodeForMemmove, so we don't need pre-isel pass or these local changes.

PavelKopyl: > Sorry for jumping in, but can we avoid these local changes by creating pre-isel pass (e.g. EVMCodeGenPrepare) to convert memcpy and memmove calls to the EVM specific intrinsics? Just an idea to think about.

Thank you, Vladimir! That's a nice way to get rig of this change.

PavelKopyl: > I checked the implementations of `getMemcpyLoadsAndStores` and `getMemmoveLoadsAndStores` and it seems to me that if we override `findOptimalMemOpLowering` and return false for memcpy and memmove, it will always end up calling EmitTargetCodeForMemcpy or EmitTargetCodeForMemmove, so we don't need pre-isel pass or these local changes.

Thank you. That would work, but as I see we well get an assertion when calling:
```
    SDValue Result = getMemcpyLoadsAndStores(
        *this, dl, Chain, Dst, Src, ConstantSize->getZExtValue(), Alignment,
        isVol, false, DstPtrInfo, SrcPtrInfo, AAInfo, AA);
```
because of 
` ConstantSize->getZExtValue()`,

when ConstantSize takes more that 64 bits. This case is difficult to imagine, but anyway..

vladimirradosavljevic: Ah, makes sense why we should avoid this approach. Thanks for looking into it 🙏

akiramenai: So are we going to proceed with a pass?

PavelKopyl: I guess so, especially if such changes in a common code are highly undesirable, but this will require more code.

llvm/test/CodeGen/Generic/2013-03-20-APFloatCrash.ll
```diff
@@ -1,3 +1,6 @@
+; UNSUPPORTED: evm
```
akiramenai, line 1: eravm is not supported too. I wonder how does it compile.

llvm/test/CodeGen/Generic/multiple-return-values-cross-block-with-invoke.ll
```diff
@@ -1,3 +1,6 @@
+; UNSUPPORTED: evm
+; EVM has no support of returning multiple values from a call
```
akiramenai: Is it the matter of not yet implemented calling conventions or is it a VM restriction?

llvm/test/CodeGen/EVM/logical.ll
```diff
@@ -0,0 +1,40 @@
+; RUN: llc --mtriple=evm --evm-keep-registers < %s | FileCheck %s
```
akiramenai: nitpicking: bitwise, not logical.

llvm/lib/Target/EVM/EVMRegColoring.cpp
```diff
@@ -0,0 +1,175 @@
...
+bool EVMRegColoring::runOnMachineFunction(MachineFunction &MF) {
+  LLVM_DEBUG({
+    dbgs() << "********** Register Coloring **********\n"
+           << "********** Function: " << MF.getName() << '\n';
+  });
+
+  // If there are calls to setjmp or sigsetjmp, don't perform coloring. Virtual
+  // registers could be modified before the longjmp is executed, resulting in
+  // the wrong value being used afterwards. (See <rdar://problem/8007500>.)
+  // TODO: Does EVM need to care about setjmp for register coloring?
```
akiramenai: No, afaik there aren't exceptions / error recovery mechanism within a contract.

llvm/lib/Target/EVM/EVMRegColoring.cpp
```diff
@@ -0,0 +1,175 @@
...
+      continue_outer:;
+      }
+
+    Register New = SortedIntervals[Color]->reg();
+    SlotMapping[I] = New;
+    Changed |= Old != New;
+    UsedColors.set(Color);
+    Assignments[Color].push_back(LI);
+    /*
+    // If we reassigned the stack pointer, update the debug frame base info.
```
akiramenai: ?

llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -0,0 +1,572 @@
...
+  // 4.3.1).
+  // This is similar to what we have in the ExpandIntRes_MUL() from
+  // LegalizeIntegerTypes.cpp.
+
+  SDValue LHS = Op.getOperand(0);
+  SDValue RHS = Op.getOperand(1);
+
+  unsigned Bits = VT.getSizeInBits();
+  unsigned HalfBits = Bits >> 1;
+  SDValue Mask = DAG.getConstant(APInt::getLowBitsSet(Bits, HalfBits), DL, VT);
```
akiramenai, line 275: If we don't have executable tests on it yet, please create a ticket to write them.

llvm/lib/Target/EVM/EVMInstrInfo.cpp
```diff
@@ -0,0 +1,209 @@
...
+    BuildMI(&MBB, DL, get(EVM::JUMP)).addMBB(FBB);
+    ++InstrCount;
+  }
+
+  return InstrCount;
+}
+
+bool EVMInstrInfo::reverseBranchCondition(
+    SmallVectorImpl<MachineOperand> &Cond) const {
+  // TODO: Try to add support for branch reversing. The main problem is that
```
akiramenai: Do we have a ticket for that?

llvm/lib/MC/ELFObjectWriter.cpp
```diff
@@ -1065,7 +1065,12 @@ uint64_t ELFWriter::writeObject(MCAssembler &Asm, const MCAsmLayout &Layout) {
   std::map<const MCSymbol *, std::vector<const MCSectionELF *>> GroupMembers;
 
   // Write out the ELF header ...
-  writeHeader(Asm);
+  // EVM local begin
+  // HACK!!! For EVM target we don't need the whole EFL file,
```
akiramenai, line 1069: Is there a way to guard it against our target?
We would need something similar for EraVM. 
@hedgar2017 did it come from the fact that we want to ship a single binary? Is it going to change with disassembler and maybe other tools?

llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
akiramenai: Ok, a ticket is ok for now

llvm/lib/Target/EVM/EVM.td
```diff
@@ -0,0 +1,48 @@
+//===---- EVM.td - Describe the EVM Target Machine -----*- tablegen -*-----===//
+//
```
akiramenai, line 2: license

llvm/lib/Target/EVM/EVMArgumentMove.cpp
```diff
@@ -0,0 +1,85 @@
+//===---------- EVMArgumentMove.cpp - Argument instruction moving ---------===//
+//
```
akiramenai, line 2: license

llvm/lib/Target/EVM/EVMAsmPrinter.cpp
```diff
@@ -0,0 +1,75 @@
+//===-------- EVMAsmPrinter.cpp - EVM LLVM assembly writer ----------------===//
+//
```
akiramenai, line 2: license

llvm/lib/Target/EVM/EVMFrameLowering.cpp
```diff
@@ -0,0 +1,17 @@
+//===-------- EVMFrameLowering.cpp - EVM Frame Information ----------------===//
+//
```
akiramenai, line 2: license

llvm/lib/Target/EVM/EVMFrameLowering.h
```diff
@@ -0,0 +1,33 @@
+//==----- EVMFrameLowering.h - Define frame lowering for EVM --*- C++ -*----==//
+//
```
akiramenai, line 2: license

llvm/lib/Target/EVM/EVMISD.def
```diff
@@ -0,0 +1,17 @@
+//----------------- EVMISD.def - EVM ISD ---------------------------*- C++ -*-//
+//
```
akiramenai, line 2: license

akiramenai (changes requested):
Could we make sure the license is in every file we created?

llvm/lib/Target/EVM/EVMSingleUseExpression.cpp
```diff
@@ -0,0 +1,802 @@
...
+  auto NextI = std::next(MachineBasicBlock::const_iterator(DefI));
+  for (auto E = MBB->end(); NextI != E && NextI->isDebugInstr(); ++NextI)
+    ;
+  if (NextI == Insert)
+    return true;
+
+  /*
+  // 'catch' and 'catch_all' should be the first instruction of a BB and cannot
+  // move.
+  if (EVM::isCatch(DefI->getOpcode()))
```
akiramenai: ?

llvm/test/CodeGen/Generic/2013-03-20-APFloatCrash.ll
PavelKopyl: Unlike EVM, EraVM BE calls GlobalDCEPass that removes completely this function (because it has 'internal' visibility and is not called by any other functions in the module).
For EVM gets crashed inside MC (on an assertion), because of an undefined symbol. This symbol corresponds to an FP math library call.

llvm/test/CodeGen/Generic/multiple-return-values-cross-block-with-invoke.ll
PavelKopyl: Thank you, I removed this restriction. Now EVM BE can compile this.

llvm/test/CodeGen/EVM/logical.ll
PavelKopyl: Thank you:)

llvm/lib/Target/EVM/EVMRegColoring.cpp
PavelKopyl: Thanks.

llvm/lib/Target/EVM/EVMRegColoring.cpp
PavelKopyl: Removed this.

llvm/lib/Target/EVM/EVMISelLowering.cpp
PavelKopyl: There are some tests from compiler-tester that required multiplication of i512 types.

llvm/lib/Target/EVM/EVMInstrInfo.cpp
PavelKopyl: Not yet. Created.

llvm/include/llvm/IR/IntrinsicsEVM.td
PavelKopyl: Added TODO with the ticket.

llvm/lib/Support/Triple.cpp
PavelKopyl: Yep, thank you.

llvm/lib/Target/EVM/EVM.h
PavelKopyl: Added.

llvm/lib/CodeGen/SelectionDAG/SelectionDAG.cpp
PavelKopyl: I've already added a final solution for this, please see changes in EVMCodegenPrepare.cpp, EVMISelLowering.cpp (LowerINTRINSIC_VOID) and IntrinsicsEVM.td.

llvm/lib/Target/EVM/TargetInfo/EVMTargetInfo.h
```diff
@@ -0,0 +1,27 @@
+//===-------- EVMTargetInfo.h - EVM Target Implementation -------*- C++ -*-===//
+//
```
akiramenai, line 2: license

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+// this means adding stack manipulation instructions (POP, PUSH, DUP or SWAP)
+// to the Machine IR and removing stack operands of instructions. Logically,
+// this corresponds to allocation of stack slots and mapping them on virtual
+// registers.
+//
+// Current implementation is based on the article M. Shannon, C. Bailey
+//   "Global Stack Allocation – Register Allocation for Stack Machines",
+// but in quite simplified form.
+//
+// The algorithm is quite simple and straightforward. It doesn't involve any
```
akiramenai: Nitpicking
Only the last sentence of this paragraph carries information. I'd even specify on needed optimizations.

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+// The algorithm runs, in outline, as follows (on a function level):
+//
+// 1. Determine content of the X-stack. Here we include all the virtual
+// registers, whose live intervals cross a BB boundary. Its content is fixed
+// during the whole function. See, allocateXStack()
+//
+// 2. Perform handling of each instruction. See handleInstrUses().
+//    As an example let's see how
+//
+//     %vreg7 = ADD %arg1 kill, %arg2 kill
```
akiramenai: %arg1 -> %vreg1
%arg2 -> %vreg2

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+//     SWAP5  // Swap top item with 6th (SWAP opcodes have shifted names)
+//     POP    // Remove top item from the stack
+//
+//   Note:
+//
+//     - resulted code has actually 25% of payload instructions. One the
+//       goals of stackification is to minimize the stack manipulation
+//       instructions.
+//
+//     - %vreg1 is dead after the instruction, so it theoretically should be
```
akiramenai, line 141: Can we avoid copying it from L-stack in the first place? I mean can we swap it or something?

PavelKopyl:
> Could we make sure the license is in every file we created?

I've just checked this via the following bash cmd-s:
Get the list of newly added files that are not tests and among them shows those that have no "Apache-2.0" string:
`git diff -U0 remotes/origin/dev-15 --name-only --diff-filter=A | grep -v test | xargs -nr -I{} grep -rL "Apache\-2\.0" {}`

I got the following list:
```
llvm/lib/Target/EVM/CMakeLists.txt
llvm/lib/Target/EVM/Disassembler/CMakeLists.txt
llvm/lib/Target/EVM/MCTargetDesc/CMakeLists.txt
llvm/lib/Target/EVM/TargetInfo/CMakeLists.txt
llvm/lib/Target/EVM/evm-stdlib.ll
```
These files do not need a license head (am I right?), which means the license was added to all the required files.

llvm/lib/Target/EVM/TargetInfo/EVMTargetInfo.h
PavelKopyl: Thank you, fixe. I've just run a one-line script to find files with missing license. This file and its .cpp one were that last two files with missing license.

llvm/lib/Support/Triple.cpp
```diff
@@ -173,11 +176,14 @@ StringRef Triple::getArchTypePrefix(ArchType Kind) {
...
```
akiramenai: Still here.

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+  void leaveBB(MachineBasicBlock *BB) {
+    // Clear L-stack.
+    assert(EStack.empty());
+    peelPhysStack(StackType::L, LStack.size(), BB, BB->end());
+  }
+
+  void dumpState() const;
+
+private:
+  StackLocation allocateStackModelLoc(const Register &Reg);
```
akiramenai, line 250: Documentation improvement is welcome here, but it's ok to postpone and create a ticket for that.

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+} // end anonymous namespace
+
+// FIXME: this needs some kind of hashing.
+std::optional<StackLocation>
+StackModel::getStackLocation(const Register &Reg) const {
+  auto It = std::find(LStack.begin(), LStack.end(), Reg);
+  if (It != LStack.end()) {
+    assert(std::count(LStack.begin(), LStack.end(), Reg) == 1);
+    unsigned Depth = std::distance(LStack.begin(), It);
+    return std::optional<StackLocation>{{StackType::L, Depth}};
```
akiramenai: Wouldn't `return {StackType::L, Depth}` work?

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+      continue;
+
+    const auto &Loc = getStackLocation(Reg);
+    assert(Loc);
+    RegLocs.push_back({Reg, *Loc});
+    LLVM_DEBUG(dbgs() << "Use: " << Reg << ", Depth: " << getPhysRegDepth(Reg)
+                      << '\n');
+  }
+
+  // Try to make a small optimization if MI has two args and both are
```
akiramenai: Could we avoid evaluative judgements in documentation (big, small, important, etc). They aren't informative while create line noice. `If both instruction arguments are on the top of stack, try ...`.
Ok to add to general doc improvement ticket I asked to create before.

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+    loadRegToPhysStackTop(Loc.first, MI);
+
+  if (NeedsSwap)
+    BuildMI(*MI->getParent(), MI, MI->getDebugLoc(), TII->get(EVM::SWAP1));
+
+  if (MI->getOpcode() == EVM::STACK_STORE)
+    handleStackStore(MI);
+
+  // Pop from the E-stack the registers consumed by the MI.
+  for (unsigned I = 0; I < RegLocs.size(); ++I)
```
akiramenai: We can erase `size()` elements instead of this loop.

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+
+  // Handle the registers for which we need to allocate X/L stack, performing
+  // migration E -> L/X stack.
+  for (const auto Reg : reverse(EStack)) {
+    assert(!getStackLocation(Reg));
+    allocateStackModelLoc(Reg);
+    EStack.pop_back();
+  }
+
+  const unsigned NumDeadRegs = std::count_if(
```
akiramenai, line 503: Better is to extract the logic below of swapping live and dead registers to a function. Name and document it properly.

llvm/lib/Target/EVM/EVMStackify.cpp
```diff
@@ -0,0 +1,1195 @@
...
+  for (const auto *It = StartUse; It != E; ++It)
+    migrateToEStack(It->first, It->second);
+
+  return StartUse;
+}
+
+bool StackModel::migrateTopRegToEStack(const Register &Reg, MachineInstr *MI) {
+  const auto Uses = MI->explicit_uses();
+  const unsigned NumUses =
+      std::count_if(Uses.begin(), Uses.end(), [Reg](const MachineOperand &MO) {
```
akiramenai, line 635: Nipicking
We can use `llvm::` counterpart for algorithms to use ranges.
(Or maybe we should go for C++20 if it works with LLVM ranges, as it looks like LLVM is going to switch soon anyway. @vladimirradosavljevic, @PavelKopyl wdyt?)

akiramenai (commented):
I'd say EVMStackify could be refactored and documented better. It's hard to read now. But let's not be blocked by that especially if we plan to improve stackification, we can move to better design gradually.
Licence problem is resolved. So can we create a ticket for EVMStackify and squash the commits so they reflect what we actually plan to land on top of `dev-15`? Then I approve.

llvm/lib/Target/EVM/EVMISelLowering.cpp
hedgar2017: Those are for EraVM only. EVM doesn't support 2-register output.

llvm/lib/MC/ELFObjectWriter.cpp
hedgar2017: Shipping a single binary is critical for UX. `solc` is coming in single binary, and the mere need of shipping `zksolc` in addition to it creates a lot of invariants for plugins, tooling, and project configs.

So we should put as much effort as possible to keep compilation a single command to a single executable.

akiramenai: Just to clarify to have a single binary and to have a single command for compilation are 2 different goals. `rustc` or even better `cargo` are examples of the latter w.o. the former.
Do we plan to have a separate binary for disassembler for EraVM or do you want to link it into `zksolc`? If the former, we are to break single binary shipment anyway.

hedgar2017: Good point.
I'd like to have everything linked to zksolc, that is to have a single binary with multiple commands for different purposes.

llvm/lib/Target/EVM/EVMISelLowering.cpp
PavelKopyl: Hm.. as I remember there was some tests from tests/solidity/simple, whose LLVM IR contained something like:
`%prode = mul i512 %arg1e, %arg2e`
 To lower this IR into the machine IR, I supported UMUL_LOHI primitive, which is used by Instruction Selection to represent multiplication of wider types than are natively supported by a target machine.
 But you are right, I also used eravm-stdlib.ll code, which contains `mul i512` operation, as a test for EVM.

llvm/lib/MC/ELFObjectWriter.cpp
PavelKopyl: Technically, I'll remove this particular change after finishing with EVM arch support in LLD
 linker (I'll move this functionality to LLD).
 
 I'd like to list some technical moments we need to have a single binary with a toolchain functionality:

-  extend assembler, disassembler and linker tools (in case of EVM) to have them as shared libraries.
    Only LLD has such a functionality out of a box (starting from llvm-17, I guess). All other modifications need to be implemented and maintained by ourselves.
-  Extend LLVM API to reflect new functionality
-  Extend and maintain `inkwell` package accordingly, as eventually all the toolchain functionality will be invoked from Rust.
-  Support new functionality in the `zksolc` cmd line.

@hedgar2017, do I understand it right?

llvm/lib/Target/EVM/EVMISelLowering.cpp
hedgar2017: i512 will never be generated for EVM.

For EraVM it is not a native type either, but `mul` can write output to two registers (low and high), although inputs are i256.

llvm/lib/MC/ELFObjectWriter.cpp
hedgar2017: @PavelKopyl yes, exactly.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: Sure, it will. Thank you.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: Removed that.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: Thanks, done.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: I've added this question to the ticket.

llvm/lib/Support/Triple.cpp
PavelKopyl: Fixed.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: Thanks, fixed.

llvm/include/llvm/IR/IntrinsicsEVM.td
hedgar2017: If it should've been `within a function`, yes. `gasprice` isn't changing within a transaction.

PavelKopyl: Thank you. So, the attributes are fine.

akiramenai (approved):
LGTM, thanks.

llvm/lib/Target/EVM/EVMSingleUseExpression.cpp
PavelKopyl: Removed that.

PavelKopyl:
> I'd say EVMStackify could be refactored and documented better. It's hard to read now. But let's not be blocked by that especially if we plan to improve stackification, we can move to better design gradually. Licence problem is resolved. So can we create a ticket for EVMStackify and squash the commits so they reflect what we actually plan to land on top of `dev-15`? Then I approve.

You are right, this version of stackification is just an intermediate step (mostly with draft ideas) towards more a sophisticated algorithm.
It allows to compile more-or-less complicated tests which gives a base line for perf, code size, and so on.

hedgar2017 (approved):

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: Fixed that.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: Sure, will do as a part of refactoring.

llvm/lib/Target/EVM/EVMStackify.cpp
PavelKopyl: If I got the question right, we can swap only the top stack item with an item in the depths.

llvm/lib/Target/EVM/EVMISelLowering.cpp
PavelKopyl: I've added a ticket to remove this functionality later (https://linear.app/matterlabs/issue/CPR-1567/remove-umul-lohi-primitive-from-isel)

vladimirradosavljevic:
All changes from `.github/workflows/compiler-integration-tests.yml` should be reverted before merging, right? What is the reason of adding `; UNSUPPORTED: evm` in EraVM tests? Is it because `lit.local.cfg` file is missing in EraVM folder, or something else?

PavelKopyl:
> All changes from `.github/workflows/compiler-integration-tests.yml` should be reverted before merging, right? What is the reason of adding `; UNSUPPORTED: evm` in EraVM tests? Is it because `lit.local.cfg` file is missing in EraVM folder, or something else?

You are right, I'll remove compiler-integration-tests.yml, as we are not going to support EVM BE in CI right now (though it would be better to support).

While working on EVM I used to build and run lit tests (`ninja check-llvm`) for both targets EraVM and EVM (in order not to cause regressions in EraVM).  I added `; UNSUPPORTED: evm` to some (those that do explicitly specify triple = eravm) EraVM test, otherwise they fail for EVM.  But we can remove that, as anyway EVM is not in CI.

vladimirradosavljevic:
We should also add config files for the EVM LIT tests.

PavelKopyl:
> We should also add config files for the EVM LIT tests.

Yep, thank you for recalling!
I've checked another build +  test config. It looks like `machine-outliner-tail.mir`, `machine-outliner.mir` need explicit triple specification.

PR #391, opened 2024-02-14
[SeparateConstOffsetFromGEP] Fix constant generation for negative AccumulativeByteOffset

vladimirradosavljevic:


vladimirradosavljevic:
Since there is no way to call `SeparateConstOffsetFromGEPPass` with `LowerGEP = true` from opt, I needed to create an asm test to reproduce the issue.
Change in `lowerToSingleIndexGEPs` function can't be tested, since it will be called for architectures where `useAA() == true`, which is false for us and we are calling `lowerToArithmetics` function instead. Please note that I made `useAA() == true` change locally for us just to test this function, and yes, the same issue was present there.

akiramenai:
Could you elaborate more on the issue? Is the fix upstreamable? If so you anyway need a description to upstream it.

vladimirradosavljevic:
> Could you elaborate more on the issue? Is the fix upstreamable? If so you anyway need a description to upstream it.

You can find more details in CPR-1551. Upstream ticket is CPR-1552.

llvm/test/CodeGen/EraVM/separate-const-offset-from-gep-bug.ll
```diff
@@ -0,0 +1,37 @@
...
+  %9 = load i256, ptr addrspace(1) %8, align 4
+  %10 = getelementptr inbounds i8, ptr addrspace(1) %0, i256 %7
+  store i256 %9, ptr addrspace(1) %10, align 4
+  %11 = icmp eq i256 %7, 0
+  br i1 %11, label %12, label %5
+
+12:
+  ret void
+}
+
```
akiramenai, line 35: Could we put a short description from CPR-1551 here? Otherwise `.cell 18446744073709551584` is surprising.

akiramenai (approved):

llvm/test/CodeGen/EraVM/separate-const-offset-from-gep-bug.ll
vladimirradosavljevic: Done.

PR #393, opened 2024-02-19
Updates workflows to make them public.

yorik:
fix: BFT-355

### DIFF
--- a/.github/workflows/compiler-benchmarks.yml
+++ b/.github/workflows/compiler-benchmarks.yml
@@ -59,17 +59,13 @@ jobs:
       - name: Checking out the compiler-tester repository
         uses: actions/checkout@v4
         with:
-          repository: matter-labs/compiler-tester
+          repository: matter-labs/era-compiler-tester
           ref: ${{ env.COMPILER_TESTER_BRANCH_NAME }}
           submodules: recursive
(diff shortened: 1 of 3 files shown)


akiramenai:
@yorik is it still needed? If so, can we update the PR, review and merge?

antonbaliasnikov:
Hi @yorik , if we change the `compiler-llvm` to `era-compiler-llvm`, this will break the testing, because the other repository will be tested, instead of this one. We cannot do that. We can change only `compiler-tester` repo here.

yorik:
This will be needed to make the repo public, due to circle dependency with `compiler-tester` repo it has to be done in one go.

I think the plan was to sync `compiler-tester` into `era-compiler-tester`, and rename this repo into `era-compiler-llvm`, and at this moment merge this PR.

Do you know when and how this is going to public?

akiramenai:
@yorik, @hedgar2017 said after we update LLVM to 17. Maybe it will happen next week as integration is what is left.

hedgar2017:
> @yorik, @hedgar2017 said after we update LLVM to 17. Maybe it will happen next week as integration is what is left.

Just noticed that this repo's workflows are not refactored yet.
@antonbaliasnikov could you take a look?

PR #401, opened 2024-02-23
[EraVM] Add __return_forward and __revert_forward stdlib functions

vladimirradosavljevic:


llvm/unittests/Analysis/TargetLibraryInfoTest.cpp
```diff
@@ -608,11 +608,12 @@ TEST_F(TargetLibraryInfoTest, ValidProto) {
       "declare i256 @__sar(i256, i256)\n"
       "declare i256 @__byte(i256, i256)\n"
       "declare void @__return(i256, i256, i256)\n"
+      "declare void @__return_forward(i256 addrspace(3)*)\n"
       "declare void @__revert(i256, i256, i256)\n"
       "declare void @__mstore8(i256 addrspace(1)* nocapture writeonly, i256)\n"
       "declare i256 @__sha3(i256 addrspace(1)*, i256, i1)\n"
       "declare i256 @__signextend(i256, i256)\n");
-      // EraVM local end
+  // EraVM local end
```
vladimirradosavljevic, line 617: FYI, this is clang-format change.

PavelKopyl: Yep, but it seems for me that clang-format does a strange formatting here, because "EraVM local begin" has another indentation.

PavelKopyl:
Am I right that the FE will emit these call instead of just calls to plain return/revert?

llvm/unittests/Analysis/TargetLibraryInfoTest.cpp
vladimirradosavljevic: I assume it is because `EraVM local begin` is inside of the expression, where `EraVM local end` is outside, so it is aligned to `parseAssembly`.
I think we have 3 options here:
1. Add clang-format off.
2. leave it as is.
3. Move bracket at the end, and it will look like this:
```
      "declare i256 @__sha3(i256 addrspace(1)*, i256, i1)\n"
      "declare i256 @__signextend(i256, i256)\n"
      // EraVM local end
  );
```
What do you prefer?

vladimirradosavljevic:
> Am I right that the FE will emit these call instead of just calls to plain return/revert?

I would assume it will emit all calls, depending on the case. @hedgar2017 Please correct me if I'm wrong.

llvm/unittests/Analysis/TargetLibraryInfoTest.cpp
PavelKopyl: Let's leave as is:)

hedgar2017:
Please don't merge until I approve after testing.

hedgar2017:
> > Am I right that the FE will emit these call instead of just calls to plain return/revert?
> 
> 
> 
> I would assume it will emit all calls, depending on the case. @hedgar2017 Please correct me if I'm wrong.

Yes, it is just another use case.

PavelKopyl (approved):
LGTM.

hedgar2017 (changes requested):
A few changes are needed:
1. Instead of setting the bit with bitwise ops, use `ptr.pack` intrinsic which sets the upper 128 bits of a fat pointer. The byte will be `96..104` if I'm not mistaken.
2. We cannot do any non-pointer operations with the pointer before `return`. Otherwise the pointer will be converted to integer and invalidated. If you need to move the pointer to `r1`, make sure you're using a pointer operation like `add.p`.
3. We could change the `return` signature to accept a pointer instead of an integer, but in that case we'll probably need an intrinsic for each address space. WDYT?

vladimirradosavljevic:
> A few changes are needed:
> 
> 1. Instead of setting the bit with bitwise ops, use `ptr.pack` intrinsic which sets the upper 128 bits of a fat pointer. The byte will be `96..104` if I'm not mistaken.
> 2. We cannot do any non-pointer operations with the pointer before `return`. Otherwise the pointer will be converted to integer and invalidated. If you need to move the pointer to `r1`, make sure you're using a pointer operation like `add.p`.
> 3. We could change the `return` signature to accept a pointer instead of an integer, but in that case we'll probably need an intrinsic for each address space. WDYT?

1. Makes sense. Since it will take upper 128 bit from the second non-fatptr input, I think we need to put the forwarding mode to 224...231 bits, so the constant will be the same as in the current patch.
2. Since `return` and `revert` IR intrinsics only accept integers, we would need to do `pack` + `ptrtoint`, but there is a tricky thing behind it. `ptrtoint` instruction is a no-op, but if it is a `ptrtoint` from AS3, we are generating `PTR_TO_INT` instruction in the backend that can generate normal `add` instruction if source and destination are not the same (I don't know the history behind this, and why it is needed).
Here is the example:
```
  %0:grptr = COPY $r1
  %1:grptr = PTR_PACKyrr_p %const.0, 0, %0:grptr
  %2:gr256 = PTR_TO_INT killed %1:grptr          <----- This can invalidate pointer.
  $r1 = COPY %2:gr256
  RETURN $r1
```
What we want is the following:
```
  %0:grptr = COPY $r1
  %1:grptr = PTR_PACKyrr_p %const.0, 0, %0:grptr
  $r1 = COPY %1:grptr          <----- This will create PTR_ADDrrr_s in case LLVM doesn't allocate %1 to r1
  RETURN $r1
```
The safest solution here is to create 2 new intrinsics (return.ptr and revert.ptr) that will accept pointer from AS3, so we don't need to worry whether for `PTR_TO_INT` we will get plain add and invalidate pointer.
3. As far as I understand, for pointers that are not from AS3, we can convert them to integers and do the bitwise operations, since we don't have instruction like .pack for other AS.

hedgar2017:
@vladimirradosavljevic
1. IIRC it accepts upper 128 bits in lower 128 bits, so it's `[96..104]`.
2. Maybe we should wait for @akiramenai's response here as he's worked the most with these pointer workarounds. However, since FE doesn't emit these intrinsics directly, I'd change all their arguments to pointers of various AS, similarly to `memcpy`. IIRC, these changes were benefical for the optimizer and we've adopted this approach throughout the standard library.
3. Pointers behave the same regardless of AS. Afaik, `gep`s on AS 1 and 2 are also lowered to `add.p`. AS 3 is only different in that it points to memory pages of other contracts, and we're using it for external calls and returns.

vladimirradosavljevic:
> @vladimirradosavljevic
> 
> 1. IIRC it accepts upper 128 bits in lower 128 bits, so it's `[96..104]`.
> 2. Maybe we should wait for @akiramenai's response here as he's worked the most with these pointer workarounds. However, since FE doesn't emit these intrinsics directly, I'd change all their arguments to pointers of various AS, similarly to `memcpy`. IIRC, these changes were benefical for the optimizer and we've adopted this approach throughout the standard library.
> 3. Pointers behave the same regardless of AS. Afaik, `gep`s on AS 1 and 2 are also lowered to `add.p`. AS 3 is only different in that it points to memory pages of other contracts, and we're using it for external calls and returns.

1. This is what I found in the documentation:
```
ptr.pack in1, in2, ou1
Concatenates the lower 128 bit of in1 and the higher 128 bits of in2, writes result to out.
2. Ensure the op1 is tagged as a pointer, and op2 is not tagged as a pointer. Otherwise panic.
3. Ensure that the lower 128 bits of op2 are zero. Otherwise panic.
```
If I'm reading correctly, we would need to do `llvm.eravm.ptr.pack(addr, i256 1 << 224)` to set the forwarding byte to 1.

2.  Does this mean we will change signatures of `__return` and `__revert` functions also to accept pointers? I'm asking this because for AS3 it is forbidden to use `inttoptr` (not sure about historical reasons for this), so we won't be able to create AS3 ptr from integers.
I'm not sure if optimizer will benefit from changing to ptr, since these are return instructions and we won't have other instructions below them to optimize.
3. I checked the code, and we are creating `ptr.` like instructions only for the AS3, others are using plain instructions to do the pointer arithmetics.

vladimirradosavljevic:
@hedgar2017 I updated implementations just to test them. Could you please run the tests with this?

hedgar2017:
@vladimirradosavljevic
1. Thanks, sorry for overlooking.
2-3. Ok, let's postpone if for now. Totally fine to keep as-is, but let's guarantee that only pointer operations are used before `ret`.

hedgar2017:
> @hedgar2017 I updated implementations just to test them. Could you please run the tests with this?

Sure, running now.

vladimirradosavljevic:
> @vladimirradosavljevic
> 
> 1. Thanks, sorry for overlooking.
>    2-3. Ok, let's postpone if for now. Totally fine to keep as-is, but let's guarantee that only pointer operations are used before `ret`.

1. No worries, and thanks for double check. 🙏 
2. In parallel, I did some experiments with `PTR_TO_INT` instruction, and saw that we can actually allocate the same physical register for src and dst registers, so `PTR_TO_INT` will become no-op. #406 change does that, so if we include that change into this, it is guaranteed that before `ret` will always be pointer operations.

vladimirradosavljevic:
Created the new return and revert intrinsics that accept ptr as3 input to guarantee that before `ret` will always be pointer operations.

hedgar2017:
> Created the new return and revert intrinsics that accept ptr as3 input to guarantee that before `ret` will always be pointer operations.

Great!
Let's wait for @akiramenai's review then.

akiramenai:
@vladimirradosavljevic @hedgar2017 I agree with the solution to add intrinsics for `fatptr` type. It seems consistent whith LLVM approach to intrinsics.
As for @vladimirradosavljevic  question about `PTR_TO_INT`. It was a part of Alan's design for `MVT::fatptr` because we need to inform LLVM about this type conversion as our arithmetic instructions don't accept `MVT::fatptr`. The alternative was to make it to accept.
We didn't spend time on optimizing this conversion but making it no op by any means makes sense. @sayon mentioned an attribute to inform that registers are to be coalesced but it might be not enough.
Will review the patch shortly.

akiramenai (approved):
LGTM, thanks

hedgar2017 (approved):

PR #408, opened 2024-02-28
[EraVM] Enable jump tables

vladimirradosavljevic:

PR #409, opened 2024-03-01
[EraVM] Implement isPredicated in EraVMInstrInfo

vladimirradosavljevic:


vladimirradosavljevic:
This is to fix issue in `MachineBlockPlacement` pass where it can wrongly reorder fallthrough BBs. This pass is calling `MachineBasicBlock::getFallThrough` and [here](https://github.com/matter-labs/compiler-llvm/blob/7f924f377ad5c0f79b6c840dd2a4513fbcddd516/llvm/lib/CodeGen/MachineBasicBlock.cpp#L947) if this is not implemented, it can return wrong result for the new jump table instruction (#408) after it gets predicated in IfConversion pass.

vladimirradosavljevic:
This fix is included in #408.

PR #425, opened 2024-03-07
Fix no asserts build

vladimirradosavljevic:
After https://github.com/matter-labs/era-compiler-llvm-builder/pull/15, build started to fail with unused variable errors. This PR fixes that.

### DIFF
--- a/llvm/lib/Target/EraVM/EraVMExpandPseudoInsts.cpp
+++ b/llvm/lib/Target/EraVM/EraVMExpandPseudoInsts.cpp
@@ -110,11 +110,12 @@ bool EraVMExpandPseudo::runOnMachineFunction(MachineFunction &MF) {
         Register FromReg = MI.getOperand(1).getReg();
         if (ToReg != FromReg) {
           LLVM_DEBUG(dbgs() << "Found PTR_TO_INT: "; MI.dump());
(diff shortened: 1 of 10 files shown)


vladimirradosavljevic:
Please note that reference benchmark will fail, since it doesn't contain this fix.

vladimirradosavljevic:
Added fixes for the EVM backend.

llvm/lib/Target/EraVM/EraVMExpandPseudoInsts.cpp
```diff
@@ -115,6 +115,7 @@ bool EraVMExpandPseudo::runOnMachineFunction(MachineFunction &MF) {
                            .addReg(FromReg)
                            .addReg(EraVM::R0)
                            .addImm(0);
+          (void)NewMI;
```
akiramenai: Could we add `maybe_unused` to the definition?

akiramenai (approved):

llvm/lib/Target/EraVM/EraVMExpandPseudoInsts.cpp
vladimirradosavljevic: Yes, that should also work. Done.

PavelKopyl:
LGTM.

PR #438, opened 2024-03-15
[EraVM] Remove setRequiresStructuredCFG

vladimirradosavljevic:


llvm/lib/Target/EraVM/EraVMInstrInfo.cpp
```diff
@@ -249,6 +249,11 @@ void EraVMInstrInfo::anchor() {}
 EraVMInstrInfo::EraVMInstrInfo()
     : EraVMGenInstrInfo(EraVM::ADJCALLSTACKDOWN, EraVM::ADJCALLSTACKUP), RI() {}
 
+unsigned int
+EraVMInstrInfo::getTailDuplicateSize(CodeGenOpt::Level OptLevel) const {
+  return 2;
```
akiramenai: Could we clarify?

akiramenai (approved):

llvm/lib/Target/EraVM/EraVMInstrInfo.cpp
vladimirradosavljevic: Yes, I will add comment. I tried with 2, 4, 6, 10, let me try with 3 to check the numbers.

vladimirradosavljevic: Tail duplication size 3 for -O3 reduced worst real time cycles from -1.316% to -0.375%, so I think we should keep it. Wdyt?

akiramenai: Ok, let's keep it. Please document size values we experimented with.

PR #443, opened 2024-03-19
[EraVM][NFC] Remove dead code in analyzeBranch

vladimirradosavljevic:


akiramenai:
Thanks for addressing this.
Was this code originally dead or it became dead after a patch? If the latter, could you add the patch to the commit message. I usually squish things like that when rebasing.

vladimirradosavljevic:
> Thanks for addressing this. Was this code originally dead or it became dead after a patch? If the latter, could you add the patch to the commit message. I usually squish things like that when rebasing.

It is the latter, and I think this became dead after #160.
Committed using `git commit --fixup 2e49bd18d745237832202a48446923158bd4bbdb` command.

akiramenai (approved):

PR #446, opened 2024-03-21
Suppress some clang-tidy errors in kpv-cpr-1568

FlashSheridan:
## Purpose

Suppress some clang-tidy errors in branch [kpv-cpr-1568-adding-support-of-evm-arch-to-lld-linker](https://github.com/matter-labs/compiler-llvm/tree/kpv-cpr-1568-adding-support-of-evm-arch-to-lld-linker); I’m proposing to add these changes to that branch while it‘s in development.  (Some of the debugging output in compiler-integration-tests.yml may not be suitable for main.)

## Ticket Number
CPR-1600

## Implementation
* Pass the googlemock and googletest paths on the `clang-tidy-diff_Zegar.py` command-line for otherwise apparently unsuppressible failures.
* Work around a clang-tidy bug in `lld/.clang-tidy` by having one positive check (clang-analyzer-osx.cocoa.MissingSuperCall, which is unlikely to conflict with anything.)
* Suppress all clang-tidy warnings in `lld/unittests`.
* Manually apply upstream updates to `clang-tidy-diff_Zegar.py`, and silence some Pylint (and probably runtime) warnings.
* Revert to specifying `clang-tidy-15` r/t generic `clang-tidy`.

This does not fix the unsuppressible breakage from “llvm/llvm/tools/llvm-driver/llvm-driver.cpp:20:10: error: 'LLVMDriverTools.def' file not found [clang-diagnostic-error]”, which is in Chris Bieneman’s code, not ours, and which I don’t understand the CMake setup well enough to address.  

There are remaining warnings in, for instance, `lld/MinGW/Driver.cpp`, many of them due to Alexandre Ganea; I hope that the `WarningsAsErrors: ''` in `lld/.clang-tidy` will prevent these from breaking the GitHub Action once the `[clang-diagnostic-error]`s are fixed.  (I agree with that we should ignore them rather than fix them.)  If not, or if there‘s too much noise, reverting e91ad1fac02a (“Remove the universal suppression from lld, now that it's in lld/unittests”) may help.


akiramenai:
No longer needed as we changed the implementation

Issue #435, opened 2024-03-13
[MC][AsmParser] Parse symbolic stack and const references


atrosinenko:
Parse stack- and const-referring operands that mention `@symbol` instead of just using `reg+imm` for address computation (factored out of issue #418).

asl:
Need tests, need examples

atrosinenko:
Based on the discussion in #418, parsing of the following syntax is implemented at now:
* `@sym[r1 + 1]`, `code[@sym + r1 + 1]` for code
* `stack[@sym + r1 + 1]`, `stack=[@sym + r1 + 1]` for **absolute** stack references
* not implemented for SP-relative and SP-modifying addressing modes
* for simplicity, in `code[...]` and `stack[...]` syntax, `@sym` is only parsed as the first term of the "subscript", the remainder is parsed as always (i.e. both `code[@sym + r1 + 1]` and `code[@sym + 1 + r1]` works, but not `code[1 + @sym]`)

Tests (links to the HEAD of `era-vm/asm-parser` branch):
* [`symbolic-operands.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/symbolic-operands.s)
* [`arith-operands-errors.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/arith-operands-errors.s)

akiramenai:
@atrosinenko, @hedgar2017, @sayon any suggestions on SP modifying syntax?

akiramenai:
Sorry, found in #417. It looks good to me.

asl:
@atrosinenko Please check and close

Issue #423, opened 2024-03-05
[MC][AsmParser] Parse misc instructions


asl:
- [x] `log.precompile in1, in2, out1`
 - [x] `context.this out` / `context.caller out` / `context.code_source out` / `context.ergs_left out` / `context.sp out` / `context.get_context_u128 out` / `context.set_context_u128 out` / `context.meta out` / `context.inc_tx_num out` / `context.set_ergs_per_pubdata in`

Spec:
 -  https://matter-labs.github.io/eravm-spec/spec.html#PrecompileCallDefinition
 - https://matter-labs.github.io/eravm-spec/spec.html#ContextDefinitions

atrosinenko:
From the spec:
```
Abstract Syntax: OpContextIncrementTxNumber
Assembly: context.inc_tx_num out
```

As far as I understand, `context.inc_tx_num` should be zero-operand instruction.

atrosinenko:
`log.precompile` seems to have `precompile` mnemonic in current backend implementation. Supporting `context.<op>` should require adding a bit more code in EraVMAsmParser.

atrosinenko:
Should work, but not all mnemonics are named as per the specification (see tests).

atrosinenko:
Works, except for three mnemonics having different name in the backend and the spec (see #441).

Tests (links to the HEAD of `era-vm/asm-parser` branch):
* [`misc.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/misc.s)
* [`mnemonic-errors.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/mnemonic-errors.s) - rejecting invalid `<op>` in `context.<op>` mnemonic

Note that `context.inc_tx_num` is parsed as a no-operand instruction (see [OpContextIncrementTxNumber](https://matter-labs.github.io/eravm-spec/spec.html#OpContextIncrementTxNumber) as well as "Affected parts of VM state: only tx counter") assuming `context.inc_tx_num out` syntax mentions `out` by mistake.

Issue #422, opened 2024-03-05
[MC][AsmParser] Parse event instructions


asl:
- [x] `log.event in1, in2` aliased as `event in1, in2`
 - [x] `log.event.first in1, in2` aliased as `event.i in1, in2`
 - [x] `log.to_l1 in1, in2` aliased as `event in1, in2`
 - [x] `log.to_l1 event.first in1, in2` aliased as `event.i in1, in2`

Spec:
 - https://matter-labs.github.io/eravm-spec/spec.html#EventDefinition
 - https://matter-labs.github.io/eravm-spec/spec.html#ToL1Definition

atrosinenko:
I rechecked the spec, and aliased instructions seem to be indeed the same...

atrosinenko:
Should work, but not all mnemonics are named as per the specification (see tests).

atrosinenko:
Works, but see #441.

Tests (links to the HEAD of `era-vm/asm-parser` branch):
* [`events.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/events.s)
* [`mnemonic-errors.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/mnemonic-errors.s) - rejecting invalid mnemonic suffixes

Issue #421, opened 2024-03-05
[MC][AsmParser] Parse storage instructions


asl:
- [x] `log.swrite in1, in2` aliased as `sstore in1, in2`
 - [x] `log.sread in1, out` aliased as `sload in1, out`

Spec:
  - https://matter-labs.github.io/eravm-spec/spec.html#SStoreDefinition
  - https://matter-labs.github.io/eravm-spec/spec.html#SLoadDefinition

Relevant tests:
* llvm/test/MC/EraVM/asm-parser/storage.s

atrosinenko:
Works.

atrosinenko:
Tests (links to the HEAD of `era-vm/asm-parser` branch):
* [`storage.s`](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/test/MC/EraVM/asm-parser/storage.s)

PR #448, opened 2024-03-28
Implement parsing EraVM assembler

atrosinenko:
This PR contains implementation of assembler parser for EraVM target as well as several refactorings made to backend's internal representation of EraVM instructions.

Related issues: #411, #412, #414, #416, #417, #418, #419, #420, #421, #422, #423, #435.


akiramenai:
It looks like there are plenty of functional and non-functional changes combined in this PR, so it's hard to review. Can we divide the functionality to meaningful commits and review them one by one? I mean divide to less than 114 commits; form a commit structure as we actually plan to land these changes.
Also please clang format and pass the linter.

asl:
> It looks like there are plenty of functional and non-functional changes combined in this PR, so it's hard to review. Can we divide the functionality to meaningful commits and review them one by one? I mean divide to less than 114 commits; form a commit structure as we actually plan to land these changes.

The commits itself are quite well-divided and represent fine-grained incremental set of changes, I would say there are almost no non-functional changes. However, we had to restructure the instructions definitions entirely as in the original form they were unsuitable for instruction encoding and had some inherent issues. Certainly, some commits could be combined, for example:

 - Support parsing name.s for commutative instructions (such as add.s)
 - Support parsing "name.s reg, reg, reg" for non-commutative instructions
 - Parse context.<cop> instructions
 - ...

Let me know if you'd want us to do this.

akiramenai:
@asl it sounds like a reasonable change. I think we should go for it.
Could we also make sure no diff `clang-format`, `clang-tidy` produce no warnings, and `check-llvm` pass for all commits?

asl:
> Could we also make sure no diff `clang-format`, `clang-tidy` produce no warnings, and `check-llvm` pass for all commits?

We are working on silencing linter warnings.

asl:
@hedgar2017 @akiramenai @sayon 

Looks like the integration tests fail because they use standalone assembler that cannot handle the assembler syntax described in the spec. How would you want to address this?

akiramenai:
@asl 
It looks like we miss a register as the first argument of `ret.to_label` if to satisfy our assembler. It should be easy to fix.
But in general, if I remember it right, we agreed to stick with the current syntax and made the changes requested by @hedgar2017, @sayon in separate patches. Which looks like a sane approach. So, with that in mind, passing the integration tests shouldn't be that hard, and the sooner we integrate, the better. We become confident in our asm printer this way.

asl:
@akiramenai Yes, the renaming is a part of https://github.com/matter-labs/compiler-llvm/issues/441

However, it looks like there are 3 different sources of information that do no correspond to each other: backend, assembler and the ISA spec. There were some obsolete instructions, some obsolete syntax, some undocumented syntax, etc. So, in such cases we had to resort to the existing spec as an ultimate source of information. Some changes (e.g. removal of obsolete stuff) is certainly a part of initial code drop as it does not make any sense to make them supported only to remove them immediately afterwards. 

In this particular case, it seems that standalone assembler does not support `revert` instruction alias as specified in ISA spec. We will add remained aliases to please the assembler, certainly :)

akiramenai:
@asl 
I understand that it's our fault. We haven't tested the spec against the assembler. So thank you for finding the discrepancy. We can raise an asm issue there and fix it. The fix could likely be landed on Monday. If you can do an alias, that could unblock you faster.
I still think it worth to integrate earlier. When we pass system contract compilation, the CI won't stop after the first failure.

akiramenai:
@asl I asked @sayon, [cc @shamatar]
But I believe the spec is wrong here. It doesn't mention the argument for `ret`, `revert` but to the best of my memory we pass the pointer to the returning slice, which makes sense if it's a far return or revert. But as we don't distinguish near and far in case of returns and reverts, we always pass an argument.
N.B. @sayon is OOO today.

hedgar2017:
@akiramenai @asl @sayon I'd kindly ask you guys to remove all aliases as soon as possible (e.g. the `revert` mentioned above). If it's just for the initial phase to reach conformance, that's fine.
It would be great to start over without legacy.

asl:
> @akiramenai @asl @sayon I'd kindly ask you guys to remove all aliases as soon as possible (e.g. the `revert` mentioned above). If it's just for the initial phase to reach conformance, that's fine.

Absolutely. However, here we are having chicken-and-egg problem: existing assembler supports only some aliases, but not others (e.g. `ret.revert` but not `revert`) or always require register for `ret.ok` regardless what is in the spec, even for default one, also it seems that it supports only some subset of specified assembler syntax (as it comes with stack operands, for example). So we'd need to add everything and then remove legacy :)

However, we structured the changes in such way so it will be more or less straightforward.

atrosinenko:
I pushed a few commits that (temporary?) rollback potentially breaking changes to syntax. Let's see if tests pass.

sayon:
I have pushed more updates to the the spec available online; now the returns are properly categorized:
- far return/revert and near return/revert are merged and became  `ret reg` and `rev reg`. Register argument is ignored if we return from near calls.
- returns/reverts/panics to label  are marked "near" because far returns ignore labels
- `pnc` (panic) does not accept arguments because panic ignores register argument

This type specifies the abstract syntax of the programmer-facing assembler instructions:
https://matter-labs.github.io/eravm-spec/spec.html#asm_instruction

asl:
@akiramenai What does "exec format error" error mean?

hedgar2017:
> > @akiramenai @asl @sayon I'd kindly ask you guys to remove all aliases as soon as possible (e.g. the `revert` mentioned above). If it's just for the initial phase to reach conformance, that's fine.
> 
> Absolutely. However, here we are having chicken-and-egg problem: existing assembler supports only some aliases, but not others (e.g. `ret.revert` but not `revert`) or always require register for `ret.ok` regardless what is in the spec, even for default one, also it seems that it supports only some subset of specified assembler syntax (as it comes with stack operands, for example). So we'd need to add everything and then remove legacy :)
> 
> However, we structured the changes in such way so it will be more or less straightforward.

Understood!
@akiramenai do you think we need to compare both assemblers for binary equality, or would a simple migration be sufficient? Assembly is low-level enough for our testset to walk all the paths, so I don't need to implement another comparison mode on compiler-tester, especially as a one-time job.

akiramenai:
@asl, I have never seen that before. But from the fact that only Viper programs are affected I infer that it's likely to be a CI issue. Could you rebase the work on ToT?
@hedgar2017 could you also have a look at the integration tests failures. You should know it better.

asl:
> @akiramenai do you think we need to compare both assemblers for binary equality, or would a simple migration be sufficient?

Here is an idea: we do have a set of assembler files (under `test/MC/EraVM`) to test the functionality. We can just run the existing assembler and`llvm-mc` and compare the outputs? Is there a quick way to dump the encoding from the existing assembler?

akiramenai:
@hedgar2017 we can migrate I think. Binary comparison is faster though, could have less dependence, and could identify issues that aren't touched by input data.

hedgar2017:
> @asl, I have never seen that before. But from the fact that only Viper programs are affected I infer that it's likely to be a CI issue. Could you rebase the work on ToT? @hedgar2017 could you also have a look at the integration tests failures. You should know it better.

I can only see LIT failures on CI. Could you point me directly?

hedgar2017:
> @hedgar2017 we can migrate I think. Binary comparison is faster though, could have less dependence, and could identify issues that aren't touched by input data.

Binary comparison isn't faster than simply running the CI after migration. I'll see if I can do some hacking like dump all builds from compiler-tester.

atrosinenko:
> I can only see LIT failures on CI. Could you point me directly?

@hedgar2017 Adjusted recently added `rol.ll` and `ror.ll` tests for the minor changes in SP-increment syntax, now `check-llvm` should pass.

atrosinenko:
These "Exec format error" messages look like something returned by `exec()` syscall to me - maybe something related to `binfmt-misc` configuration or trying to execute binaries for incorrect CPU architecture.

asl:
Now integration tests seem to pass. So likely was some random CI breakage. Please review

llvm/lib/Target/EraVM/AsmParser/EraVMAsmParser.cpp
```diff
@@ -182,32 +253,17 @@ class EraVMOperand : public MCParsedAsmOperand {
   }
 
   static std::unique_ptr<EraVMOperand> CreateToken(StringRef Str, SMLoc S) {
-    return std::make_unique<EraVMOperand>(Str, S);
+    return Create(k_Tok, S, S, [&](auto &Op) { Op.Tok = Str; });
```
akiramenai: Could we avoid capturing everything by reference where it's redundant? We only use `Str`, and it can be passed by value. Same applies to other lambdas.

llvm/lib/Target/EraVM/EraVMOpcodes.td
```diff
@@ -0,0 +1,197 @@
+//===-- EraVMInstrFormats.td - EraVM Instruction Formats ---*- tablegen -*-===//
```
sayon: Please update the filename/description in comment.

llvm/lib/Target/EraVM/EraVMExpandPseudoInsts.cpp
```diff
@@ -85,8 +86,10 @@ bool EraVMExpandPseudo::runOnMachineFunction(MachineFunction &MF) {
         auto func_name = func_opnd->getName();
         if (func_name == "__cxa_throw") {
           BuildMI(*MI.getParent(), &MI, MI.getDebugLoc(),
-                  TII->get(EraVM::THROW)).addReg(EraVM::R1)
-            .copyImplicitOps(MI);
+                  TII->get(EraVM::THROW))
+              .addReg(EraVM::R1)
+              .addImm(0)
```
sayon: IIUC this is a predicate; if so, would using `COND_NONE` instead of magic numbers be preferable everywhere in the codebase?

In the current codebase there are no instruction building facilities that would properly decompose an instruction structurally, or compose it out of the parts, therefore `addImm(0)` is used both as a part of code/stack composed operand, and as a predicate. It may look confusing.

llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -44,125 +36,122 @@ def ToStack        : DestAddrModeValue<2>;
...
 def ToStackReg     : DestAddrModeValue<4>;
 
-// FlagSetting:
-// true (1) : `_v` instructions ; false (0) : `_s` instructions
-
 //===----------------------------------------------------------------------===//
-// Generate mapping table to relate non-flag setting instruction (`_s`) to flag-setting
-// instruction (`_v`)
-//
+// Instruction mapping tagles
```
akiramenai: tables?

llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -44,125 +36,122 @@ def ToStack        : DestAddrModeValue<2>;
...
+
+class AddrModeRel; // generic map of addressing mode of operands
+class RetAddrModeRel; // generic map of addressing mode of results
+
+
 def mapRRInputTo : InstrMapping {
   let FilterClass = "AddrModeRel";
-  let RowFields = ["BaseOpcode", "DestAddrMode", "FlagSetting"];
+  let RowFields = ["BaseOpcode", "DestAddrMode", "Silent"];
   let ColFields = ["OperandAM"];
```
akiramenai, line 105: I introduced that duplication because I had issues with `OperandAddrMode` as the ColField. As you refactor our `td`s anyway, may I ask if it's hard to get rid of `OperandAM`?

llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -173,440 +162,560 @@ def RFPanic          : RetFlags<0b0010>;
...
+  dag InOperandList  = ins;
+  dag OutOperandList = outs;
+  let Pattern        = pattern;
+  let Size           = 8;
+
+  let isCodeGenOnly = 1;
+  let isPseudo = 1;
+}
+
+class Encoding {
```
sayon, line 183: Thank you for separating the encoding from the logical instruction set!

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -369,796 +355,544 @@ def : Pat<(int_eravm_iflt imm16:$src0, imm16:$src1), (SELiir imm16:$src0, imm16:
...
+                   "$imm, $rs1, $rd0",
+                   [(set GR256:$rd0, (node imm16:$imm, GR256:$rs1))]>,
+                   FlagSetting, Swappable, AddrModeRel, RetAddrModeRel;
+  def xrr_s : Iir_r<opcode, Swap, PreserveFlags,
+                   (outs GR256:$rd0), (ins imm16:$imm, GR256:$rs1),
+                   "$imm, $rs1, $rd0",
+                   [(set GR256:$rd0, (node GR256:$rs1, imm16:$imm))]>,
+                   FlagSetting, Swappable, RetAddrModeRel;
+
+  // The four *_alias instructions below are defined to make disassembler parse
```
akiramenai: disassembler?

llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -173,440 +162,560 @@ def RFPanic          : RetFlags<0b0010>;
...
-           string asmstr,
-           list<dag> pattern>
-  : IBinaryS<opcode, as, swap_operands, silent, outs, ins, asmstr, pattern> {
-  bits<20> src0; // reg + imm
+class Isr_s<EraVMOpcode opcode,
+            mod_swap swap_operands, mod_set_flags silent,
+            dag outs, dag ins,
+            string asmstr,
+            list<dag> pattern>
+  // FIXME: SrcStackAbsolute is incorrect here, we need to be able to properly encode relative and post-increment
```
sayon: Could you elaborate? I am not sure I understood what is going on here.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -99,86 +100,92 @@ def EraVMTrap : SDNode<"EraVMISD::TRAP", SDTNone,
...
   return Imm.getZExtValue() < 2;
 }]> {
   let PrintMethod = "printFirstOperand";
 }
 
-def spadv_op : Operand<i256>, IntImmLeaf<i256, [{
-  return Imm.getZExtValue() < 2;
-}]> {
-  let PrintMethod = "printSPAdvanceOperand";
+// This operand class represents those jump targets described
```
sayon, line 131: Question (not for this PR): can/should we get more operand types, for code and stack in different addressing modes?

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -1172,253 +906,299 @@ def : Pat<(EraVMcmp GR256:$lhs, (load_code memaddr:$rhs)), (SUByrr_v memaddr:$rh
...
-                         "context$cop$cc\t$rs0", []>;
-def CTXr_se0 : IContextR<152, (outs), (ins context_se_op:$cop, cc:$cc),
-                         "context$cop$cc", []>;
+def CTXGasLeft : IContext<OpContextErgsLeft, (outs GR256:$rd0), (ins),
+                          "$rd0",
+                          [(set GR256:$rd0, (int_eravm_gasleft))]>;
+def CTXGetSp   : IContext<OpContextSp, (outs GR256:$rd0), (ins),
+                          "$rd0",
+                          [(set GR256:$rd0, (EraVMget_sp))]>;
+def CTXGetU128 : IContext<OpContextGetContextU128, (outs GR256:$rd0), (ins),
```
sayon, line 1094: This instruction is marked with `hasSideEffects` in the resulting tablegen, but it is basically outputting a constant value within a single contract execution instance. It can be cached, it can be read at any place and freely moved around etc.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -1172,253 +906,299 @@ def : Pat<(EraVMcmp GR256:$lhs, (load_code memaddr:$rhs)), (SUByrr_v memaddr:$rh
...
+                        [(set GR256:$rd0, (int_eravm_caller))]>;
+def CTXCodeSource : IContext<OpContextCodeAddress, (outs GR256:$rd0), (ins),
+                            "$rd0",
+                            [(set GR256:$rd0, (int_eravm_codesource))]>;
+def CTXMeta : IContext<OpContextMeta, (outs GR256:$rd0), (ins),
+                        "$rd0",
+                        [(set GR256:$rd0, (int_eravm_meta))]>;
+
 let hasSideEffects = 1 in {
+def CTXSetU128 : IContext<OpContextSetContextU128, (outs), (ins GR256:$rs0),
```
sayon, line 1081: I wonder if this instruction should be marked as `hasSideEffects`. The only constraint that we need to enforce is that SetU128 would be executed before the next far call in the trace.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMAsmBackend.cpp
```diff
@@ -179,3 +179,45 @@ MCAsmBackend *llvm::createEraVMMCAsmBackend(const Target &T,
                                             const MCTargetOptions &Options) {
   return new EraVMAsmBackend(STI, ELF::ELFOSABI_STANDALONE);
 }
+
+void EraVM::appendMCOperands(MCContext &Ctx, MCInst &MI, MemOperandKind Kind,
```
sayon: IMO we could use more types here nicely to get rid of the (situationally used) arguments. 
Like a struct and an overload of `appendMCOperands` for each argument type.

llvm/lib/Target/EraVM/EraVMInstrInfo.cpp
```diff
@@ -871,8 +878,7 @@ EraVMInstrInfo::getOutliningType(MachineBasicBlock::iterator &MBBI,
   }
 
   // Don't outline context.gas_left instruction.
-  if (MI.getOpcode() == EraVM::CTXr_se &&
-      getImmOrCImm(MI.getOperand(1)) == EraVMCTX::GAS_LEFT)
+  if (MI.getOpcode() == EraVM::CTXGasLeft)
```
sayon, line 902: Why do we omit the CTXr_se (did it become "gasleft" or was it removed?)

llvm/test/CodeGen/EraVM/intrinsic.ll
```diff
@@ -18,7 +18,7 @@ define i8* @stacksave() {
 define void @stackrestore(i8* %ptr) noinline optnone {
 ; CHECK: context.sp r2
 ; CHECK-NEXT: sub r1, r2, r1
-; CHECK-NEXT: nop stack+=[r1]
+; CHECK-NEXT: nop stack+=[0 + r1]
```
sayon, line 21: Why do we need to write the fullest form `[0 + reg]` here and in other tests?

llvm/test/CodeGen/EraVM/intrinsic.ll
```diff
@@ -391,7 +384,6 @@ declare i256 @llvm.eravm.this()
 declare i256 @llvm.eravm.caller()
```
sayon, line 384: Should we treat these in the same way as `gasleftr`?

llvm/test/MC/EraVM/asm-parser/arith-instructions-errors.s
```diff
@@ -0,0 +1,74 @@
...
+  add r1, r2, r3, r4
+  mul r1, r2, r3
+
+; commutative instructions cannot have swapped variants because
+; there are no corresponding opcodes for machine instructions
+  add.s     42, r2, r3
+  and.s     42, r2, r3
+  or.s      42, r2, r3
+  xor.s     42, r2, r3
+  mul.s     42, r2, r3, r4
```
sayon, line 30: Some additions:
- arith instructions can not have any other modifiers (like is first, heap variant selectors etc.)
- repetition: `add.s.s.s` should fail

llvm/test/MC/EraVM/asm-parser/arith-operands-errors.s
```diff
@@ -0,0 +1,64 @@
...
+
+; incorrect "subscript" index
+  add    stack[-r1], r1, r2
+  add    stack[r1 + r1], r1, r2
+  add    stack[1 + 2], r1, r2
+
+; symbolic operands (const)
+  add  @a[@b + 1], r2, r3
+; symbolic operands (SP-relative)
+  add  stack-[@global + 1 + r1], r2, r3
```
sayon, line 23: also `stack + [anything]`

llvm/test/MC/EraVM/asm-parser/events.s
```diff
@@ -0,0 +1,31 @@
+; RUN: llvm-mc -triple eravm -filetype=asm %s -o - | FileCheck %s
+
+  .text
+foo:
+
+; mnemonics
+  ; FIXME The spec mentions different mnemonic names
```
sayon: Syntax should be coherent now, from what I see. Please ping if the problem persists.

llvm/test/MC/EraVM/asm-parser/sp-changes.s
```diff
@@ -0,0 +1,30 @@
...
+
+  add stack-=[r1], r1, r2
+  add stack-=[42], r1, r2
+  add stack-=[r1+42], r1, r2
+  add r0, r1, stack+=[r1]
+  add r0, r1, stack+=[42]
+  add r0, r1, stack+=[r1+42]
+  add stack-=[r1+42], r1, stack+=[r1+42]
+
+  nop r0, r0, stack+=[r1+42]
```
sayon, line 14: There is no test for `nop` without any operands.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
asl: For this you'd need to split the instructions. So, you'd end with # of instruction blow-up.

llvm/lib/Target/EraVM/EraVMInstrInfo.cpp
asl: As I mentioned: we created separate instruction instead of "contex operand". So, there is no `CTXr_se` and similar instructions anymore (see .td file).

llvm/test/CodeGen/EraVM/intrinsic.ll
asl: This is just how the things are emitted for now. We have not added simpler "forms".

llvm/test/MC/EraVM/asm-parser/arith-instructions-errors.s
asl: @atrosinenko Please add cases

llvm/lib/Target/EraVM/EraVMInstrInfo.td
asl: Yep. But this is how it was previously (it was `CTXr_se` instruction that had `hasSideEffects = 1` set).

llvm/lib/Target/EraVM/EraVMInstrInfo.td
asl: Well, I believe we kept the pre-existing behavior. So, if some things could be relaxed, this could be done in a separate change.

llvm/test/CodeGen/EraVM/intrinsic.ll
sayon: thanks for clarifying!

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -993,7 +993,6 @@ def JCs: IJump<OpJump, SrcStackAbsolute, (ins stackin:$dest), "$dest", []>;
 let isTerminator = 1, isBranch = 1, isBarrier = 1 in {
 def J: Pseudo<(outs), (ins jmptarget:$dest), [(EraVMbrcc bb:$dest, 0)]>,
        PseudoInstExpansion<(JC jmptarget:$dest, 0)>;
-def J_s: Pseudo<(outs), (ins stackin:$dest), []>;
```
vladimirradosavljevic: In some optimizations, we utilize the implicit Flags register to identify the defs and uses of it. With this change, we would always assume that `JCs` instruction is using flags, which could be incorrect (as it is shown in the test). Even thought I don't think this change currently creates an issue, my suggestion is to keep this instruction, and think about how to remove this overhead later on. Wdyt?

llvm/lib/Target/EraVM/EraVMOpcodes.td
atrosinenko: Fixed

llvm/lib/Target/EraVM/EraVMExpandPseudoInsts.cpp
atrosinenko: Grepped for `addImm(0)`, replaced with `addImm(EraVMCC::COND_NONE)` where appropriate.

llvm/test/MC/EraVM/asm-parser/arith-instructions-errors.s
atrosinenko: Fixed

llvm/test/MC/EraVM/asm-parser/arith-operands-errors.s
atrosinenko: Added two test cases:
```
  add  stack+[@global + 1 + r1], r2, r3
  add  r1, r2, stack+[@global + 1 + r3]
; CHECK:       <stdin>:{{[0-9]+}}:14: error: '=' expected
; CHECK-NEXT:    add  stack+[@global + 1 + r1], r2, r3
; CHECK-NEXT:               ^
; CHECK:       <stdin>:{{[0-9]+}}:22: error: '=' expected
; CHECK-NEXT:    add  r1, r2, stack+[@global + 1 + r3]
; CHECK-NEXT:                       ^
```

llvm/test/MC/EraVM/asm-parser/events.s
atrosinenko: Removed `FIXME`

llvm/test/MC/EraVM/asm-parser/sp-changes.s
atrosinenko: Added `NOPrrr` instruction and `nop r0, r0, r0 -> nop` alias + test cases.

llvm/lib/Target/EraVM/EraVMInstrFormats.td
atrosinenko: Fixed

llvm/lib/Target/EraVM/EraVMInstrInfo.td
atrosinenko: Fixed: `s/to make disassembler parse/to make asm parser handle/` (3 instances).

llvm/lib/Target/EraVM/EraVMInstrInfo.td
atrosinenko: Dropped the commit that removes `J_s` for now.

llvm/lib/Target/EraVM/EraVMInstrFormats.td
asl: I've created a separate issue for this. Let's do afterwards.

asl:
@hedgar2017 @akiramenai @antonbaliasnikov regression failures look like CI issue. Will you please check what is going on? Thanks!

asl:
@antonbaliasnikov I think https://github.com/matter-labs/era-compiler-llvm/commit/6e708f8c2bde3b20eb2186e9ec00ce9acf0bad33#diff-f0de00012d74c09571bcabcb38ab9ba495177a8e76c00e5ab27cc887118e9b79R62 does not look properly with auto-generated files (during the build process) and LLVM has plenty of them.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
atrosinenko: Finally had to introduce a separate commit https://github.com/matter-labs/era-compiler-llvm/pull/448/commits/fbdba35e29de815a70f550c76e85aded75635b2b and fold the expansion into https://github.com/matter-labs/era-compiler-llvm/pull/448/commits/3090f94936e71b1ea18274d1fa73419ce6e9f674, so J_s is properly expanded at any point in history (though, it seems integration tests are only executed for the HEAD commit of the branch).

hedgar2017:
@asl this can happen if the branch is not rebased. Benchmarks and run against a newer `main` which is more efficient.
Rebasing now.

asl:
> Benchmarks and run against a newer `main` which is more efficient.
> Rebasing now.

I do not think this is the issue. As these are per-commit regression tests failing due to being unable to build LLVM. We checked locally these revisions and they are fine. So, definitely a CI issue.

hedgar2017:
@asl understood, I thought the slight changes in benchmark were the problem, but I'm OK with ignoring them.
@antonbaliasnikov could you check why `benchmark` succeeds and `default` and fails? Please fix in `main` and we'll rebase.

sayon:
I can only build the latest commits in debug mode, otherwise I am getting errors like that one:

```
Undefined symbols for architecture arm64:
  "llvm::MachineInstr::dump() const", referenced from:
      (anonymous namespace)::EraVMAsmPrinter::emitInstruction(llvm::MachineInstr const*) in libLLVMEraVMCodeGen.a[5](EraVMAsmPrinter.cpp.o)
ld: symbol(s) not found for architecture arm64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
[199/296] Linking CXX executable bin/bugpoint
```

(thanks to @vladimirradosavljevic)

asl:
@sayon Are you running a build w/o asserts? Which revision fails in such way?

sayon:
@asl yes without asserts (for building multiple artifacts, the debug build is sometimes painfully slow).

I used commit 3ecd9c4b40e585bd6c7d79d14352d1d13ac5f2d8.

asl:
> @asl yes without asserts (for building multiple artifacts, the debug build is sometimes painfully slow).
> 
> I used commit [3ecd9c4](https://github.com/matter-labs/era-compiler-llvm/commit/3ecd9c4b40e585bd6c7d79d14352d1d13ac5f2d8).

Thanks. This might be simple to fix.

asl:
> @asl yes without asserts (for building multiple artifacts, the debug build is sometimes painfully slow).

Ok, this was fixed.

llvm/test/MC/EraVM/asm-parser/sp-changes.s
sayon: I have double checked with the existing assembler and the syntax `nop stack+=[reg+imm]` as well as `nop stack -= [reg + imm]` are accepted; I have updated the spec -- please add the test for these cases too.

sayon (changes requested):

atrosinenko:
I pushed an initial version of the branch rebased to LLVM 17 to run CI as early as posible. Though, more changes have to be done, at least the very first commit has to be folded somewhere.

atrosinenko:
* dropped a few changes - [reverted](https://github.com/matter-labs/era-compiler-llvm/compare/ac0e8c554a032602ffe6826dfc21b0ae0bc22a78..e0a87253a6bf59fa2a994a419d16c7121e18544a) back to overloaded constructors in `EraVMOperand` as there are only four now (exactly one per union member), so passing lambdas to static function is too complicated
* dropped a temporary commit as #486 is merged now

sayon (approved):

akiramenai (approved):
LGTM, thanks.
I'm going to wait until the integration tests pass and then merge this PR.

antonbaliasnikov:
> LGTM, thanks. I'm going to wait until the integration tests pass and then merge this PR.

@akiramenai , there are some problems with our runners who cannot pick up jobs. I canceled them for now, but I'll restart them as soon as the problem is fixed and let you know.

antonbaliasnikov:
@akiramenai , please, rebase this branch from the latest `main` to take CI fixes. CI should work now.

asl:
@akiramenai Looks like there is unresolved conversation that was attached to a commit that does not exist anymore. As a result, the PR cannot be merged :) For the record, the comment was resolved.

PR #449, opened 2024-03-29
[EraVM] Don't traverse over subloop BBs in EraVMIndexedMemOpsPrepare

vladimirradosavljevic:


llvm/lib/Target/EraVM/EraVMIndexedMemOpsPrepare.cpp
```diff
@@ -224,7 +224,15 @@ bool EraVMIndexedMemOpsPrepare::isValidGEPAndIncByOneCell(
...
   const unsigned LastIndexOperandId = BasePtr->getNumOperands() - 1;
   Value *LastIndexOperand = BasePtr->getOperand(LastIndexOperandId);
-  return isa<PHINode>(LastIndexOperand);
+  const PHINode *PHI = dyn_cast<PHINode>(LastIndexOperand);
+  if (!PHI)
+    return false;
+
+  // In rewriteToFavorIndexedMemOps, new PHI is created with incoming values
+  // from preheader and latch. Check that this PHI has the same incoming BBs.
+  return PHI->getNumIncomingValues() == 2 &&
```
akiramenai, line 233: The commit message says it's a paranoia check, so could it be an `assert`?

llvm/test/CodeGen/EraVM/indexed-memops-skip-subloop-bbs.ll
```diff
@@ -0,0 +1,33 @@
...
+  br i1 %cond, label %bb6, label %bb5
+
+bb5:                                              ; preds = %bb4
+  unreachable
+
+bb6:                                              ; preds = %bb4
+  %add = add i256 %phi, 32
+  br i1 false, label %bb4, label %bb7
+
+bb7:                                             ; preds = %bb6, %bb3
```
akiramenai: nit: formatting

akiramenai (approved):
Nit: Could we fix grammar in the commit message? it misses definite and indefinite articles.

llvm/test/CodeGen/EraVM/indexed-memops-skip-subloop-bbs.ll
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMIndexedMemOpsPrepare.cpp
vladimirradosavljevic: I prefer adding asserts to code that is responsible for functionality, but since this is an optimization I would skip adding assert and leave this paranoia check.
The reason why I left this check is this was my first idea how to fix this bug, but at the end I found the better solution with skipping subloop BBs. If you think this check is not needed, I can remove it.

vladimirradosavljevic:
> Nit: Could we fix grammar in the commit message? it misses definite and indefinite articles.

Fixed grammar. Thanks for the review!

llvm/lib/Target/EraVM/EraVMIndexedMemOpsPrepare.cpp
akiramenai: @vladimirradosavljevic removing is definitely not the best idea.
The minimum we should do is to document that it should never happen like in assert. Otherwise it looks like a mere expected case we bail out.

But I'm still not convinced in the policy. Perhaps it's indeed better not to shock a user with an ICE in an optimization which wouldn't happen otherwise, but we broke one of our fundamental assumptions. Thus something went quite different from what we expected. @FlashSheridan, @antonbaliasnikov, @PavelKopyl may I ask your opinion on the matter? [Please read the discussion above]

vladimirradosavljevic: AFAIK we disabled assert build for releases, thus users can hit segfault or get wrongly optimized code if this check is not true and we do rewrite regardless.

FlashSheridan: Here’s some background:

* 2024-03-01  DB [agrees](https://matter-labs-workspace.slack.com/archives/D05UK4GGG10/p1709312784239619) with enabling assertions for MUSL (the only place at which they were then disabled.)
* 2024-03-06  [FS](https://matter-labs-workspace.slack.com/archives/C05A2100E11/p1709756137926469?thread_ts=1709747625.205549&cid=C05A2100E11):  “Some of the LLVM code relies on assertions, e.g. in the following Klocwork defect, so this may be unsafe:… All our non-MUSL builds enable assertions; Alive2 also recommends Release with assertions, though they don’t give a motive.  Apple’s Clang doesn’t enable them, but their QA is not a good example, and a miscompilation on a phone app is less worrying than in a financial smart contract.…
“There’s even clearer, but less relevant, reliance at llvm/lib/Demangle/MicrosoftDemangle.cpp:2005: 
``` 
TagTypeNode *Demangler::demangleClassType(StringView &MangledName) {
  TagTypeNode *TT = nullptr;

  switch (MangledName.popFront()) {
  case 'T':
…
    break;
  default:
    assert(false);
  }
  ```
* 2024-03-06 AZ [disables](https://matter-labs-workspace.slack.com/archives/C05A2100E11/p1709760610812729?thread_ts=1709747625.205549&cid=C05A2100E11) them for release.
* 2024-03-14 “enable assertions for integration tests and benchmarks” [#434](https://github.com/matter-labs/compiler-llvm/pull/434).

FlashSheridan: My personal preference, given the uncertainty, is both to build Release with Assertions, and to have code which will crash rather than misoptimize.  Quoting myself:

> Rather unusually for quality assurance, crashes for a compiler are relatively benign, at least compared with execution-correctness bugs. A compiler crash or assertion failure affects only the programmer, and will be immediately obvious to him or her. The programmer may be able to work around the problem and, if the problem is reproducible, it will probably be straightforward to fix the compiler. In our experience, such crashes usually occurred, at least after the initial testing, when exercising relatively obscure compiler functionality, and hence were unlikely to affect a large number of programmers.

> A code-correctness bug, on the other hand, may not be immediately apparent to the programmer. Ideally it will be noticed by testers, but it is possible (especially if the bug occurs only at high optimization) that it will only be noticed by end-users after a product built with the compiler has shipped. This could require a recall of products built with the buggy compiler and sold to a large number of end-users.

PR #450, opened 2024-03-29
[EraVM] Tag fatptr copies in PHI node elimination pass

vladimirradosavljevic:


akiramenai:
Thanks, @vladimirradosavljevic 
I suggest starting to be more explicit on intent in commit messages. We have time for that.

The first commit could be tagged with either MIR or CodeGen tag. The rational behind it is that we introduced a new MVT, but didn't introduce printing and parsing for it. So could we add it to the message, or perhaps better is to turn it into a fixup commit. Also a test for the introduced functionality is very welcome.
The second commit shows a (probably unexposed) bug. It's not apparent from just mentioning hooks.
And the reason we introduce hooks is to be explained in the third commit. Also it worth adding a short doc in the code for the hooks themselves.

vladimirradosavljevic:
> Thanks, @vladimirradosavljevic I suggest starting to be more explicit on intent in commit messages. We have time for that.
> 
> The first commit could be tagged with either MIR or CodeGen tag. The rational behind it is that we introduced a new MVT, but didn't introduce printing and parsing for it. So could we add it to the message, or perhaps better is to turn it into a fixup commit. Also a test for the introduced functionality is very welcome. The second commit shows a (probably unexposed) bug. It's not apparent from just mentioning hooks. And the reason we introduce hooks is to be explained in the third commit. Also it worth adding a short doc in the code for the hooks themselves.

1. Converted 1st commit into a fixup and added testcase.
2. Updated commit subject for the 2nd commit and changed the name of the test.
3. Updated commit subject and message for the 3rd commit, and added comment in for the hooks.

Please note that I added this PR just for the completeness as I don't think it is needed, since for the fatptr design, it should be sufficient that we only add `IsFatPtr` flag just before virtual to physical register rewrite, like we do in `VirtRegMap.cpp` and in `RegAllocFast.cpp`. I think all local changes related to fatptr build copy and tagging copy instruction PreRA are redundant if we provided correct information for common passes like `RegisterCoalescer`.

akiramenai:
@vladimirradosavljevic 
> Please note that I added this PR just for the completeness as I don't think it is needed, since for the fatptr design, it should be sufficient that we only add IsFatPtr flag just before virtual to physical register rewrite, like we do in VirtRegMap.cpp and in RegAllocFast.cpp. I think all local changes related to fatptr build copy and tagging copy instruction PreRA are redundant if we provided correct information for common passes like RegisterCoalescer.

Could you kindly create a ticket to refactor that?

akiramenai (approved):
I'd prefer to have more explicit documentation of issues we encounter. Documenting the problem in the commit message could help with maintenance in future. We have a test, but navigating to the specific pattern and to what is happening will help. But I don't insisnt.

vladimirradosavljevic:
> @vladimirradosavljevic
> 
> > Please note that I added this PR just for the completeness as I don't think it is needed, since for the fatptr design, it should be sufficient that we only add IsFatPtr flag just before virtual to physical register rewrite, like we do in VirtRegMap.cpp and in RegAllocFast.cpp. I think all local changes related to fatptr build copy and tagging copy instruction PreRA are redundant if we provided correct information for common passes like RegisterCoalescer.
> 
> Could you kindly create a ticket to refactor that?

Done, CPR-1639.

PR #462, opened 2024-04-02
[EraVM] Move ExpandPseudo before IfConverter

vladimirradosavljevic:


akiramenai (approved):
If you have time, please describe in the first commit message what is happening and what is supposed to happen instead. (By the time of `ifcvt` we deal with `loadconst` pseudo that can't be predicated).

PR #466, opened 2024-04-03
[EraVM] Allow rematerialization of the move 16-bit immediates

vladimirradosavljevic:


vladimirradosavljevic:
This PR depends on #462.

PR #474, opened 2024-04-08
[EraVM] Rematerialize and combine 16-bit immediates

vladimirradosavljevic:


llvm/test/CodeGen/EraVM/fold-move-imm.ll
```diff
@@ -0,0 +1,193 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
+; RUN: llc -O3 < %s | FileCheck %s
+
+target datalayout = "E-p:256:256-i256:256:256-S32-a:256:256"
+target triple = "eravm"
+
+define i256 @test_pos(i256 %arg) {
```
akiramenai, line 7: Can we minimize the test cases?

llvm/lib/Target/EraVM/EraVMCombineAddressingMode.cpp
```diff
@@ -553,30 +566,45 @@ bool EraVMCombineAddressingMode::combineConstantUse(MachineFunction &MF) {
   }
 
   // 2. Combine.
-  for (auto [LoadConst, Uses] : Deleted) {
+  for (auto [MoveImm, Uses] : Deleted) {
+    SmallVector<MachineOperand, 2> In0Ops;
+    // TODO: CPR-1669 If all users are COPY, don't create a constant pool entry.
```
akiramenai: We run the pass in post-ra.

akiramenai (changes requested):

llvm/lib/Target/EraVM/EraVMCombineAddressingMode.cpp
vladimirradosavljevic: Updated comment.

llvm/test/CodeGen/EraVM/fold-move-imm.ll
vladimirradosavljevic: I saw this pattern in one of our benchmarks, and I minimize it as much as possible. I know that for this purpose is more complex than it should be, so if you want I can create smaller MIR test.

akiramenai (approved):

PR #479, opened 2024-04-12
[EraVM] Update expansion of memcpy with unknown size

vladimirradosavljevic:


akiramenai (approved):
LGTM, but I would explain why simplified form benefits us. It's because of Terry's pass.

Basically if I understood it right your goal was to use incremented ld/st in the loop body and because the loop wasn't in the simplified form, Terry's pass failed to transform it.
It's inferable from the code but it requires time even for a person who knows our backend well. So could we be more explicit in purposes and effects of a patch. Let's keep inference distance for a reader minimally possible. It's ok to assume that a reader knows LLVM and compiler theory terms (or can google them), but I think the rest is to be explained. What is the effect? Why it's profitable (and when it isn't)? What was changed to get the effect and when it's not obvious why it works?

vladimirradosavljevic:
> LGTM, but I would explain why simplified form benefits us. It's because of Terry's pass.
> 
> Basically if I understood it right your goal was to use incremented ld/st in the loop body and because the loop wasn't in the simplified form, Terry's pass failed to transform it. It's inferable from the code but it requires time even for a person who knows our backend well. So could we be more explicit in purposes and effects of a patch. Let's keep inference distance for a reader minimally possible. It's ok to assume that a reader knows LLVM and compiler theory terms (or can google them), but I think the rest is to be explained. What is the effect? Why it's profitable (and when it isn't)? What was changed to get the effect and when it's not obvious why it works?

Updated commit message, and added comments in the code why we are creating empty preheader and exit BBs.

PR #488, opened 2024-04-19
[EraVM] Fix branch probability calculation in probabilityIsProfitable

vladimirradosavljevic:
In some cases, IfConverter can pass max 32-bit unsigned value for TrueCycles and 0 for FalseCycles. Problem in this case is that Denominator will be 0, and BranchProbability will assert. Add check to avoid assertion in BranchProbability constructor.


akiramenai:
The title is too long. Maybe `Fix branch probability calculation [in ...]`?

vladimirradosavljevic:
> The title is too long. Maybe `Fix branch probability calculation [in ...]`?

Done.

PR #493, opened 2024-04-23
[EraVM] Optimize signed compares with immediate

vladimirradosavljevic:


llvm/lib/Target/EraVM/EraVMTargetMachine.cpp
```diff
@@ -262,6 +264,12 @@ void EraVMPassConfig::addIRPasses() {
   TargetPassConfig::addIRPasses();
 }
 
+void EraVMPassConfig::addCodeGenPrepare() {
+  TargetPassConfig::addCodeGenPrepare();
+  if (getOptLevel() != CodeGenOpt::None)
+    addPass(createEraVMPostCodegenPreparePass());
```
akiramenai, line 270: What was the reason to extract it to a separate pass?
Codegen prepare technically is also about optimization, not correctness.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -0,0 +1,128 @@
+//===-- EraVMPostCodegenPrepare.cpp - EraVM Post CodeGenPrepare -*- C++ -*-===//
+//
+// Part of the LLVM Project, under the Apache License v2.0 with LLVM Exceptions.
+// See https://llvm.org/LICENSE.txt for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//===----------------------------------------------------------------------===//
+//
+// This pass performs optimizations after CodeGenPrepare pass. Currently, it is
```
akiramenai: Does "currently" imply that we plan to perform all signed to unsigned comparisons in this pass one day?
Generally, if we plan to do something aside from what is implemented currently, it's better to be explicit on that: create an issue and link it in a TODO.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -0,0 +1,128 @@
...
+// only performing transformation from signed to unsigned compares where it is
+// feasible.
```
akiramenai: Could we refine this statement? We only do the transformation for `s<`, `s>` and a constant in RHS. We probably don't care about `s>=`, `s<=` as we expect LLVM to convert them to strict comparisons. So ultimately feasibility == const in RHS.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -0,0 +1,128 @@
...
+           "Expected unsigned predicate in Pred2");
+    auto *Cmp0 = Builder.CreateICmp(
+        Pred2, Op256, Builder.getInt(CmpVal->getValue().zext(256)));
+    NewCmp = Builder.CreateBinOp(Opc, NewCmp, Cmp0);
+  }
+  return NewCmp;
+}
+
+// Since we don't have signed compare instructions, try to transform them into
+// unsigned compares where possible. Currently, it is trying to do following
```
akiramenai: It's always possible. But we limit this pass with constants in RHS and handle the case of variables on DAG. (TODO: move it here - if you think it will improve maintainability).

akiramenai (changes requested):
Documentation is to be improved.

llvm/lib/Target/EraVM/EraVMTargetMachine.cpp
vladimirradosavljevic: Reason for moving this optimization after CodeGen prepare is not for the correctness, but it is for optimizations. In this case, we can leverage compare optimizations (namely sinkCmpExpression).

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Currently in this context means we want to use it for other optimizations, and not only for this one. Here is my suggestion what we can do:
1. Move handling of call instruction from `EraVMCodegenPrepare`.
2. Rename `EraVMCodegenPrepare` to something like `EraVMConvertPointerArithmetics`.
3. Rename `EraVMPostCodegenPrepare` to `EraVMCodegenPrepare`.
4. Implement `rearrangeOverflowHandlingBranches` from [here](https://github.com/matter-labs/era-compiler-llvm/commit/f31aa85caee7ccb207d66807cb6792cb7b8fa60b#diff-9b0cab2bf4202033e73cf5e22c78e0d1d72186d157542fca76d292a04985d637) to this pass, since CodeGenPrepare can also generate overflow intrinsics.

If you are OK with this, I will add TODOs and create tickets for these items.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Will update comment. We don't need to worry about ule and uge, since these will be canonicalized to ult and ugt as it is shown [here](https://godbolt.org/z/v1z6cxzf4).

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: I don't think we want that, since ISEL can also generate signed compares. Yes, we can move lowering here and disable all these ISEL optimizations, but it looks like an overkill to me since we would introduce a lot of local changes + we would need to worry about this every time we do the rebase to a newer version of the LLVM.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
akiramenai: Yes, please

akiramenai (approved):
LGTM, thanks

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Done. Created ticket for renaming, and added ticket numbers in the code.

PR #494, opened 2024-04-23
[EraVM] Add a pass to rewrite dead defs to R0

vladimirradosavljevic:


akiramenai:
I see that we noticeably reduced size of the precompiles in M3B3, but aside from that the effect is negligible. Could we add some information on the pass rationale better than it could improve regalloc? When we expect it to be profitable and what causes regressions? Do other targets use the similar approach?

vladimirradosavljevic:
> I see that we noticeably reduced size of the precompiles in M3B3, but aside from that the effect is negligible. Could we add some information on the pass rationale better than it could improve regalloc? When we expect it to be profitable and what causes regressions? Do other targets use the similar approach?

Will add info where I saw improvements with this patch in the commit message. Even though the overall improvements seem negligible, number of tests where cycles are improved (48075) and `SHA3` opcode improvements of `0.714` seem decent.
I observed the following pattern that is generated after this patch that is causing degradations:
```
	add	r1, r0, stack-[3]               ; 32-byte Folded Spill
	add	stack-[3], r0, r1               ; 32-byte Folded Reload
```
Reload here to `r1` is not needed since we spilled that register in the instruction above. There is an optimization to remove unneeded spills if the previous instruction is reload from the same stack slot (`StackSlotColoring::RemoveDeadStores`), so I assume in order to fix this degradation, we would need to create a similar optimization to remove unneeded reloads.
Yes, both AArch64 and RISCV (in the trunk) have the same optimization for some of their instructions.

akiramenai:
Now it's better and we identified a way to improve the results further. Thank you!
```
Also, in some cases it can help regalloc to
do a better job so lowering of select instruction can
be done in one instruction while reversing condition.
```

It's vague. Could it be clarified?
Also I think the pass sources is a better place to explain why do we need it and how it interacts with the rest of optimization pipeline.

vladimirradosavljevic:
> Now it's better and we identified a way to improve the results further. Thank you!
> 
> ```
> Also, in some cases it can help regalloc to
> do a better job so lowering of select instruction can
> be done in one instruction while reversing condition.
> ```
> 
> It's vague. Could it be clarified? Also I think the pass sources is a better place to explain why do we need it and how it interacts with the rest of optimization pipeline.

Removed explanation from the commit message, and move it to the pass preheader + added an example how this can reduce lowering of select instruction.

akiramenai (approved):

PR #497, opened 2024-04-25
[EraVM][MC] Implement basic support for code emission

atrosinenko:
Implement support for encoding simple instructions - those only using register and immediate operands.

In TableGen descriptions, connect Inst{...} fields and operands of various instructions, split instruction classes that have different sets of operands.

In C++ code, make use of tablegen-erated instruction encoders, add support for encoding register and immediate operands, as well as condition codes.

Support for the instructions with in_any and out_any operands requires custom handling of stack operands and will be added via subsequent commits.


llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -642,56 +712,102 @@ class IUMAr_r<EraVMOpcode opcode,
...
                string asmstr,
                list<dag> pattern>
- : IUMA<opcode, SrcImm, 1, outs, ins, asmstr, pattern> {
-}
+ : IUMA<opcode, SrcSpecialImm, 1, outs, ins, asmstr, pattern> {
+  bits<16> addr;
+  bits<4> rs1;
+  bits<4> rd0;
 
-// TODO: There should be IUMAs / IUMAc
```
atrosinenko, line 714: If I understand correctly, IUMA class is only used for heap access instructions (i.e. pointer *dereference*, but not pointer *arithmetics*) that only accept `in_reg` / `in_regimm` and `out_reg` operands.

akiramenai: Yes, it's correct.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
```diff
@@ -91,7 +99,26 @@ uint64_t EraVMMCCodeEmitter::getMemOpValue(const MCInst &MI, unsigned Idx,
 uint64_t EraVMMCCodeEmitter::getCCOpValue(const MCInst &MI, unsigned Idx,
                                           SmallVectorImpl<MCFixup> &Fixups,
                                           const MCSubtargetInfo &STI) const {
-  llvm_unreachable("Unable to encode predicate operand!");
+  switch (MI.getOperand(Idx).getImm()) {
+  default:
+    llvm_unreachable("Invalid condition code");
+  case EraVMCC::COND_INVALID:
+    llvm_unreachable("Cannot encode COND_INVALID");
+  case EraVMCC::COND_NONE:
```
akiramenai: For non-invalid, not none members of the enum, can't we just static cast to `uint64_t` and if necessary reenumerate the enum?

akiramenai:
Add @sayon as a reviewer as I think he is the best person to check the encoding tests against the spec.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
atrosinenko: Turned out, EraVMInstrInfo.td has patterns with hardcoded `2`, `4` and `8` constants (though, with corresponding `COND_*` mentioned in comments). I will upload a separate PR soon.

atrosinenko:
Pushed a script to simplify comparison with the standalone assembler.

atrosinenko:
Made this PR stacked as further PRs will depend on both this and `R_ERAVM_*` definitions.

llvm/test/MC/EraVM/compare-with-standalone-assembler.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
```
akiramenai, line 1: Could tell me in which scenarios do you use this script and if we should add some CI workflows to start using it automatically?

akiramenai (commented):
@atrosinenko could you clarify why don't we test 2 output instructions? Also we don't plan to test all opcodes, do we?

akiramenai (approved):
I found more tests in #507, so it looks good to me.

atrosinenko:
Thank you, I would like to update `getCCOpValue` in this PR soon, so that it uses #502.

akiramenai:
Please update, so we can merge.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
atrosinenko: Addressed via 388ead4c8c9cc83158551e3422570f8f382da955

atrosinenko:
Updated. Now I will squash the three commits and update downstream #500 and upstream #503 PRs so that this PR has `main` as its destination branch, if you don't mind.

atrosinenko:
Updated #500 and this PR, but will not touch #503 until actual changes to account for `COND_*` renumbering - its CI jobs were not automatically triggered, so let's not trigger them manually without a reason. :)

llvm/test/MC/EraVM/compare-with-standalone-assembler.sh
atrosinenko: I consider this script useful for manual testing, but I doubt it should be executed by CI as it may require manual analysis of the results for less trivial cases. It can be used like this:

```
$ cd /path/to/era-compiler-llvm/build
$ ../llvm/test/MC/EraVM/compare-with-standalone-assembler.sh ../llvm/test/MC/EraVM/encoding/*.s
Paths to be used:
llvm-mc: /path/to/era-compiler-llvm/build/bin/llvm-mc
reader:  /path/to/era-zkEVM-assembly/target/debug/reader
Test source: ../llvm/test/MC/EraVM/encoding/events.s
Filtered lines: 17 of 17
LLVM MC produced 4 lines

Reference assembler produced 12 lines, cropping to 4 lines

== DIFF BEGIN ==
== DIFF END ==

Test source: ../llvm/test/MC/EraVM/encoding/inst-fields.s
Filtered lines: 65 of 65
LLVM MC produced 24 lines

Reference assembler produced 36 lines, cropping to 24 lines

== DIFF BEGIN ==
== DIFF END ==

Test source: ../llvm/test/MC/EraVM/encoding/misc.s
Filtered lines: 31 of 31
LLVM MC produced 11 lines

Reference assembler produced 20 lines, cropping to 11 lines

== DIFF BEGIN ==
== DIFF END ==

Test source: ../llvm/test/MC/EraVM/encoding/returns.s
Filtered lines: 25 of 25
LLVM MC produced 6 lines

Reference assembler produced 20 lines, cropping to 6 lines

== DIFF BEGIN ==
== DIFF END ==

Test source: ../llvm/test/MC/EraVM/encoding/storage.s
Filtered lines: 13 of 13
LLVM MC produced 2 lines

Reference assembler produced 12 lines, cropping to 2 lines

== DIFF BEGIN ==
== DIFF END ==

Test source: ../llvm/test/MC/EraVM/encoding/uma.s
Filtered lines: 53 of 53
LLVM MC produced 18 lines

Reference assembler produced 28 lines, cropping to 18 lines

== DIFF BEGIN ==
== DIFF END ==
```

This is a best-effort tool that assumes the first N 8-byte words of the standalone assembler output are the instructions that we have to generate. Should I move it out of `llvm/test/MC/EraVM`?

PR #500, opened 2024-04-26
[EraVM] Make ELF support code know about EraVM target

atrosinenko:
Add EraVM definitions to be used by ELF-related code here and there.

### DIFF
--- a/llvm/test/tools/llvm-readobj/ELF/reloc-types-eravm.test
+++ b/llvm/test/tools/llvm-readobj/ELF/reloc-types-eravm.test
@@ -1,0 +1,23 @@
+## Test that llvm-readobj shows proper relocation type
+## names and values for the EraVM target.
+
+# RUN: yaml2obj %s -o %t-eravm.o
+# RUN: llvm-readobj -r --expand-relocs %t-eravm.o | FileCheck %s
+
+# CHECK: Type: R_ERAVM_NONE (0)
+# CHECK: Type: R_ERAVM_16_SCALE_32 (1)
+# CHECK: Type: R_ERAVM_16_SCALE_8 (2)
+
+--- !ELF
+FileHeader:
+  Class:   ELFCLASS32
+  Data:    ELFDATA2MSB
+  Type:    ET_REL
+  Machine: EM_ERAVM
+Sections:
+  - Name:         .rel.text
+    Type:         SHT_REL
+    Relocations:
+       - Type:   R_ERAVM_NONE
+       - Type:   R_ERAVM_16_SCALE_32
+       - Type:   R_ERAVM_16_SCALE_8
--- a/llvm/unittests/Object/ELFObjectFileTest.cpp
+++ b/llvm/unittests/Object/ELFObjectFileTest.cpp
@@ -286,6 +286,13 @@ TEST(ELFObjectFileTest, MachineTestForXtensa) {
     checkFormatAndArch(Data, Formats[Idx], Triple::xtensa);
 }
 
+TEST(ELFObjectFileTest, MachineTestForEraVM) {
+  std::array<StringRef, 4> Formats = {"elf32-eravm", "elf32-eravm",
+                                      "elf64-unknown", "elf64-unknown"};
+  for (auto [Idx, Data] : enumerate(generateData(ELF::EM_ERAVM)))
+    checkFormatAndArch(Data, Formats[Idx], Triple::eravm);
+}
+
 // ELF relative relocation type test.
 TEST(ELFObjectFileTest, RelativeRelocationTypeTest) {
   EXPECT_EQ(ELF::R_CKCORE_RELATIVE, getELFRelativeRelocationType(ELF::EM_CSKY));
--- a/llvm/include/llvm/BinaryFormat/ELF.h
+++ b/llvm/include/llvm/BinaryFormat/ELF.h
@@ -320,6 +320,9 @@ enum {
   EM_VE = 251,            // NEC SX-Aurora VE
   EM_CSKY = 252,          // C-SKY 32-bit processor
   EM_LOONGARCH = 258,     // LoongArch
+  // EraVM local begin
+  EM_ERAVM = 260,         // EraVM
+  // EraVM local end
 };
 
 // Object file classes.
@@ -948,6 +951,11 @@ enum : unsigned {
 // ELF Relocation types for Xtensa
 enum {
 #include "ELFRelocs/Xtensa.def"
+};
+
+// ELF Relocation types for EraVM
+enum {
+#include "ELFRelocs/EraVM.def"
 };
 
 #undef ELF_RELOC
--- a/llvm/include/llvm/BinaryFormat/ELFRelocs/EraVM.def
+++ b/llvm/include/llvm/BinaryFormat/ELFRelocs/EraVM.def
@@ -1,0 +1,8 @@
+
+#ifndef ELF_RELOC
+#error "ELF_RELOC must be defined"
+#endif
+
+ELF_RELOC(R_ERAVM_NONE,               0)
+ELF_RELOC(R_ERAVM_16_SCALE_32,        1)
+ELF_RELOC(R_ERAVM_16_SCALE_8,         2)
--- a/llvm/include/llvm/Object/ELFObjectFile.h
+++ b/llvm/include/llvm/Object/ELFObjectFile.h
@@ -1225,6 +1225,10 @@ StringRef ELFObjectFile<ELFT>::getFileFormatName() const {
       return "elf32-loongarch";
     case ELF::EM_XTENSA:
       return "elf32-xtensa";
+    // EraVM local begin
+    case ELF::EM_ERAVM:
+      return "elf32-eravm";
+    // EraVM local end
     default:
       return "elf32-unknown";
     }
@@ -1351,6 +1355,11 @@ template <class ELFT> Triple::ArchType ELFObjectFile<ELFT>::getArch() const {
 
   case ELF::EM_XTENSA:
     return Triple::xtensa;
+
+  // EraVM local begin
+  case ELF::EM_ERAVM:
+    return Triple::eravm;
+  // EraVM local end
 
   default:
     return Triple::UnknownArch;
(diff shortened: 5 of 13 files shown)


llvm/unittests/Object/ELFObjectFileTest.cpp
```diff
@@ -286,6 +286,14 @@ TEST(ELFObjectFileTest, MachineTestForXtensa) {
     checkFormatAndArch(Data, Formats[Idx], Triple::xtensa);
 }
 
+TEST(ELFObjectFileTest, MachineTestForEraVM) {
+  // TODO Should we support both LSB and MSB?
```
asl: No. EraVM is BE

asl:
@akiramenai @antonbaliasnikov How can we disable format checks for unrelated LLVM files?

antonbaliasnikov:
> @akiramenai @antonbaliasnikov How can we disable format checks for unrelated LLVM files?

Hi @asl!

Could you, please, clarify for me which files you categorize as "unrelated"? There is already filtering of the files, and the CI is working exactly how it is working in the LLVM upstream.

Please, let me know which enhancements you would like to see, I will take a look.

asl:
@antonbaliasnikov See https://github.com/matter-labs/era-compiler-llvm/actions/runs/8852451395/job/24311225008?pr=500

antonbaliasnikov:
> @antonbaliasnikov See https://github.com/matter-labs/era-compiler-llvm/actions/runs/8852451395/job/24311225008?pr=500

Got you, thanks for reporting. There is definitely some bug here with identifying changes for formatting, I'll take a look tomorrow.

antonbaliasnikov:
> > @antonbaliasnikov See https://github.com/matter-labs/era-compiler-llvm/actions/runs/8852451395/job/24311225008?pr=500
> 
> Got you, thanks for reporting. There is definitely some bug here with identifying changes for formatting, I'll take a look tomorrow.

@asl , I've investigated the root cause of the issues and why we see some unrelated code like that in the formatting:
```
-  ENUM_ENT(EM_NONE,          "None"),
-  ENUM_ENT(EM_M32,           "WE32100"),
-  ENUM_ENT(EM_SPARC,         "Sparc"),
-  ENUM_ENT(EM_386,           "Intel 80386"),
```

CI is correctly identifying the changes and commits, no issues here. The thing is that we're adding an enum entry [here](https://github.com/matter-labs/era-compiler-llvm/pull/500/files#diff-d7e0bfa4d6ead9eab1e71e8b50eabddcbf12d99b53a576bce053252d91be183eR1261-R1263) that results in `clang-format` to check the whole `const EnumEntry<unsigned> ElfMachineType[]` enum and by some reason it adds additional indentation to it. We can reproduce it locally by running `clang-format` for the `llvm/tools/llvm-readobj/ELFDumper.cpp` file.

I'm not sure how we should approach this particular issue, because the formatting of our changes in this PR looks absolutely OK to me, but `clang-format` still raises an issue because this file is probably not formatted accordingly to the latest LLVM standards that are in use by default.

CC: @FlashSheridan , @hedgar2017

atrosinenko:
Rebased to `main`.

atrosinenko:
... and defined two actual `R_ERAVM_*` relocation types. These are what is to be supported by subsequent stacked PRs, so dropping TODOs from code and an awkward note from the commit message.

FlashSheridan:
> I'm not sure how we should approach this particular issue, because the formatting of our changes in this PR looks absolutely OK to me, but `clang-format` still raises an issue because this file is probably not formatted accordingly to the latest LLVM standards that are in use by default.
> CC: @FlashSheridan , …

Yes, we made a [policy decision](https://linear.app/matterlabs/issue/CPR-1233/clang-tidy-simple-git-hook-checks#comment-39cf4d62) for clang-tidy that when necessary we would bypass “checks that the rest of LLVM does not actually enforce”.

antonbaliasnikov:
> > I'm not sure how we should approach this particular issue, because the formatting of our changes in this PR looks absolutely OK to me, but `clang-format` still raises an issue because this file is probably not formatted accordingly to the latest LLVM standards that are in use by default.
> > CC: @FlashSheridan , …
> 
> Yes, we made a [policy decision](https://linear.app/matterlabs/issue/CPR-1233/clang-tidy-simple-git-hook-checks#comment-39cf4d62) for clang-tidy that when necessary we would bypass “checks that the rest of LLVM does not actually enforce”.

Then, I believe, we need to split formatting and regression tests jobs, because if formatting fails, we still can proceed with the merge.

antonbaliasnikov:
Good, the fix is in https://github.com/matter-labs/era-compiler-llvm/pull/511. As soon as we merge this (need review), we can rebase this PR, and only formatting that we should accept will fail.

antonbaliasnikov:
> Good, the fix is in #511. As soon as we merge this (need review), we can rebase this PR, and only formatting that we should accept will fail.

Done. @atrosinenko , please, rebase and if only formatting fails, feel free to merge when all other criteria are ok.

akiramenai (commented):
LGTM, but I prefer @PavelKopyl to have a look.

antonbaliasnikov:
There is also one `clang-tidy` error:
```
llvm/llvm/unittests/Object/ELFObjectFileTest.cpp:289:1: 
error: class 'ELFObjectFileTest_MachineTestForEraVM_Test' 
defines a copy constructor and a copy assignment operator 
but does not define a destructor, a move constructor or 
a move assignment operator [cppcoreguidelines-special-member-functions,-warnings-as-errors]
```

PavelKopyl (approved):
LGTM.

llvm/unittests/Object/ELFObjectFileTest.cpp
atrosinenko: Looks like other targets generally use endianness information to distinguish between two valid options (such as `mips` vs. `mipsel`) but not to reject invalid combinations - say, i386 usually just assumes "the only possible" little-endian variant (there are [exceptions](https://github.com/matter-labs/era-compiler-llvm/blob/95cb0e966ba7411a9b27eab241b3ff074135cc8f/llvm/include/llvm/Object/ELFObjectFile.h#L1320), though even AMDGPU target seems to not reject big-endian everywhere). Thus, simply dropped this TODO and replaced LSB with MSB in test inputs: df28628823eb21b8d84d0c9338a54547aa60df2e.

atrosinenko:
Slightly updated the tests: changed ELFDATA2LSB to ELFDATA2MSB in a couple of test inputs and added two more test cases to the test for relocation type names (forgot to update the test when added non-trivial relocs to `EraVM.def`).

@antonbaliasnikov I'm not familiar with GoogleTest - I expect my test case to be just another copy of the existing `TEST(ELFObjectFileTest, ...`, this time adjusted for `eravm`. But I vaguely remember some clang-tidy warnings reported for other GoogleTest-based unit-tests when debugging asm-parser PR locally...

atrosinenko:
@antonbaliasnikov Meanwhile, after all issues are resolved, is it possible to merge this PR as a single squashed commit via GitHub or should it be squashed manually and pushed after another CI run?

antonbaliasnikov:
> @antonbaliasnikov Meanwhile, after all issues are resolved, is it possible to merge this PR as a single squashed commit via GitHub or should it be squashed manually and pushed after another CI run?

@atrosinenko , with the current policies set, a developer should do a manual squash.

antonbaliasnikov:
> Slightly updated the tests: changed ELFDATA2LSB to ELFDATA2MSB in a couple of test inputs and added two more test cases to the test for relocation type names (forgot to update the test when added non-trivial relocs to `EraVM.def`).
> 
> @antonbaliasnikov I'm not familiar with GoogleTest - I expect my test case to be just another copy of the existing `TEST(ELFObjectFileTest, ...`, this time adjusted for `eravm`. But I vaguely remember some clang-tidy warnings reported for other GoogleTest-based unit-tests when debugging asm-parser PR locally...

Hmm, I also marked that if you're not merging to `main`, we're not running formatting, clang-tidy, check-llvm. I'm not sure how critical it is, in this case, keeping in mind these commits will be tested later as part of `eravm-mc/basic-code-emission`.

atrosinenko:
> Hmm, I also marked that if you're not merging to main, we're not running formatting, clang-tidy, check-llvm. I'm not sure how critical it is, in this case, keeping in mind these commits will be tested later as part of eravm-mc/basic-code-emission.

@antonbaliasnikov IIUC GitHub understands "stacked PRs" if they are chained, so that the source branch of the earlier PR is the destination branch of the later one, so I guess "merging to main" is not so different from merging to anything else:

```
 +- dst branch - [PR497] - src branch -+- dst branch - [PR500] - src branch -+- dst branch - [PR503] - ...
 v                                    v                                      v
main        <---         eravm-mc/basic-code-emission       <---       eravm-mc/elf-defs <- ...
```
(@asl is this correct?)

> keeping in mind these commits will be tested later as part of eravm-mc/basic-code-emission

Just in case, if I understand this assumption correctly, it was likely invalidated as today I swapped #497 and this PR, so `main..eravm-mc/basic-code-emission` does not include changes from this PR anymore (though, there are more PRs in the chain, but see above).

llvm/test/tools/llvm-readobj/ELF/reloc-types-eravm.test
```diff
@@ -0,0 +1,23 @@
+## Test that llvm-readobj shows proper relocation type
```
atrosinenko, line 1: Squash everything into a single commit. _(Just a formally unresolved comment to not accidentally push fix-up separately)_

asl:
@antonbaliasnikov Looks like `clang-tidy` is triggering on gtest boilerplate. We cannot do anything about this.

antonbaliasnikov:
> @antonbaliasnikov Looks like `clang-tidy` is triggering on gtest boilerplate. We cannot do anything about this.

Feel free to ignore then, I believe.

CC: @akiramenai

atrosinenko:
> Feel free to ignore then, I believe.

@antonbaliasnikov Looks like clang-tidy failure prevents lit tests from running.

akiramenai:
I run `check-llvm` locally. It passed, so I merge.
@antonbaliasnikov please proceed with decoupling `clang-tidy` from `check-llvm` (#513)

PR #501, opened 2024-04-29
[EraVM] Allow rematerialization of some of the context instructions

vladimirradosavljevic:


vladimirradosavljevic:
@sayon Could you please check this patch? I saw in our ISA that these instructions are cheaper than an add that we use for move, spills and reloads. Also, could you confirm that it is safe to do rematerialization of these context instructions?

sayon (approved):
LGTM. You are right that these insns are semantically like loading a constant (within the same runtime contract instance).

akiramenai (approved):
LGTM.
Thank you for improving on commit messages.

PR #502, opened 2024-04-29
[EraVM] Renumber COND_* constants and move their definitions to *.td

atrosinenko:
Renumber `EraVMCC:COND_*` constants, so that their values in the `CondCodes` enum match their binary encoding. Additionally, migrate their definition to tablegen, so they can be used both by hand-written C++ code and from patterns described in *.td files.


akiramenai (approved):
LGTM, thank you!

Issue #440, opened 2024-03-18
[MC][AsmParser] Check the list of supported condition codes


atrosinenko:
At now, a number of conditional code names are supported across the instructions. These are short forms like `.eq`, `.le`, `.ge` as well as the longer forms mentioned in the [VM Primer](https://docs.zksync.io/zk-stack/components/zkEVM/vm-specification/vm-primer.html#modifiers) (such as `.if_eq`, `.if_le`, `.if_ge`, `.if_gt_or_eq`). The mapping of strings to `EraVMCC` constants is listed in `parseExplicitCondition` function in [EraVMAsmParser.cpp](https://github.com/matter-labs/compiler-llvm/blob/eravm-mc/asm-parser/llvm/lib/Target/EraVM/AsmParser/EraVMAsmParser.cpp) file ([permalink](https://github.com/matter-labs/compiler-llvm/blob/3f7cf0cf80c9d6cf3c648fb13c7a0ed8a49a7051/llvm/lib/Target/EraVM/AsmParser/EraVMAsmParser.cpp#L446) at the time of writing).

After redesigning the assembly syntax, only the shorter form is likely to remain, so `parseExplicitCondition` function should be updated accordingly.

asl:
@sayon Will you please comment on what should be supported?

sayon:
The longer form (like .if_eq) is unnecessary, the short forms are sufficient to us.  So, 

- eq
- lt
- gt
- ne
- ge
- le

@akiramenai @hedgar2017 do you think we should support something like `of` as an alias to `lt`?

hedgar2017:
@sayon I dislike unnecessary aliases personally.

akiramenai:
The issue here is that our `lt` is not quite `lt`, it's `lt` for `sub`. But it's more ISA issue than asm, we reuse flags for different purposes in different context. Whatever naming we choose, it's misleading for some instructions. Maybe we could rename `eq` -> `z` as it works like a conventional zero flag almost everywhere. Maybe, we want to rename `lt` to some single letter same as `gt` and make `xz`, `yz` suffixes to check 2 flags combinations. An abstract 2nd flag and 3rd flag with no associations with the name might be the best solution in our case.

asl:
@akiramenai @hedgar2017 @sayon Ok, let me assign this issue to you, so you will decide on the intended functionality :)

akiramenai:
[...]

PR #503, opened 2024-04-29
[EraVM][Disassembler] Implement basic support for instruction decoding

atrosinenko:
Implement support for decoding simple instructions - those only using register and immediate operands.

Support for the instructions with in_any and out_any operands requires custom handling of stack operands and will be added via subsequent commits.


llvm/lib/Target/EraVM/Disassembler/EraVMDisassembler.cpp
```diff
@@ -39,8 +39,55 @@ class EraVMDisassembler : public MCDisassembler {
...
 };
+
+Register decodeGR256(const MCDisassembler *D, unsigned EncodedReg) {
+  const auto *MCRI = D->getContext().getRegisterInfo();
+  const auto &RC = MCRI->getRegClass(EraVM::GR256RegClassID);
+  assert(RC.getNumRegs() == 16 && "Unexpected change to register class");
+  return RC.getRegister(EncodedReg);
+}
+
+int64_t decodeCC(unsigned EncodedCC) {
```
atrosinenko, line 50: Will update this to leverage #502.

atrosinenko: Updated via 0916a7bc56339f35e37ed6a76a68d03ae0f6361c

asl:
@atrosinenko Please rebase

llvm/lib/Target/EraVM/Disassembler/EraVMDisassembler.cpp
```diff
@@ -39,8 +39,53 @@ class EraVMDisassembler : public MCDisassembler {
...
+  const unsigned EncodingForIfGTOrLT = 7;
```
akiramenai, line 51: Don't we have it defined somewhere? Also if we plan to support `GTOrLT` it makes sense to add `TODO: <issue number>`.

akiramenai (approved):
LGTM

akiramenai:
We don't use `gsl` so I'm going to disable array to pointer decay checker.

llvm/lib/Target/EraVM/Disassembler/EraVMDisassembler.cpp
atrosinenko: At now, I assume `IfGTOrLT` to be deprecated, please correct me if I'm wrong. Related: #440.

akiramenai: We probably didn't pay enough attention to this particular issue as it was discussed in context of other abbreviations.
I didn't hear about plans to remove it nor I can imagine a reason for doing so: we'd use 7 out of 8 possible predicates without it.
I would name it `gl`. :)
@sayon @hedgar2017 could you confirm?

hedgar2017: Does is stand for `greater than or less than`? Why not name it `not equal` (`ne`)? Sounds more conventional.

akiramenai: @hedgar2017 that why I proposed to rename flags to `x`, `y`, `z` to remove these associations. `gt`, `lt`, `eq` is a lie. It's true for `sub` instruction only. Instructions set some flags on overflow, on division by zero, etc. So `ne` means that `eq` flag is not set for this particular instruction; `gl` means that `gt` or `lt` is set regardless of whether `eq` is set (and if you ever see it in asm generated by our backend, gl with that)

akiramenai: Keeping this conversation unresolved until we decide on #440.

PR #506, opened 2024-04-30
[EraVM] Add patterns to select selectcc with large immediates

vladimirradosavljevic:


llvm/test/CodeGen/EraVM/ra_regress.ll
```diff
@@ -10,13 +10,14 @@ define i256 @test_length(i256 %arg1) {
...
 ; CHECK-NEXT: add     r0, r{{[0-9]+}}, r{{[[0-9]+}}
-; CHECK-NEXT: add.lt  r{{[0-9]+}}, r{{[0-9]+}}, r{{[0-9]+}}
+; CHECK-NEXT: add.lt  @CPI0_0[0], r{{[0-9]+}}, r{{[0-9]+}}
   %comparison_result24 = icmp slt i256 35, %arg1
   %comparison_result_extended25 = zext i1 %comparison_result24 to i256
 
 ; CHECK:      and     @CPI0_0[0], r{{[0-9]+}}, r{{[[0-9]+}}
 ; CHECK-NEXT: sub!    r{{[0-9]+}}, r0, r{{[[0-9]+}}
-; CHECK-NEXT: add.le  r0, r{{[0-9]+}}, r{{[[0-9]+}}
+; CHECK-NEXT: add     r0, r0, r{{[[0-9]+}}
```
vladimirradosavljevic, line 17: Even though this looks like a degradation after this patch, it's not actually.
Asm before this patch:
```
	add	r1, r0, r3
	add	@CPI0_0[0], r0, r1
	sub.s!	36, r3, r2
	add	r0, r0, r2
	add.lt	r1, r0, r2
	and	@CPI0_0[0], r3, r3
	sub!	r3, r0, r4
	add.le	r0, r0, r1
	sub.s!	@CPI0_0[0], r3, r3
	add.ne	r2, r0, r1
	sub!	r1, r0, r1
	add	0, r0, r1
	add.ne	1, r0, r1
	ret
```
Asm after this patch:
```
	sub.s!	36, r1, r2
	add	r0, r0, r2
	add.lt	@CPI0_0[0], r0, r2
	and	@CPI0_0[0], r1, r3
	sub!	r3, r0, r1
	add	r0, r0, r1
	add.gt	@CPI0_0[0], r0, r1
	sub.s!	@CPI0_0[0], r3, r3
	add.ne	r2, r0, r1
	sub!	r1, r0, r1
	add	0, r0, r1
	add.ne	1, r0, r1
	ret
```

hedgar2017:
@vladimirradosavljevic why the precompiles' size suffered so much? :(

vladimirradosavljevic:
> @vladimirradosavljevic why the precompiles' size suffered so much? :(

In `p256verify`, before this patch we had following case:
```
BB1:
...       <---- computation
// fallthrough
BB2:
...       <---- reused computation between BB1 and BB3
BB3:
...       <---- computation
jump BB2
```
After this patch we have following case:
```
BB1:
...       <---- computation
BB3:
...       <---- computation
```
The reason for the observed size degradation after this patch is that BB1 and BB3 are no longer similar. The assigned physical registers are different, which prevents us from creating BB2 and reusing computations between BB1 and BB3.
I suspect that for the same reason, we see degradations in #494 and #521.

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
```diff
@@ -58,55 +58,35 @@ char EraVMTieSelectOperands::ID = 0;
...
-/// try to emit a new select instruction which contains an implicit tie
-/// so that the register allocator can coalesce both registers.
-/// Return true if a new instruction was emitted.
+/// Try to create an implicit tie so that the register allocator can coalesce
+/// both registers.
+/// Return true if we managed to do so.
 bool EraVMTieSelectOperands::tryPlacingTie(MachineInstr &MI,
                                            EraVM::ArgumentKind Arg) const {
   assert(Arg == EraVM::ArgumentKind::In0 || Arg == EraVM::ArgumentKind::In1);
-  // Given that we've already made constraints on the case of single
```
akiramenai, line 67: We still handle immediate (not CP) case in td, so the comment is to be updated, not removed.

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
```diff
@@ -58,55 +58,35 @@ char EraVMTieSelectOperands::ID = 0;
...
-  // register operand SELECTs, we will now focus only on the case of two
-  // register selects. Also note that majority of the opportunities come
-  // from reg-reg selects.
-  if (MI.getOpcode() != EraVM::SELrrr)
+  if (MI.getOpcode() != EraVM::SELrrr &&
```
akiramenai: Nit: this is hard to follow. I suggest to extract a function which checks applicability of tie placement, name it properly (e.g. `canPlaceTie`) and document with an updated comment from above.

llvm/test/CodeGen/EraVM/selectcc.ll
```diff
@@ -1,3 +1,4 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
```
akiramenai, line 1: What is your motivation for replacing tests with minimal numbers of checks with auto-generated ones?

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
vladimirradosavljevic: Added comment to the newly introduced function.

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
vladimirradosavljevic: Extracted this code to the new function.

llvm/test/CodeGen/EraVM/selectcc.ll
vladimirradosavljevic: I chose to generate checks using a command instead of writing them manually because it's more convenient. If you prefer, I can revert this change, but it could make future modifications easier.

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
```diff
@@ -58,55 +64,42 @@ char EraVMTieSelectOperands::ID = 0;
 INITIALIZE_PASS(EraVMTieSelectOperands, DEBUG_TYPE,
                 ERAVM_TIE_SELECT_OPERANDS_NAME, false, false)
 
-/// try to emit a new select instruction which contains an implicit tie
-/// so that the register allocator can coalesce both registers.
-/// Return true if a new instruction was emitted.
+bool EraVMTieSelectOperands::canPlaceTie(MachineInstr &MI,
+                                         EraVM::ArgumentKind Arg) const {
+  return MI.getOpcode() == EraVM::SELrrr ||
```
akiramenai, line 69: Nit: we can also add a check that output is not tied yet as early return to this function.

akiramenai (approved):
LGTM

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
vladimirradosavljevic: If we do that before the opcode check, then we would assume that all instructions have at least one operand and the 0 operand is a definition. This assumption can be wrong for common opcodes like KILL, and we would hit an assert.

PR #507, opened 2024-04-30
[EraVM][CodeEmitter] Emit arithmetics and pointer arithmetics

atrosinenko:
Support emission of instructions requiring opcode fixup to account for various kinds of stack operands (i.e. the instructions with `in_any` or `out_any` operands).


llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
```diff
@@ -60,15 +66,66 @@ class EraVMMCCodeEmitter : public MCCodeEmitter {
...
   void encodeInstruction(const MCInst &MI, raw_ostream &OS,
                          SmallVectorImpl<MCFixup> &Fixups,
                          const MCSubtargetInfo &STI) const override;
+
+private:
+  uint64_t getRegWithAddend(const MCInst &MI, unsigned BaseReg,
+                            int Addend) const;
 };
 
+int modeEncodingByMarker(const MCOperand &Op) {
```
akiramenai: Can it be `static`?

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
```diff
@@ -60,15 +66,66 @@ class EraVMMCCodeEmitter : public MCCodeEmitter {
...
+  unsigned MarkerReg = Op.getReg();
+  assert(MarkerReg == EraVM::R0 || MarkerReg == EraVM::SP);
+  return (MarkerReg == EraVM::SP) ? EraVM::ModeSpRel : EraVM::ModeSpMod;
+}
+
+uint64_t EraVMMCCodeEmitter::fixOpcode(const MCInst &MI, uint64_t EncodedInstr,
+                                       const MCSubtargetInfo &STI) const {
+  const MCInstrDesc &Desc = MCII.get(MI.getOpcode());
+  auto GetOpIndex = [&Desc](unsigned UseIndex) {
+    return EraVM::getLogicalOperandIdx(Desc.getOpcode(),
```
akiramenai, line 107: @atrosinenko could you navigate me to the code which produces `getLogicalOperandIdx`, `getLogicalOperandSize`?
The latter is generated to
```
LLVM_READONLY static unsigned
getLogicalOperandSize(uint16_t Opcode, uint16_t LogicalOpIdx) {
   return LogicalOpIdx;
}
```
Which doesn't seem to be correct if I understand the meaning of these functions.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
```diff
@@ -60,15 +66,66 @@ class EraVMMCCodeEmitter : public MCCodeEmitter {
                         SmallVectorImpl<MCFixup> &Fixups,
                         const MCSubtargetInfo &STI) const;
 
+  uint64_t fixOpcode(const MCInst &MI, uint64_t EncodedInstr,
```
akiramenai: Documentation is welcome here.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
atrosinenko: @akiramenai This function is tablegen-erated when `let UseLogicalOperandMappings = true` is set in the `Instruction` class (in our case - it is set by this particular patch in the `IForm` class in `EraVMInstrFormats.td`).

To get the correct implementation, you need to build at least once after checking this revision out - but I agree the "default" implementation (and its presence at all) looks strange.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
atrosinenko: Sure. I thought it is inside an anonymous namespace, but it is not. Fixed in 2d74a1a6e5e84a592b0bcd1ff8480abd17f4e391.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
atrosinenko: Added an explanation and gave the function more descriptive name in 8b42a8d4ca7550cb137fc131fa52057037df1847.

atrosinenko:
Rebased and squashed (links to the particular fixes are in comments).

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
```diff
@@ -60,15 +66,77 @@ class EraVMMCCodeEmitter : public MCCodeEmitter {
...
+    const MCInst &MI, uint64_t EncodedInstr, const MCSubtargetInfo &STI) const {
+  const MCInstrDesc &Desc = MCII.get(MI.getOpcode());
+  auto GetOpIndex = [&Desc](unsigned UseIndex) {
+    return EraVM::getLogicalOperandIdx(Desc.getOpcode(),
+                                       Desc.getNumDefs() + UseIndex);
+  };
+
+  int OldSrcMode = -1;
+  int OldDstMode = -1;
+  int EncodedOpcode = EncodedInstr & EraVM::EncodedOpcodeMask;
```
akiramenai, line 116: `uint64_t` for opcodes and modes will make cast below unnecessary, won't it?

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCTargetDesc.h
```diff
@@ -50,6 +51,14 @@ MCAsmBackend *createEraVMMCAsmBackend(const Target &T,
 
 std::unique_ptr<MCObjectTargetWriter> createEraVMELFObjectWriter(uint8_t OSABI);
 
+struct EraVMOpcodeInfo {
```
akiramenai, line 71: Could we add that it represents the table `EraVMOpcodesList` from `EraVMOpcodes.td` and that fields description are there (or perhaps it worth to duplicate their meaning)?

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCCodeEmitter.cpp
atrosinenko: Not sure if possible widening of negative `int` to `uint64_t` in the code below can hurt, but let's use the same type everywhere to not even analyze this. Added a comment in cbaafc9ec0600213758768b86f0b384979f4d5fc.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMMCTargetDesc.h
atrosinenko: Added an extensive description in c42552c8301676cb065fad63ca4d3c1ad0689324 as the meaning of these fields are indeed tricky.

akiramenai: Thank you! It's very helpful.

akiramenai (approved):

atrosinenko:
Fixed clang-tidy warning and made `findOpcodeInfo` assert on unexpected input instead of returning `nullptr` (69541b9b523cd0ea13609e056917eb9283d91602): now it should be impossible due to sentinel value with `BaseOpcode == 0` and any `Opcode` should be accepted when disassembling user input.

PR #509, opened 2024-04-30
[EraVM][MC] Support encoding symbol references

atrosinenko:
Describe binary encoding for symbol-only operands (namely, jump targets) -
this makes corresponding instructions supported by both code emitter
and disassembler.

Implement emission of relocations for jump targets as well as code- and
stack-referring operands.


PavelKopyl (approved):
LGTM.

atrosinenko:
Fixed clang-tidy warning.

PR #510, opened 2024-04-30
[EraVM][LLD] Basic support for resolution of relocations

atrosinenko:


lld/ELF/Arch/EraVM.cpp
```diff
@@ -0,0 +1,75 @@
...
+  assert(definedSym && "Defined rel.sym expected");
+  uint64_t addressInSection =
+      definedSym->getVA(rel.addend) - definedSym->section->getVA();
+
+  auto add16scaled = [loc, &rel](uint64_t value, uint64_t scale) {
+    uint64_t scaledValue = value / scale;
+    checkAlignment(loc, value, scale, rel);
+    checkIntUInt(loc, scaledValue, 16, rel);
+    write16be(loc, read16be(loc) + scaledValue);
+  };
```
PavelKopyl: As I see the 'val' param is not used in the relocate(). To get in the context, could you please, describe how the `R_ERAVM_16_SCALE*` relocations are supposed to be resolved. If you can also remind in which cases the compiler emits these relocations, that would be great.

atrosinenko: > As I see the 'val' param is not used in the relocate().

In my understanding, it was easier to recalculate the value I had to use as `definedSym->getVA(rel.addend) - definedSym->section->getVA()`, but I will recheck. Anyway, this initial implementation has to be fine-tuned according to expectations of the VM, I guess.

> To get in the context, could you please, describe how the R_ERAVM_16_SCALE* relocations are supposed to be resolved. If you can also remind in which cases the compiler emits these relocations, that would be great.

At now, these are only explained in #509 (see [EraVMFixupKinds.h](https://github.com/matter-labs/era-compiler-llvm/pull/509/files#diff-729b54ebf8139b7ea94470c9ef5435be7bfbc4c3c0c748ee43d3c6d6f58257eaR23), these two fixups corespond to `R_ERAVM_16_SCALE*` relocations). These fixups are generated by [EraVMMCCodeEmiter.cpp](https://github.com/matter-labs/era-compiler-llvm/pull/509/files#diff-6ca79aa8d678c0e2fd9c139a8e4fbdf9e51c8501e516f3cf3eb4ee4fefa7156eR176).

PavelKopyl: Thank you for the clarification. The only thing I'm worrying about is `addressInSection`: we calculate a symbol address relatively to its input section. If we are going to link at least two objects files, I suppose, we need to calculate offsets , for example, relative to the output .text segment.

PavelKopyl:
Please, move lld-unrelated changes to another PR.

lld/ELF/Arch/EraVM.cpp
PavelKopyl: As for now we can ignore this issue, because we do not need to support linking of several object files.

asl:
@antonbaliasnikov Will you please take a look at https://github.com/matter-labs/era-compiler-llvm/actions/runs/8979795940/job/24662547336 ?

antonbaliasnikov:
> @antonbaliasnikov Will you please take a look at https://github.com/matter-labs/era-compiler-llvm/actions/runs/8979795940/job/24662547336 ?

@asl , CI should be fixed now. The additional fix is in the https://github.com/matter-labs/era-compiler-llvm/pull/528. An issue with directories permissions on self-hosted runners. Thanks for reporting!

lld/ELF/Arch/EraVM.cpp
atrosinenko: Thank you for pointing this out, removed the hardcoded re-computation of `val` and switched tests to using custom linker script at now in bf943f0341fe6b409f41714d001e99b7bb240af9.

atrosinenko:
> Please, move lld-unrelated changes to another PR.

@PavelKopyl These changes probably belonged to the stacked reviews this PR depended on - in that case they disappeared after rebase.

lld/ELF/Arch/EraVM.cpp
atrosinenko: Created a separate issue on fine-tuning the implicit linker script: #529.

PavelKopyl (approved):
LGTM.

akiramenai (approved):

PR #515, opened 2024-05-03
[EraVM] Rework EraVMTargetStreamer class

atrosinenko:
Separate ELF- and asm-specific logic, remove unused field.

In AsmParser code, replace MCAsmStreamer::emitIntValue() override with a better target-specific solution.


atrosinenko:
The test here is a simplified and slightly improved `encoding/data.s` from #509.

asl (approved):

PR #521, opened 2024-05-06
Restore the "Fold overflow intrinsics with select instructions"

TerryGuo:
Restore this reverted patch along with two changes:

# Code Review Checklist

## Requirements

## Implementation
- [x] Does this code change accomplish what it is supposed to do?
- [x] Can this solution be simplified?
- [x] Does this change add unwanted compile-time or run-time dependencies?
- [x] Could an additional framework, API, library, or service improve the solution?
- [x] Could we reuse part of LLVM instead of implementing the patch or a part of it?
- [x] Is the code at the right abstraction level?
- [x] Is the code modular enough?
- [x] Can a better solution be found in terms of maintainability, readability, performance, or security?
- [x] Does similar functionality already exist in the codebase? If yes, why isn’t it reused?
- [x] Are there any best practices, design patterns or language-specific patterns that could substantially improve this code? 

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


llvm/lib/Target/EraVM/MCTargetDesc/EraVMInstPrinter.cpp
```diff
@@ -100,6 +100,9 @@ void EraVMInstPrinter::printCCOperand(const MCInst *MI, unsigned OpNo,
   case EraVMCC::COND_LE:
     O << ".le";
     break;
+  case EraVMCC::COND_OF:
```
asl: You also need to ensure this is properly encoded / parsed, not just printed.

Tagging @akiramenai @sayon for spec update

asl (changes requested):

llvm/lib/Target/EraVM/MCTargetDesc/EraVMInstPrinter.cpp
akiramenai: @TerryGuo could we not introduce this flag, OF is an alias of LT. Let's use LT.

TerryGuo: Hi @asl Thanks for comments. @akiramenai Sure, I will update patch to use LT.

TerryGuo:
Hi @akiramenai , what do you think of the score? They don't look good to me. For example we got "Worst -19.108" for "Size All M3B3", I guess this means my patch caused a code size bloating. As for Cycles, the mean value is pretty small. Do they match your expectations?

TerryGuo:
Hi @akiramenai , what do you think of the score? They don't look good to me. For example, we got "Worst -19.108" for "Size All M3B3", I guess this means my patch caused a code size bloating. As for Cycles, the mean value is pretty small. Do they match your expectations?

TerryGuo:
Hi @akiramenai , what do you think of the score? They don't look good to me. For example, we got "Worst -19.108" for "Size All M3B3", I guess this means my patch caused a code size bloating. As for Cycles, the mean value is pretty small. Do they match your expectations?

akiramenai:
The numbers are expected. Perhaps except for size regressions.
A similar logic for branches expected to give us more.

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
```diff
@@ -1029,6 +1045,38 @@ SDValue EraVMTargetLowering::LowerBR_JT(SDValue Op, SelectionDAG &DAG) const {
...
+  int OFGlueResult = 2;
+  // MUL has 2 results
+  if (LoweredOpc == EraVMISD::MUL_V) {
+    FoldedVT = DAG.getVTList(MVT::i256, MVT::i256, MVT::Other, MVT::Glue);
+    OFGlueResult = 3;
+  }
+  FoldedArith = DAG.getNode(LoweredOpc, DL, FoldedVT,
+                            {DAG.getEntryNode(), MatchedUArithO.getOperand(0),
+                             MatchedUArithO.getOperand(1)});
+  // DAG.ReplaceAllUsesOfValueWith(MatchedUArithO.getValue(0),
```
akiramenai: ?

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
```diff
@@ -1029,6 +1045,38 @@ SDValue EraVMTargetLowering::LowerBR_JT(SDValue Op, SelectionDAG &DAG) const {
   return SDValue(Res, 0);
 }
 
+SDValue EraVMTargetLowering::LowerSELECT(SDValue Op, SelectionDAG &DAG) const {
+  // try to fold select with u{add|sub|mul}.with.overflow
```
akiramenai: Could we document the transformation better?

llvm/test/CodeGen/EraVM/overflow.ll
```diff
@@ -0,0 +1,80 @@
...
+declare { i256, i1 } @llvm.uadd.with.overflow.i256(i256, i256)
+declare { i256, i1 } @llvm.usub.with.overflow.i256(i256, i256)
+declare { i256, i1 } @llvm.umul.with.overflow.i256(i256, i256)
+
+; CHECK-LABEL: add_test
+define i256 @add_test(i256 %a, i256 %b, i256 %x, i256 %y) {
+entry:
+  %res1 = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 %y)
+; CHECK:      add!    r3, r4, r{{[0-9]+}}
+; add.ge is add.of reversed
```
akiramenai: This is generally a wrong statement. The correct one is that `add!`, `sub!`, `mul!` set flags so if `LT` or `EQ` is set `GT` is reset and vice versa, so for `uaddo`, `umulo`, `usubo` the select condition is indeed reversible.

akiramenai (changes requested):
Could you please navigate me to the code which guarantees that if glue contradicts to data dependency a copy of a node is inserted? We need to make sure that testing is sufficient.
You also mentioned that there are cases with suboptimal codegen, could you add them to the regression tests as well?

llvm/test/CodeGen/EraVM/overflow.ll
akiramenai: Oh, worse. My statement is wrong too.
The flag is not reversible.
https://matter-labs.github.io/eravm-spec/spec.html#AddDefinition says that `lt` is set when overflow, `eq` is set when the result is `0`. So both overflow and `0` result could be produced at the same time, thus `ge` doesn't identify that the overflow hasn't happened.
@sayon Am I right?

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
TerryGuo: Comments added now.

llvm/lib/Target/EraVM/MCTargetDesc/EraVMInstPrinter.cpp
TerryGuo: OF is replaced with LT now.

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
TerryGuo: The document is improved now.

llvm/test/CodeGen/EraVM/overflow.ll
TerryGuo: > Could you please navigate me to the code which guarantees that if glue contradicts to data dependency a copy of a node is inserted? We need to make sure that testing is sufficient.

I tried several times and failed. It makes me think it is not easy and straightforward to detect the case you mentioned in our slack discussion. In order to properly handle this "glue contradicts to data dependency" case, I modified the original patch by not calling the `DAG.ReplaceAllUsesOfValueWith` function for the newly created `FoldedArith` instruction, hence we will end up with two arithmetic instructions always, the original one will be kept for generating result, the second one `FoldedArith` will be used for generating flag for SELECT (the GLUE is used here to prevent DAGSchedule from inserting another flag-setting instruction between them).

Case 1: If the result of the first overflow arithmetic isn't used anywhere, then it will be optimized out. The final code is optimal.
Case 2: For the case you mentioned in Slack, we need those two arithmetic instructions anyway, so the final code is expected.
Case 3: This is the sub-optimal case. If there is just one SELECT instruction and no other flag-setting instruction and the result of the overflow arithmetic is used, then we still get two arithmetic instructions with this patch since we are not calling `DAG.ReplaceAllUsesOfValueWith`.

llvm/lib/Target/EraVM/AsmParser/EraVMAsmParser.cpp
```diff
@@ -328,6 +328,7 @@ static int parseExplicitCondition(StringRef Code) {
       .Case("ne", EraVMCC::COND_NE)
       .Case("ge", EraVMCC::COND_GE)
       .Case("le", EraVMCC::COND_LE)
+      .Case("of", EraVMCC::COND_LT)
```
akiramenai: We don't plan to support this alias in asm.

akiramenai (changes requested):
Overall the approach with a pseudo flag looks ok to me, but the tests imply that it needs to be addressed in some other places as well.
I expect the flag to be pure pseudo (with no exposure to our asm) and carry semantic of irreversibility in Select and Br  which is to be documented in appropriate places with examples. Maybe there is a better way to get the same behaviour w.o. introducing new a pseudo flag [@vladimirradosavljevic perhaps you have some ideas on that]

TerryGuo:
> Overall the approach with a pseudo flag looks ok to me, but the tests imply that it needs to be addressed in some other places as well. I expect the flag to be pure pseudo (with no exposure to our asm) and carry semantic of irreversibility in Select and Br which is to be documented in appropriate places with examples. Maybe there is a better way to get the same behaviour w.o. introducing new a pseudo flag [@vladimirradosavljevic perhaps you have some ideas on that]

Hi @akiramenai , thanks for the comments, I will spend more time thinking about how to implement a better one w.o. this new pseudo flag. A possible idea is to lower `oveflow instrinsic + selection` directly to conditional moves.
Hi @vladimirradosavljevic , if you have any thoughts, please share. Thanks in advance.

llvm/lib/Target/EraVM/AsmParser/EraVMAsmParser.cpp
TerryGuo: Removed.

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
```diff
@@ -1029,6 +1045,63 @@ SDValue EraVMTargetLowering::LowerBR_JT(SDValue Op, SelectionDAG &DAG) const {
...
+  int OFGlueResult = 2;
+  // MUL has 2 results
+  if (LoweredOpc == EraVMISD::MUL_V) {
+    FoldedVT = DAG.getVTList(MVT::i256, MVT::i256, MVT::Other, MVT::Glue);
+    OFGlueResult = 3;
+  }
+  FoldedArith = DAG.getNode(LoweredOpc, DL, FoldedVT,
+                            {DAG.getEntryNode(), MatchedUArithO.getOperand(0),
+                             MatchedUArithO.getOperand(1)});
+  // Deliberately comment out this call in the original patch to support below
```
akiramenai: I think it's better to avoid referencing the original patch as not everybody should know history of our project.
I suggest to put it this way:
Note that we create a new node while keeping the original one. This is because of the following case.
<example>
We need to keep...

llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -231,6 +231,7 @@ def COND_E    : EraVMCondCode<3>;
 def COND_GE   : EraVMCondCode<4>;
 def COND_LE   : EraVMCondCode<5>;
 def COND_NE   : EraVMCondCode<6>;
+def COND_OF   : EraVMCondCode<8>;
```
akiramenai, line 230: Please document here that it's a pseudo flag, alias of LT and that we added it to distinguish between `lt` set by `sub` instruction which is reversible and `lt` set to indicate overflow by `add`, `mul` which aren't.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -96,6 +98,15 @@ def EraVMlog_decommit : SDNode<"EraVMISD::LOG_DECOMMIT", SDT_EraVMLogDecommit, [
 def EraVMTrap : SDNode<"EraVMISD::TRAP", SDTNone,
     [SDNPHasChain, SDNPOptInGlue]>;
 
+// Those instructions will set flag register. The DAG schedule shouldn't schedule another
+// flag-setting instruction betweem them and their users. The original patch uses Glue to
```
akiramenai: Please rephrase to not to mention the original patch.

akiramenai (commented):
@vladimirradosavljevic could you also have a look?

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -1248,3 +1259,95 @@ def : Pat<(EraVMTrap), (PANICr R0)>;
 
 def : Pat<(EraVMselectcc GRPTR:$rs0, GRPTR:$rs1, imm:$cc),
           (FATPTR_SELrrr GRPTR:$rs0, GRPTR:$rs1, imm:$cc)>;
+
+// ============================ Overflow ADDs =====================================
```
vladimirradosavljevic, line 1284: Patterns for `large_imm` are missing.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -1248,3 +1259,95 @@ def : Pat<(EraVMTrap), (PANICr R0)>;
...
+def : Pat<(store_stack (EraVMAdd_v GR256:$rs0, (load_code memaddr:$rs1)), stackaddr:$stack),
+          (ADDcrs_v memop:$rs1, GR256:$rs0, stackin:$stack, 0)>;
+def : Pat<(store_stack (EraVMAdd_v (load_code memaddr:$rs0), GR256:$rs1), stackaddr:$stack),
+          (ADDcrs_v memop:$rs0, GR256:$rs1, stackin:$stack, 0)>;
+def : Pat<(store_stack (EraVMAdd_v (load_stack stackaddr:$rs0), GR256:$rs1), stackaddr:$stack),
+          (ADDsrs_v stackin:$rs0, GR256:$rs1, stackin:$stack, 0)>;
+def : Pat<(store_stack (EraVMAdd_v GR256:$rs0, (load_stack stackaddr:$rs1)), stackaddr:$stack),
+          (ADDsrs_v stackin:$rs1, GR256:$rs0, stackin:$stack, 0)>;
+
+// ============================ Overflow SUBs =====================================
```
vladimirradosavljevic, line 1317: Patterns for large_imm are missing.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -1248,3 +1259,95 @@ def : Pat<(EraVMTrap), (PANICr R0)>;
...
+          (SUBsrs_v stackin:$rs0, GR256:$rs1, stackin:$stack, 0)>;
+
+def : Pat<(store_stack (EraVMSub_v GR256:$rs0, imm16:$rs1), stackaddr:$stack),
+          (SUBxrs_v imm16:$rs1, GR256:$rs0, stackin:$stack, 0)>;
+def : Pat<(store_stack (EraVMSub_v GR256:$rs0, (load_code memaddr:$rs1)), stackaddr:$stack),
+          (SUByrs_v memop:$rs1, GR256:$rs0, stackin:$stack, 0)>;
+def : Pat<(store_stack (EraVMSub_v GR256:$rs0, (load_stack stackaddr:$rs1)), stackaddr:$stack),
+          (SUBzrs_v stackin:$rs1, GR256:$rs0, stackin:$stack, 0)>;
+
+// ============================ Overflow ADDs =====================================
```
vladimirradosavljevic: Typo, it should be Overflow `MULs` and patterns for large_imm are missing.

llvm/lib/Target/EraVM/EraVMExpandSelect.cpp
```diff
@@ -134,13 +135,10 @@ bool EraVMExpandSelect::runOnMachineFunction(MachineFunction &MF) {
 
       if (ShouldInverse) {
         buildMOV(EraVM::ArgumentKind::In0, EraVMCC::COND_NONE);
```
vladimirradosavljevic, line 139: Please add here assert that we have an InverseCond for CCVal.

llvm/lib/Target/EraVM/EraVMTargetMachine.cpp
```diff
@@ -292,8 +292,15 @@ void EraVMPassConfig::addPreRegAlloc() {
...
-    addPass(createEraVMTieSelectOperandsPass());
   }
+  // Since some conditional codes like overflow LT hasn't reversal version (the
+  // GE isn't suitable according to spec), we have to rely on this pass to tie
+  // the output with the second input operand (the false part), otherwise if we
+  // get below IR, we will be unable to expand it correctly:
+  //   $r1 = SELrrr killed $r1, killed $r6, COND_OF, implicit $flags, implicit
+  //   $r1(tied-def 0)
+  // Therefor we enable it for all optimization level.
+  addPass(createEraVMTieSelectOperandsPass());
```
vladimirradosavljevic: Please don't make this pass available for all optimization levels. This is pass to optimize select patterns and it is not pass where we want to resolve functional problems. This problem is present in all SELr variants, not only SELrrr. Please check the explanation how this should be addressed.

llvm/test/CodeGen/EraVM/overflow.ll
```diff
@@ -0,0 +1,117 @@
+; RUN: llc < %s | FileCheck %s
```
vladimirradosavljevic, line 2: Please add tests with all variants of select instructions, since here we are only checking for few cases. Also, please add pre-commit test without changes, and after that this support to see the actual improvements (E.g. pre-commit https://github.com/matter-labs/era-compiler-llvm/commit/0604f9f7726d393aa602f3c40b931a0c34885823, actual commit https://github.com/matter-labs/era-compiler-llvm/commit/7311da6251e108301d1d9b545e1386f03313f838). Don't manually create checks, please use `update_llc_test_checks.py` to do so:
`llvm/llvm/utils/update_llc_test_checks.py --llc-binary target-llvm/build-final/bin/llc llvm/llvm/test/CodeGen/EraVM/overflow.ll`

vladimirradosavljevic (changes requested):
There is no support and tests for optimizations where we are manipulating/folding with select instructions (e.g. like in `EraVMOptimizeSelectPreRA`). Please add check not to do this optimization for `.of` if we need to reverse condition, and add tests to check these optimization where we can do it and where we can't do it.

As I mentioned in the comment, problem with ties is not only present with SELrrr instruction. Please look at the following test:
```
define i256 @test(i256 %a, i256 %b, i256 %x, i256 %y) {
entry:
  %res1 = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 %y)
  %overflow = extractvalue {i256, i1} %res1, 1
  %selected = select i1 %overflow, i256 %a, i256 12345
  ret i256 %selected
}
```
Following assembly is generated with this patch:
```
test:                                   ; @test
; %bb.0:                                ; %entry
    add!    r3, r4, r2
    add     12345, r0, r1
    ret
```
In this case, we have:
```
%5:gr256 = SELrir killed %0:gr256(tied-def 0), i256 12345, i256 8, implicit $flags
```
And as you can see, first register is tied with the output and later in the expand select we would run into the same problem as with SELrrr (please note in the generated ASM we have 1 instruction after expansion, but that is the problem in `ShouldInverse` where we don't have inverse condition for `.of`, so I added a comment to add assert so we can detect these cases).

In order to fix this, we need to redesign how we are tying register in select instruction:
1. Remove all tie constraints from select instructions from `EraVMInstrInfo.td` and move tie handling to `EraVMTieSelectOperands` pass. Create a separate PR for this to check the numbers, since I expect we will get some differences, as now we would only tie killed input registers with output.
2. Include changes from 1. here, and for all SELr variants add `let hasPostISelHook = 1` in `EraVMInstrInfo.td`. In `EraVMTargetLowering::AdjustInstrPostInstrSelection` for all SELr variants with `.of` condition, mark output with `IsEarlyClobber` flag (call `setIsEarlyClobber()`) to prevent regalloc from assigning the same physical register for input and output. In `EraVMTieSelectOperands`, enable removing of  `IsEarlyClobber` for all select instructions, not only for `SELcrr` and `SELrcr`.
3. When this is addressed, move setting of `IsEarlyClobber` from `EraVMBytesToCells.cpp` (https://github.com/matter-labs/era-compiler-llvm/commit/e91f94b406af015a53dc8068766ecc4afb7fc9fc) to `EraVMTargetLowering::AdjustInstrPostInstrSelection` for all select instructions that have stack and code operands in which register is used (you can look at the positions where register should be used in `EraVMBytesToCells::convertStackMachineInstr` and `EraVMBytesToCells::convertCodeAddressMachineInstr`). This is to improve readability and maintainability, because we will have only one place where we are adding `IsEarlyClobber` for functional reasons.

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
TerryGuo: Got it and will reword.

llvm/lib/Target/EraVM/EraVMInstrFormats.td
TerryGuo: Will add comments.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
TerryGuo: Rephrased.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
TerryGuo: Hi @vladimirradosavljevic , thanks for your comments. I checked the `defm ADD  : ArithICommutable<OpAdd, add>;` and found there is no case using `large_imm`. So it looks to me I can't expand ADD_V with `large_imm` with ADD_xxx with `large_imm ` directly unless I add `large_imm` case into `ArithICommutable`. Is this understanding correct?

TerryGuo: Hi @vladimirradosavljevic , I create a separate [PR](https://github.com/matter-labs/era-compiler-llvm/pull/550) for this item

> Remove all tie constraints from select instructions from EraVMInstrInfo.td and move tie handling to EraVMTieSelectOperands pass. Create a separate PR for this to check the numbers, since I expect we will get some differences, as now we would only tie killed input registers with output.

vladimirradosavljevic: `large_imm` is placed into constant pool, so you will have to do something like this:
```
def : Pat<(EraVMAdd_v GR256:$rs0, large_imm:$rs1),
          (ADDcrr_v (constant_pool imm:$rs1), 0, GR256:$rs0, 0)>;
```

TerryGuo:
> In order to fix this, we need to redesign how we are tying register in select instruction:
> 
> 1. Remove all tie constraints from select instructions from `EraVMInstrInfo.td` and move tie handling to `EraVMTieSelectOperands` pass. Create a separate PR for this to check the numbers, since I expect we will get some differences, as now we would only tie killed input registers with output.
> 2. Include changes from 1. here, and for all SELr variants add `let hasPostISelHook = 1` in `EraVMInstrInfo.td`. In `EraVMTargetLowering::AdjustInstrPostInstrSelection` for all SELr variants with `.of` condition, mark output with `IsEarlyClobber` flag (call `setIsEarlyClobber()`) to prevent regalloc from assigning the same physical register for input and output. In `EraVMTieSelectOperands`, enable removing of  `IsEarlyClobber` for all select instructions, not only for `SELcrr` and `SELrcr`.
> 3. When this is addressed, move setting of `IsEarlyClobber` from `EraVMBytesToCells.cpp` ([e91f94b](https://github.com/matter-labs/era-compiler-llvm/commit/e91f94b406af015a53dc8068766ecc4afb7fc9fc)) to `EraVMTargetLowering::AdjustInstrPostInstrSelection` for all select instructions that have stack and code operands in which register is used (you can look at the positions where register should be used in `EraVMBytesToCells::convertStackMachineInstr` and `EraVMBytesToCells::convertCodeAddressMachineInstr`). This is to improve readability and maintainability, because we will have only one place where we are adding `IsEarlyClobber` for functional reasons.


Hi @vladimirradosavljevic , if we take out the COND_OF part, I think your above comments can be treated as a new design of how we tie the operands and set the early-clobber attribute. Therefor I create a separate PR #553  to address them (without the COND_OF), would you please give it a look?

llvm/test/CodeGen/EraVM/overflow.ll
sayon: Sorry for late reply -- indeed this is correct reasoning.
If we add 1 to $2^{256}-1$ the result will be 0 with an overflow, therefore both LT/OF flag and EQ flag will be set.

vladimirradosavljevic:
> > In order to fix this, we need to redesign how we are tying register in select instruction:
> > 
> > 1. Remove all tie constraints from select instructions from `EraVMInstrInfo.td` and move tie handling to `EraVMTieSelectOperands` pass. Create a separate PR for this to check the numbers, since I expect we will get some differences, as now we would only tie killed input registers with output.
> > 2. Include changes from 1. here, and for all SELr variants add `let hasPostISelHook = 1` in `EraVMInstrInfo.td`. In `EraVMTargetLowering::AdjustInstrPostInstrSelection` for all SELr variants with `.of` condition, mark output with `IsEarlyClobber` flag (call `setIsEarlyClobber()`) to prevent regalloc from assigning the same physical register for input and output. In `EraVMTieSelectOperands`, enable removing of  `IsEarlyClobber` for all select instructions, not only for `SELcrr` and `SELrcr`.
> > 3. When this is addressed, move setting of `IsEarlyClobber` from `EraVMBytesToCells.cpp` ([e91f94b](https://github.com/matter-labs/era-compiler-llvm/commit/e91f94b406af015a53dc8068766ecc4afb7fc9fc)) to `EraVMTargetLowering::AdjustInstrPostInstrSelection` for all select instructions that have stack and code operands in which register is used (you can look at the positions where register should be used in `EraVMBytesToCells::convertStackMachineInstr` and `EraVMBytesToCells::convertCodeAddressMachineInstr`). This is to improve readability and maintainability, because we will have only one place where we are adding `IsEarlyClobber` for functional reasons.
> 
> Hi @vladimirradosavljevic , if we take out the COND_OF part, I think your above comments can be treated as a new design of how we tie the operands and set the early-clobber attribute. Therefor I create a separate PR #553 to address them (without the COND_OF), would you please give it a look?

We are doing redesign to address problem where we don't have reverse condition for `.of` and we are doing ties on the first register operand. 3rd point is just a refactoring to have only one place where we add earlyclobber flag. I would focus more on the 2nd point and other comments and do 3rd point when everything else is addressed.

llvm/test/CodeGen/EraVM/overflow.ll
TerryGuo: Patch is updated to avoid inversing overflow LT now.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
TerryGuo: `large_imm` is added now just like `imm16` .

llvm/lib/Target/EraVM/EraVMInstrInfo.td
TerryGuo: `large_imm` is added now just like `imm16` .

llvm/lib/Target/EraVM/EraVMInstrInfo.td
TerryGuo: Correct it now.

llvm/lib/Target/EraVM/EraVMExpandSelect.cpp
TerryGuo: Added assert.

llvm/lib/Target/EraVM/EraVMTargetMachine.cpp
TerryGuo: Reverted this change now.

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
```diff
@@ -1339,4 +1411,22 @@ void EraVMTargetLowering::AdjustInstrPostInstrSelection(MachineInstr &MI,
...
+  // for output and input operands, otherwise we may get below case which can't
+  // be expanded correctly:
+  //
+  // out = SELrrr i1 %ov, in1, in2, COND_OF
+  //
+  // where out and in1 are assigned to same register.
+  if (EraVM::isSelect(MI) &&
+      getImmOrCImm(*EraVM::ccIterator(MI)) == EraVMCC::COND_OF) {
+    if (EraVM::out0Iterator(MI)->isReg()) {
+      EraVM::out0Iterator(MI)->setIsEarlyClobber();
```
vladimirradosavljevic: With this line, we will always add earlyclobber flag, but we need to check whether it is SELr variant of the select. In this case, we could use `argumentType` for first input operand to do so.

TerryGuo: Hi @vladimirradosavljevic , do you mean something like below?
```c++
  if (EraVM::isSelect(MI) &&
      (EraVM::argumentType(EraVM::ArgumentKind::In0, MI.getOpcode()) ==
       EraVM::ArgumentType::Register) &&
      getImmOrCImm(*EraVM::ccIterator(MI)) == EraVMCC::COND_OF) {
    if (EraVM::out0Iterator(MI)->isReg()) {
      EraVM::out0Iterator(MI)->setIsEarlyClobber();
    } else if (EraVM::out1Iterator(MI)->isReg()) {
      EraVM::out1Iterator(MI)->setIsEarlyClobber();
    }
  }
```

vladimirradosavljevic: Yes, but I don't think we need to check whether the output is a register, since currently we only support SEL instructions with register as an output.

TerryGuo: Got it and thanks. Code is updated now per your comments.

llvm/test/CodeGen/EraVM/overflow.ll
TerryGuo: The "add pre-commit test without changes" part is new to me. I am trying to figure it out. The tests are not added, I am working on them. Meanwhile please keep reviewing the latest code. Thanks.

TerryGuo: I see now. Will address this part tomorrow.

TerryGuo: Hi @vladimirradosavljevic , I create a separate PR #560, would you please take a look?

vladimirradosavljevic:
> There is no support and tests for optimizations where we are manipulating/folding with select instructions (e.g. like in EraVMOptimizeSelectPreRA). Please add check not to do this optimization for .of if we need to reverse condition, and add tests to check these optimization where we can do it and where we can't do it.

@TerryGuo Could you please address this?
In optimizations where we are manipulating/folding with select instructions such as `EraVMOptimizeSelectPreRA` we need to disable them if we need to revert condition for `.of`. Please take a look at the following example:
```
define i256 @test1(i256 %a, i256 %b, i256 %x, i256 %y) {
entry:
  %add = add i256 %a, %b
  %res = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 %y)
  %overflow = extractvalue {i256, i1} %res, 1
  %selected = select i1 %overflow, i256 %a, i256 %add
  ret i256 %selected
}
```
Generated assembly:
```
test1:                                  ; @test1
; %bb.0:                                ; %entry
	add!	r3, r4, r3
	add	r1, r2, r1
	ret
```
As you can see, this optimization generated add with `COND_NONE` since there is no reverse condition for `.of`. We should disable this and optimizations that can emit reverse condition in case of `.of`.

TerryGuo:
I searched the EraVM folder and updated the code to quit when need to inverse the `COND_OF`. The corresponding tests are added.

llvm/test/CodeGen/EraVM/overflow.ll
TerryGuo: The pre-commit tests are added now.

llvm/test/CodeGen/EraVM/overflow.ll
```diff
@@ -0,0 +1,875 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
+; RUN: llc < %s | FileCheck %s
+
+target datalayout = "E-p:256:256-i8:256:256:256-i256:256:256-S32-a:256:256"
```
akiramenai: i8 must be naturally aligned.

llvm/test/CodeGen/EraVM/overflow.ll
```diff
@@ -0,0 +1,872 @@
...
+; CHECK-NEXT:    add r2, r0, r1
+; CHECK-NEXT:    ret
+entry:
+  %res1 = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 26959946660873538059280334323183841250350249843923952699046031785980)
+  %overflow = extractvalue {i256, i1} %res1, 1
+  %selected = select i1 %overflow, i256 %a, i256 %b
+  ret i256 %selected
+}
+
+define i256 @add_test_large_immx(i256 %a, i256 %b, i256 %x, i256 %y) {
```
akiramenai: `add` is commutable, so what's the point in this test?

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
```diff
@@ -1029,6 +1045,66 @@ SDValue EraVMTargetLowering::LowerBR_JT(SDValue Op, SelectionDAG &DAG) const {
...
+// If the condition of SELECT instruction comes from a overflow intrinsic, lower
+// it into conditional move.  A simple case is as below:
+//
+//   %res1 = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 %y)
+//   %overflow = extractvalue {i256, i1} %res1, 1
+//   %selected = select i1 %overflow, i256 %a, i256 %b
+//
+// which can be lowered to
+//
+//   add!	  %x, %y, r3
```
akiramenai: Please reformat with spaces

llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -224,6 +224,10 @@ def COND_GE      : EraVMCondCode<4>;
 def COND_LE      : EraVMCondCode<5>;
 def COND_NE      : EraVMCondCode<6>;
 def COND_GTOrLT  : EraVMCondCode<7>;
+// This is a pseudo flag, alias of LT and used to distinguish between
+// normal LT set by sub instruction which is reversible and LT set to
+// indicate overflow caused add/sub/mul which are not reversible.
```
akiramenai: add/mul

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -278,27 +289,25 @@ let isSelect = 1, Uses = [Flags] in {
   def SELirr : Pseudo<(outs GR256:$dst), (ins imm16:$src, GR256:$src2, i256imm:$cc),
                        [(set GR256:$dst,
                           (EraVMselectcc imm16:$src, GR256:$src2, imm:$cc))]> {
-                            let Constraints = "$dst = $src2";
+                            let hasPostISelHook = 1;
```
akiramenai: We only adjust `SELrxx` pseudos. Shouldn't we remove hooks where they aren't needed?
And then we can turn `EraVM::argumentType(EraVM::ArgumentKind::In0, MI.getOpcode()) == EraVM::ArgumentType::Register` check into an assertion.

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
```diff
@@ -86,6 +85,23 @@ bool EraVMTieSelectOperands::tryPlacingTie(MachineInstr &MI,
   if (MI.isRegTiedToUseOperand(0))
     return false;
 
+  // Since COND_OF hasn't reversal version, we have to tie the output register
```
akiramenai: Could we phase it more accurately? The logic we implemented guarantee that the 1st argument wound't be tied, but it doesn't necessary mean that the second one would. It might be not a register.
It's also worth showing the example for the normal case scenario - `COND_LT` to explain that normally we reverse condition to optimize if we can place tie to the first argument.

akiramenai (commented):
Thank you, @TerryGuo. It's always a pleasure to read your documentation. I added a few comments, please address them, then I approve.
Also please mention @lialan as a co-author in the commit message.

llvm/test/CodeGen/EraVM/overflow.ll
TerryGuo: `datalayout` is corrected now.

llvm/lib/Target/EraVM/EraVMISelLowering.cpp
TerryGuo: Reformatted now.

llvm/lib/Target/EraVM/EraVMInstrFormats.td
TerryGuo: Removed `sub`.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
TerryGuo: Good point. PR is adjusted per your suggestion.

llvm/test/CodeGen/EraVM/overflow.ll
TerryGuo: This case is removed now.

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
TerryGuo: Reworked the comments in code. Please review again.

akiramenai (approved):

PR #522, opened 2024-05-06
[EraVM] Add an assert to pacify Klocwork in EraVMBytesToCells

akiramenai:
`DefMI != nullptr` is guaranteed by SSA form, add an assertion as a sanity check.


FlashSheridan (approved):
LGTM, thanks.  There are similar issues in other files in CPR-1572; I think Linear will autoclose that when this PR is merged.  Should I manually reopen (and retitle) that when it happens?

akiramenai:
I will reopen by myself and address it shortly. Thanks for noticing.

PR #526, opened 2024-05-07
[EraVM] Remove i8 alignment from data layout in tests

akiramenai:
`i8` must be naturally aligned.

### DIFF
--- a/llvm/test/CodeGen/EraVM/ra_regress.ll
+++ b/llvm/test/CodeGen/EraVM/ra_regress.ll
@@ -1,8 +1,6 @@
 ; RUN: llc -O0 < %s | FileCheck %s
 
-; ModuleID = 'Test_65'
-source_filename = "Test_65"
-target datalayout = "E-p:256:256-i8:256:256:256-i256:256:256-S32-a:256:256"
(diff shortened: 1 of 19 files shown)


llvm/test/CodeGen/EraVM/ra_regress.ll
```diff
@@ -1,8 +1,6 @@
 ; RUN: llc -O0 < %s | FileCheck %s
 
-; ModuleID = 'Test_65'
-source_filename = "Test_65"
```
FlashSheridan, line 4: Is there a reason for removing `source_filename`?

FlashSheridan (commented):
Looks good to Alive2 on the original file, and to grep for the rest; details in CPR-1688.

llvm/test/CodeGen/EraVM/ra_regress.ll
akiramenai: Yes, it gives nothing in terms of understanding the test. Just distracts the reader.

FlashSheridan (approved):
LGTM, Alive2, & grep.

PR #536, opened 2024-05-12
[EraVM] Add support for select instructions in EraVMBytesToCells

vladimirradosavljevic:


llvm/test/CodeGen/EraVM/selectcc-stack-code-in-cells.ll
```diff
@@ -0,0 +1,153 @@
...
+  %load2 = load i256, ptr %s2
+  %select = select i1 %cond, i256 %load1, i256 %load2
+  ret i256 %select
+}
+
+define i256 @selcrr(i1 %cond, ptr addrspace(4) %c, i256 %v1) {
+; CHECK-LABEL: selcrr:
+; CHECK:       ; %bb.0:
+; CHECK-NEXT:    sub! r1, r0, r1
+; CHECK-NEXT:    add.ne code[r2], r0, r3
```
akiramenai: It doesn't seem to be correct as `r3` is undefined if `eq`.

akiramenai (changes requested):

llvm/test/CodeGen/EraVM/selectcc-stack-code-in-cells.ll
vladimirradosavljevic: `r3` should hold value of `%v1`.

akiramenai: Sorry, you're right

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
```diff
@@ -96,6 +96,17 @@ bool EraVMTieSelectOperands::tryPlacingTie(MachineInstr &MI,
...
     return false;
 
+  // Since RegisterCoalescer works well when we have both early clobber and
+  // explicit tied register, but not with implicit tied register, remove the
+  // early clobber flag so RegisterCoalescer can remove unneeded copies.
+  // This is safe to do because the regalloc will allocate the same register
+  // for the output and input register operand, so if there is a register in
+  // code operand, it won't be overwritten with the output register and we can
+  // do proper expansion PostRA.
+  if ((MI.getOpcode() == EraVM::SELcrr || MI.getOpcode() == EraVM::SELrcr) &&
```
akiramenai, line 106: Nit: we might have not to put it for `SELcrr`, `SELrcr` in the first place, haven't we?

akiramenai (approved):

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
vladimirradosavljevic: Do you want me to remove opcode checking? Even though we can remove this check completely, I left it on purpose so readers can understand when and why we are doing this.

hedgar2017 (approved):

llvm/lib/Target/EraVM/EraVMTieSelectOperands.cpp
akiramenai: I mean instead of placing clobbers on all selects and then conditionally remove them, we can conditionally place them. Right?

vladimirradosavljevic: We are placing early clobber flag only in case of select where register is used in stack/code operand in `EraVMBytesToCells`, and this is because of the functionality. Here, we are removing it because of for the reason described in the comment. If we want to remove this, we would need to run `LiveVariables` pass before `EraVMBytesToCells`, and skip adding early clobber flag if we are going to tie register operand for `SELcrr` and `SELrcr`. This will just add unnecessary complexity to the `EraVMBytesToCells` code, and I would like to avoid that.

akiramenai: Got it. Thanks

PR #539, opened 2024-05-14
[EraVM][TableGen] Define .static.shard versions of far calls

atrosinenko:
In preparation for renaming of assembler mnemonics, define the variant of far calls with both .static and .shard modifiers. Refactor AsmString computation in IFarCall along the way to simplify the migration to the new assembler syntax.


atrosinenko:
Please note that the only strictly necessary change is addition of `def _STATIC_SHARDrrl` to `FarCallInst` multiclass. On the other hand, `IFarCall` is updated to make switching to the new syntax as simple as changing string literals, as `.{st,sh}` modifier syntax cannot be expressed as concatenation of optional suffixes unlike `.static.shard`.

asl (approved):

akiramenai (approved):

PR #540, opened 2024-05-14
[EraVM][TableGen] Split regular and .inc variants of heap load/store

atrosinenko:
In preparation for renaming of assembler mnemonics, split `EraVMOpcode`s corresponding to heap load/store instructions into regular and incrementing variants, so their mnemonics can be set independently.


asl (approved):
Thanks!

akiramenai (approved):

PR #545, opened 2024-05-16
[EraVM][AsmParser] Support .{st,sh} modifier of far calls

atrosinenko:
Commit d5eb4e39ae8f8886abbfdca1caf04676cc9f8e3b introduced far call instructions with `.static.shard` modifier as a placeholder to simplify later migration to the new assembler syntax.

As these two-modifier forms are not yet used by the backend and braces require special handling by the parser, rename this specific compound modifier early, while keeping other names unchanged.


atrosinenko:
This is a follow-up for #539 as far calls turned out to be a bit harder w.r.t. "just renaming" the mnemonics and modifiers: the `.{st,sh}` modifier spans multiple input tokens: `Identifier("far_call.")`, `LCurly`, `Identifier("st")`, etc. Additionally, IIUC `{` and `}` require escaping in TableGen descriptions as unescaped braces denote assembler syntax variants (such as `intel` vs. `at-t` on X86).

sayon (approved):
Thank you, it is good for me.

PR #546, opened 2024-05-16
[EraVM] Refactor enum definitions in EraVM.h (NFC)

atrosinenko:
Remove `EraVMCTX::Context` enum as it is not used anymore.

TableGen-erate `EraVMAS::AddressSpaces` enum to prevent accidentally getting C++ and TableGen definitions out of sync.


asl (approved):

akiramenai:
Thank you for refactoring that!

Issue #441, opened 2024-03-18
[MC][AsmParser] Rename mnemonics according to the specification


atrosinenko:
At now, most instructions are supported with _some_ mnemonic, but these mnemonics can differ in the public specification, pre-existing code (LLVM backend and standalone assembler) and possibly the updated spec.

This issue covers instruction mnemonics in a broad sense (for example, the `<op>` part in `context.<op>`) - those cases that can be resolved by simply changing string literals in C++ code or `*.td` files without changing any logic.

See #440 for conditional codes.

atrosinenko:
As explained by @hedgar2017, replaced `gas` with `ergs` in asm mnemonics (not touching the intrinsic names yet): 3fc096329b94a41760333347e39d65325bbd7a21

asl:
We will wait for the updated ISA spec to perform the rename

asl:
@akiramenai @hedgar2017 @sayon So, what is the ultimate resolution on the assembler syntax from 1.6 ISA spec?

For example https://matter-labs.github.io/eravm-spec/spec.html#NearRetDefinition says:
```
Syntax
  ret in1
Legacy syntax
  ret.ok aliased as ret
```

How do you want to handle the legacy:
 1. Make backend produce only new variant, but let assembler accept the old one as an alias (and silently "upgrade"). So, it will be just `ret` and `ret.ok` will be treated as a mnemonic alias. Disassembler will produce new mnemonic.
 2. Make both backend and assembler support only new variant. Old one (`ret.ok`) will be rejected everywhere.

hedgar2017:
@asl #2, let's reject all legacy.

atrosinenko:
What is the new syntax for `.first` modifier of [`log`](https://matter-labs.github.io/eravm-spec/spec.html#EventDefinition) and [`logl1`](https://matter-labs.github.io/eravm-spec/spec.html#ToL1Definition) instructions?

asl:
@sayon Please check the question above

sayon:
Let's do `.f` for first. I will update it in spec tonight. CC @hedgar2017

hedgar2017:
Hey @sayon @atrosinenko, do we have a single place listing all mnemonics to sort them all out?

sayon:
Yes, in the isa redesign proposal, but i can make a separate summary table in spec for convenience -- been thinking about it.

atrosinenko:
[...]

PR #548, opened 2024-05-18
[EraVM][AsmParser] ret, revert, panic: cleanup legacy

atrosinenko:
Remove unused tablegen definitions and excessive assembler instruction aliases for ret, revert and panic. At now, just make the affected instructions accessible via a *single* mnemonic name (this does not change the assembler output produced by the backend), but postpone the renaming of the "main" mnemonics according to the new asm syntax until later.


atrosinenko:
Related discussion: #441.

asl (approved):

akiramenai (approved):

PR #549, opened 2024-05-18
[EraVM][MC] Define jump variants with ret_addr output operand

atrosinenko:


asl (approved):

atrosinenko:
Note on `Uses = [Flags]` and todo: if I understood it right, predicate operands can be used either "automatically" by if-conversion or manually. `EraVMInstrInfo::PredicateInstruction` already handles the automatic case, so I dropped `Uses = [Flags]` from the instruction definitions, unless non-zero `$cc` can be introduced by tablegen-erated instruction selector.

atrosinenko:
As far as I can see, the standalone assembler does not know how to encode return address operands.

@sayon Could you please look at the tests?

sayon (approved):
LGTM. Thank you!

FYI (in the future -- probably VM v1.6) jump will support any output operands (regs and stack).

akiramenai:
> As far as I can see, the standalone assembler does not know how to encode return address operands.

Standalone assembler is not actively supported for a while (that's why we are doing the MC work), and like everything else... codegen doesn't produce such jumps therefore nobody managed to support them throughout the toolchain.

PR #551, opened 2024-05-20
fixup! [EraVM][AsmParser] Parse SP-modifying operands

akiramenai:


atrosinenko:
LGTM, assuming `CPR-1360` was a task on supporting `stack-=` and `stack+=` to be used by the instructions with `in_any`/`out_any` operands.

Strictly speaking, at now it is NOP that may fall behind regular arithmetic instructions: for regular instructions `stack-=` and `stack+=` should be accepted by any instruction accepting plain `stack` for the same operand (this is covered by the tests), but only a subset of 6x4 variants is defined for `NOP*` (namely, `NOPrrr`, `NOPsrr`, `NOPrrs` and `NOPsrs`) - but this should be covered by the issues on migration to the new asm syntax, I think.

atrosinenko (approved):

PR #554, opened 2024-05-21
ci: fix for forks repositories and build binaries workflow

antonbaliasnikov:
# Code Review Checklist

## Purpose

* [x] Fixes issues with testing fork repositories
* [x] Introduces LLVM build binaries workflow for two purposes: regen ccache monthly and allow manual building of LLVM for all platforms 

## Testing

Special regression tests job from the fork is tested [here](https://github.com/matter-labs/era-compiler-llvm/actions/runs/9365344548). All commits were created and tested as expected.

## Ticket Number

* Fixes [CPR-1728](https://linear.app/matterlabs/issue/CPR-1728/add-a-workflow-building-llvm-binaries-for-all-platforms)
* Related to [CPR-1706](https://linear.app/matterlabs/issue/CPR-1706/issue-with-checkout-fork-repositories)
* Related to [CPR-1717](https://linear.app/matterlabs/issue/CPR-1717/turn-off-the-on-merge-trigger-in-ci)

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -1,0 +1,100 @@
+name: Build LLVM binaries
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # every month to regenerate ccache
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Git REF to use for the build"
+        required: false
+        type: string
+      build_macos_amd64:
+        description: "Build for MacOS amd64?"
+        required: false
+        type: boolean
+        default: true
+      build_macos_arm64:
+        description: "Build for MacOS arm64?"
+        required: false
+        type: boolean
+        default: true
+      build_linux_amd64:
+        description: "Build for Linux amd64?"
+        required: false
+        type: boolean
+        default: true
+      build_linux_arm64:
+        description: "Build for Linux arm64?"
+        required: false
+        type: boolean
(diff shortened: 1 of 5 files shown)


hedgar2017 (approved):

.github/workflows/build-binaries.yml
```diff
@@ -0,0 +1,100 @@
+name: Build LLVM binaries
+
+on:
+  schedule:
+    - cron: '0 0 * * *' # every month to regenerate ccache
```
hedgar2017, line 5: @antonbaliasnikov just curious, why do we need this?

hedgar2017 (approved):

.github/workflows/build-binaries.yml
antonbaliasnikov: This is to properly regenerate `ccache` that can be used across LLVM builds in PRs.

For front-end compilers, it's easy to understand when a new cache should be regenerated - each time an LLVM version in the `LLVM.lock` file changes, but for LLVM backend ideally we need to regenerate the cache each time a new commit is pushed to `main`, but it's too expensive - we'll be out of GitHub cache space pretty soon. So, here I'm introducing a monthly trigger that will update the `main` cache each month or by request.

Each month is good to keep the cache in almost 98-99% hits and at the same time not overload GitHub cache.

PR #557, opened 2024-05-21
[EraVM] Implement __system_request_slice_fallback

vladimirradosavljevic:
This implementation is similar to __system_request, but without the initial 4 bytes for the signature, and returns a whole slice.


hedgar2017 (approved):

vladimirradosavljevic:
Since we disabled CSE in GVN and NewGVN, running EraVMCSE later in the pipeline should fix these degradations.

hedgar2017 (changes requested):
Oh lawd, sorry, I was slightly wrong about this `@__system_request_fallback`!

It must return the pointer it receives from the underlying call, that is `system_request_result_abi_data`, since the call returns not a single value, but the whole slice!

Let's also rename it to `@__system_request_slice_fallback`, as we may have all 4 combinations in the future.

vladimirradosavljevic:
> Oh lawd, sorry, I was slightly wrong about this `@__system_request_fallback`!
> 
> It must return the pointer it receives from the underlying call, that is `system_request_result_abi_data`, since the call returns not a single value, but the whole slice!
> 
> Let's also rename it to `@__system_request_slice_fallback`, as we may have all 4 combinations in the future.

Done. Changed the name to `@__system_request_slice_fallback` and this function returns whole slice (`system_request_result_abi_data`).

hedgar2017 (approved):
Thanks!
Please merge.

akiramenai (approved):

akiramenai:
Only changed comments, so merging.

PR #558, opened 2024-05-21
[EraVM] Do not set Uses=[Flags] unconditionally in JCl

atrosinenko:


atrosinenko:
Hm, I tried to not update everything in one PR but then I spotted that it may be better to update `J` instruction here, too - otherwise it may handle `JCl` with `COND_NONE` case by itself.

PR #563, opened 2024-05-22
[EraVM][AsmParser] Cleanup outdated mnemonic aliases

atrosinenko:
Remove `MnemonicAlias`es as they are not needed in the new syntax and are only used on asm parsing.


llvm/lib/Target/EraVM/AsmParser/EraVMAsmParser.cpp
```diff
@@ -289,9 +289,6 @@ bool EraVMAsmParser::MatchAndEmitInstruction(SMLoc Loc, unsigned &Opcode,
 // Auto-generated by TableGen
 static unsigned MatchRegisterName(StringRef Name);
 static unsigned MatchRegisterAltName(StringRef Name);
-static void applyMnemonicAliases(StringRef &Mnemonic,
```
asl, line 292: Maybe keep the declaration and call? So if we'd add some aliases in the future they will apply automatically?

atrosinenko: Unfortunately, it seems `applyMnemonicAliases` is only generated if at least one alias exists. Maybe defining a dummy weak function would help, but I doubt it works in the same translation unit.

asl: Ah, ok, makes sense. Ok, then!

sayon (approved):
LGTM, thank you!

PR #566, opened 2024-05-23
[EraVM] Don't fold add that sets flags to select

vladimirradosavljevic:
Don't fold add that sets flags to select in EraVMOptimizeSelectPreRA pass.


akiramenai (approved):
Please either convert it to a fixup or add
"as `add.cc` can't replicate `add!` logic in case `.cc` isn't met" or similar clarification to the commit message.

vladimirradosavljevic:
> Please either convert it to a fixup or add "as `add.cc` can't replicate `add!` logic in case `.cc` isn't met" or similar clarification to the commit message.

Converted this to a fixup. Thanks for the review!

PR #567, opened 2024-05-23
[EraVM][MC] Refine the definitions of ret-like instructions

atrosinenko:
Make R1 register be understood as the default operand of `RETrl` and
`REVERTrl` instructions (but make this alias non-default for printing
until switched to the new asm syntax).

Change panic instruction to either accept jump target operand
or none at all.


atrosinenko:
@sayon @akiramenai Please check that this patch models the spec correctly.

atrosinenko:
Undone changes a bit to merge as much as possible before renaming the mnemonics:
* always print return- and revert-to-label with a hardcoded `r1` before label (not a real operand as of this PR and not accepted **on input**)
* use just `panic` for no-operand panic instruction (instead of `ret.panic`)

akiramenai (approved):
The implementation LGTM, thanks.
@sayon could you double check the constraint on ret-like input with the spec?

sayon:
The constraints LGTM.

Question: for the disassembler, will the non-sensical variants like `ret r9, somelabel` be displayed? It is encode-able, but in any case either the register or the label will be ignored. Far return ignores labels, near returns ignore register. 
Therefore in spirit of our decisions for the other instructions it would make sense to display it in a disassembler (for diagnostics) but reject it in assembler.

atrosinenko:
@sayon Instructions like `ret r9, somelabel` are supported in the `main` branch now and this PR is mostly about removing support for them, as described in the current spec (aside from that, this PR removes register operand from PANIC altogether). Is the original encoding (`reg` vs `reg, label`) better?

On the other hand, maybe defining all 24 variants of NOP was too permissive - but it is encoded very similar to other arithmetic instructions (except for not having the `set_flags` modifier) and has no side effects except for updating SP anyway.

Considering the nonsensical-ness of `ret r9, somelabel`: are there use cases where it is not known at compile-time if it is a near or far return? The existing spec on assembler syntax seems to suggest any instruction is **statically** either near or far ret/revert/panic.

atrosinenko:
The question seems to boil down to the following: what happens if an instruction with the [opcode](https://matter-labs.github.io/eravm-spec/spec.html#encode_opcode) set to 1070 (`OpRet` with `to_label` being true) ends up being executed as far return at run-time:
* if it is rejected by the VM, then it may be worth rejecting it in the assembler for clarity
* if it is permitted _by design_, then it may be worth allowing it both in the assembler and the spec, but I'm not sure

sayon:
It is statically impossible to distinguish between emitted far returns and near returns. 
We know which one we want during codegen.
From VM's perspective, there is a single type of return instruction and it decides dynamically whether it is far or near return, depending on the topmost stack frame. 

A third party may forge an instruction like `ret r5, somelabel` which will behave differently based on the type of the topmost callframe. It should not be a problem for security or anything, but it would be nice to distinguish it in disassembler.

[Here](https://matter-labs.github.io/eravm-spec/spec.html#NearRetDefinition) we write that the semantic is decided in runtime; please tell me if it is not clear and I will try to rework/reformulate/make it more visible.

akiramenai:
+1 to @sayon. It sounds like it makes sense to emit an error if to assemble `ret r9, somelabel`, but disassemble it if it's really in the code.
I don't know though if it's a problem that we can't generally assemble what was disassebled.

sayon:
If this is a problem, we may allow to assemble `ret r9, somelabel`, but emit a warning. Later we may introduce some CLI option like "allow-extended-instruction-set" to be able to emit such instructions (same with e.g. `nop stack-=[...], r0, stack+=[...], r9`), but normally reject them.

atrosinenko:
Updated the patch and the description. Now instructions like `ret r9, @somelabel` are supported again, but the aliases are added for the canonical case of register operand being `r1`. Register operands are still rejected for any variants of `panic` instruction.

Until switching to the new syntax, the aliases for "`r1` by default" are disabled when printing (for compatibility with the standalone assembler), so the test has the following seemingly meaningless lines:
```
; CHECK:  ret.ok.to_label       r1, @label
...
; CHECK:  ret.ok.to_label       r3, @label
```
After switching to the new syntax and removing `let EmitPriority = 0 in { ... }` they will be something like
```
; CHECK:  retl       @label
...
; CHECK:  retl       r3, @label
```

akiramenai (approved):

PR #569, opened 2024-05-23
[EraVM][TableGen] Split `EraVMOpcode`s for ret-like instructions (NFC)

atrosinenko:
In preparation for renaming of assembler mnemonics, split `EraVMOpcode` definitions for ret, revert and panic instructions into "without label" and "with label" variants, so their mnemonics can be specified independently.


atrosinenko:
Follow-up for #567, merely a refactoring of tablegen definitions.

asl (approved):

PR #571, opened 2024-05-24
[EraVM][MC] Add missing variants of NOP instruction

atrosinenko:
Make NOP support the remaining combinations of operands.

Reorganize the tests: add `nop.(s|txt)` test files similar to those for arithmetic instructions, put the test cases for `nop`, `incsp` and `decsp` aliases to `sp-changes.(s|txt)` files.

PR #575, opened 2024-05-28
[EraVM] Make `@fat.ptr.call` test case check for ptr.add instruction

atrosinenko:
Clarify the `@fat.ptr.call` test case before renaming the mnemonics.

The instructions actually generated for `@fat.ptr.call` function are

    ptr.add    r2, r0, r1
    near_call  r0, @fat.ptr.arg, @DEFAULT_UNWIND
    ret

Before renaming `ptr.add` to `addp`, the first line was accidentally matched by the `add r2, r0, r1` substring.

### DIFF
--- a/llvm/test/CodeGen/EraVM/calling-convensions.ll
+++ b/llvm/test/CodeGen/EraVM/calling-convensions.ll
@@ -435,7 +435,7 @@ define void @fat.ptr.arg(i256 addrspace(3)* %ptr) {
 
 ; CHECK-LABEL: fat.ptr.call
 define void @fat.ptr.call(i256 %a1, i256 addrspace(3)* %ptr) {
-  ; CHECK: add r2, r0, r1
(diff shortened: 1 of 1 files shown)


akiramenai:
@atrosinenko do we want to merge it to `eravm-mc/reorganize-nops` or to `main`?

atrosinenko:
> do we want to merge it to eravm-mc/reorganize-nops or to main?

To `main`. This PR is stacked on top of #571, so `eravm-mc/reorganize-nops` is specified as the base branch, but #571 is not merged yet, so this PR should not either. This PR does not really depend on #571, it is only set up this way so both are the dependencies of #576.

akiramenai (approved):

atrosinenko:
The only CI failure is an obscure `git` error before running `clang-tidy`. Merging, as everything else including Lit tests passed.

PR #576, opened 2024-05-28
[EraVM][AsmParser] Migrate to new asm syntax

atrosinenko:
Resolves #441.

PR #578, opened 2024-05-28
[EraVM][TableGen] Cleanup the description of ret/revert/panic

atrosinenko:
Rename RETURN and REVERT pseudos to make their purpose easier to understand.

Remove InstAlias for panic instruction. Since register operand of panic instruction was removed, it virtually became a MnemonicAlias, so update the definition of OpPanic instead as no alias needed in the new syntax.


asl (approved):

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -1122,7 +1122,6 @@ def THROW : Pseudo<(outs), (ins GR256:$rs0, pred:$cc), [(EraVMthrow GR256:$rs0)]
 
 def : InstAlias<"ret${cc}", (RETr R1, pred:$cc)>;
 def : InstAlias<"revert${cc}", (REVERTr R1, pred:$cc)>;
-def : InstAlias<"panic${cc}", (PANIC pred:$cc)>;
```
atrosinenko, line 1125: I would rather not merge this until #576 to not introduce conflicts.

PR #582, opened 2024-05-29
[EraVM][AsmParser] Change `.{st,sh}` far call modifier to `.st.sh`

atrosinenko:
Simplify the syntax for specifying both "static" and "shard" modifiers of a far call: either `mnemonic.st.sh` or `mnemonic.sh.st`.

Please note that ".st" and ".sh" are technically just parts of the mnemonic's name, so they should go before a predicate, if any.


llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -697,7 +697,7 @@ class IFarCall<EraVMOpcode opcode,
       !cond(!eq({is_static, is_shard}, {0, 0}): "",
             !eq({is_static, is_shard}, {1, 0}): ".static",
```
hedgar2017, line 698: Does it mean that these full notations are also accepted?

llvm/test/MC/EraVM/asm-parser/far-calls.s
```diff
@@ -25,12 +25,21 @@ foo:
   far_call.delegate.shard.lt    r3, r4, @handler
   far_call.mimic.shard.lt       r3, r4, @handler
 
-  far_call.{st,sh}           r3, r4, @handler
-  far_call.delegate.{st,sh}       r3, r4, @handler
-  far_call.mimic.{st,sh}          r3, r4, @handler
-  far_call.{st,sh}.lt        r3, r4, @handler
-  far_call.delegate.{st,sh}.lt    r3, r4, @handler
-  far_call.mimic.{st,sh}.lt       r3, r4, @handler
+  far_call.st.sh           r3, r4, @handler
```
hedgar2017, line 28: And for mnemonics, do we still accept the old notation?

hedgar2017 (changes requested):

llvm/test/MC/EraVM/asm-parser/far-calls.s
atrosinenko: What is the old notation? Support for `.{st,sh}` notation is explicitly removed by this PR (replaced by `.st.sh` or `.sh.st`).

Considering the specific names used for mnemonics and modifiers, this PR uses the old mnemonic names (such as `far_call` instead of `callf`). It uses old modifier names for pre-existing modifiers as well (`.static` and `.shard`) to prevent CI failures. The combined `.st.sh` form is different as it did not exist before #539, that is why I use the new syntax even before renaming everything else.

llvm/lib/Target/EraVM/EraVMInstrFormats.td
atrosinenko: Strictly speaking, now the syntax is not fully consistent: everything uses the old names except for the combined `.st.sh` modifier (it did not exist before #539). Thus, exactly the following forms are accepted (`far_call` can be replaced with `far_call.delegate` or `far_call.mimic`):
```
  far_call        r1, r2, @foo
  far_call.static r1, r2, @foo
  far_call.shard  r1, r2, @foo
  far_call.st.sh  r1, r2, @foo
  far_call.sh.st  r1, r2, @foo ; supported via MnemonicAlias
; the same with predicate, such as far_call.shard.lt r1, r2, @foo
```
Note that #539 replaced conditional concatenation of `.static` and then `.shard` with 2x2 cartesian product because `.{st,sh}` was expected in the new syntax that cannot be represented that way.

I can simplify this back to something like
```
let AsmString = !strconcat(opcode.Name,
                           !if(is_shard, ".shard", ""),
                           !if(is_static, ".static", ""),
                           "${cc}", "\t", asmstr);
```
In this case the fourth variant (`.st.sh`) would be `.static.shard` and will be renamed in #576. Or we can postpone this simplification until #576 is merged.

hedgar2017: Understood.
Sounds like the legacy is piling up, so we need to migrate to the new syntax completely.

hedgar2017: Let's address it in another PR.
Here LGTM!

hedgar2017 (approved):

llvm/lib/Target/EraVM/EraVMInstrFormats.td
atrosinenko: Do you mean replacing `.static` and `.shard` with `.st` and `.sh`? This could break CI if these modifiers are actually emitted for system contracts, for example. My idea was to either rollback `.st.sh` to `.static.shard` temporarily (and simplify `AsmString` computation now) or do not clutter this PR and #576 by renaming modifiers back-and-forth and postpone simplification of `AsmString`.

PR #584, opened 2024-05-30
[EraVM] Remove NOPSP and NOPSPr pseudos

atrosinenko:
Update `(EraVMchange_sp GR256:$sp_change)` pattern to select NOPrrs instead of NOPSPr (used by `llvm.restorestack` intrinsic).

Remove `(EraVMchange_sp imm16:$sp_change)` pattern as it is currently unused and untested.

Change the `EraVMchange_sp` name to more descriptive `EraVMadd_to_sp`.


atrosinenko:
Something like the following might work for `(EraVMchange_sp $imm)` case (at least, it is understood by tablegen...), but I would rather not add unused (and thus untested) obscure patterns.

```
def trunc_to_i16 : SDNodeXForm<imm, [{
  return CurDAG->getTargetConstant(N->getAPIntValue(), SDLoc(N), MVT::i16);
}]>;

def : Pat<(EraVMchange_sp imm16:$sp_inc),
          (NOPrrs R0, R0, R0, R0, (trunc_to_i16 imm:$sp_inc), 0)>;
def : Pat<(EraVMchange_sp neg_imm16:$sp_dec_neg),
          (NOPsrr R0, R0, (negate_imm imm:$sp_dec_neg), R0, 0)>;
```

asl (approved):

PR #593, opened 2024-06-03
[EraVM][AsmPrinter] Reuse EraVM::analyzeMCOperandsCode/Stack functions.

atrosinenko:
Reuse the same `parseMCOperandsCode` and `parseMCOperandsStack` functions that are used by code emitter to analyze operands of MCInst.

While it makes easier to improve formatting of printed assembly code, this commit strives to keep the existing formatting as much as possible, to not combine refactoring with massive changes of tests.


llvm/test/CodeGen/EraVM/array.ll
```diff
@@ -19,15 +19,15 @@ define i256 @consti_loadconst_neg_offset(i256 %idx) nounwind {
 
 ; CHECK-LABEL: consti_loadconst_storeglobal
 define void @consti_loadconst_storeglobal() nounwind {
-  ; CHECK: add @const[1], r0, stack[@val+1]
+  ; CHECK: add @const[r0+1], r0, stack[@val + 1]
```
asl: Is this desired? Can we skip `r0` here for the sake of clarity?

atrosinenko: The idea is to simplify asm printing first (without changing too many tests, so reviewing test changes is manageable). Then, it should be easier to improve the formatting via mostly obvious changes in `EraVMInstPrinter.cpp` (but there will be lots of changes in tests).

This PR strives to collapse the original operand printing code into a call to `analyzeMCOperandsCode` or `analyzeMCOperandsStack` and a _mostly_ straightforward formatting.

Printing `r0` here is not desired, of course. At first glance it looked like a good balance between temporarily overcomplicating printer and changing many existing tests. Now I experimented a bit more and it seems things can be further simplified - I will update this PR tomorrow.

atrosinenko: Fixed

atrosinenko:
Updated the patch. Added a few test cases: check that something meaningful is printed for `[r0+0]` "subscripts".

asl (approved):

PR #601, opened 2024-06-06
[EraVM] Refactor mapping between different input operand kinds

### LINKED ISSUE #482: I introduced that duplication because I had issues with `OperandAddrMode` as the ColField. As you refactor our `td`s anyway, may I ask if it's hard to get rid of `OperandAM`?
I introduced that duplication because I had issues with `OperandAddrMode` as the ColField. As you refactor our `td`s anyway, may I ask if it's hard to get rid of `OperandAM`?

_Originally posted by @akiramenai in https://github.com/matter-labs/era-compiler-llvm/pull/448#discussion_r1553744353_

atrosinenko:
* Collapse OperandAddrMode and OperandAM fields of IBinary class
* Use named constants instead of numbers
* Drop synthetic Value field from OperandAddrModeValue class (renamed to SrcOperandMode)

### DIFF
--- a/llvm/lib/Target/EraVM/EraVMInstrFormats.td
+++ b/llvm/lib/Target/EraVM/EraVMInstrFormats.td
@@ -17,14 +17,12 @@
 // The addressing mode can be categorized into operand addressing and destination addressing
 // OperandAddrMode:
 // RR : Reg Reg; SR : Stack Reg; CR : Code Reg; IR : Imm Reg;
-class OperandAddrModeValue<bits<3> val> {
-  bits<3> Value = val;
-}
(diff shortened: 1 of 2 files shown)


akiramenai (approved):
LGTM,
But I'm not sure if we should apply TableGen tag, as it only changes td files inside EraVM. However, I don't know what is the common practice on the matter, so I leave the decision to you.

atrosinenko:
@akiramenai I glanced through the commit history and it looks like using `[TableGen]` tag to refer to "users" of TableGen is not as widespread as I expected. Updated the commit messages here and in #602, thanks.

PR #602, opened 2024-06-06
[EraVM] Refactor the three other `InstrMapping`s

### LINKED ISSUE #482: I introduced that duplication because I had issues with `OperandAddrMode` as the ColField. As you refactor our `td`s anyway, may I ask if it's hard to get rid of `OperandAM`?

atrosinenko:
Unlike mapping between different input operand kinds, the three other `InstrMapping`s only have two options each that are flipped by the mapper function.

Remove the synthetic Value field from `DestAddrModeValue` (renamed to `DestOperandMode`) class. Keep Value fields in `mod_set_flags` and `mod_swap` as these fields are meaningful for opcode computation.

Remove the `ToReg` vs. `ToRegReg` and `ToStack` vs. `ToStackReg` separation as it is not really used.

### DIFF
--- a/llvm/lib/Target/EraVM/EraVMInstrFormats.td
+++ b/llvm/lib/Target/EraVM/EraVMInstrFormats.td
@@ -25,14 +25,10 @@ def OpndSR         : SrcOperandMode;
 def OpndCR         : SrcOperandMode;
 
 // DestAddrMode:
-class DestAddrModeValue<bits<3> val> {
-  bits<3> Value = val;
-}
-def DestAddrNotSet : DestAddrModeValue<0>;
-def ToReg          : DestAddrModeValue<1>;
-def ToStack        : DestAddrModeValue<2>;
-def ToRegReg       : DestAddrModeValue<3>;
-def ToStackReg     : DestAddrModeValue<4>;
+class DestOperandMode;
+def DestAddrNotSet : DestOperandMode;
+def ToReg          : DestOperandMode;
+def ToStack        : DestOperandMode;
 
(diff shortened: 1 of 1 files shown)


akiramenai (approved):
LGTM,
But I'm not sure if we should apply TableGen tag, as it only changes td files inside EraVM. However, I don't know what is the common practice on the matter, so I leave the decision to you.

asl:
Looks like benchmarks changed... @sayon can we get results dump?

atrosinenko:
@asl Specifically, looks like the results **were updated** by the bot - probably after I merged #601 without waiting for the CI once again after changing commit message only.

asl:
@atrosinenko Ah, oh. Can you rebase then?

atrosinenko:
> Ah, oh. Can you rebase then?

@asl Strictly speaking, it is already rebased onto current HEAD of `main` branch :) Moreover, the benchmark results in #601 were updated the same way (and have seemingly the same delta values). Maybe restarting some CI job would help, but I'm not sure what to restart.

asl:
@antonbaliasnikov @akiramenai @sayon Can you help?

atrosinenko:
Finally, after pushing more commits to `main` and rebasing this PR, the benchmark results were recomputed again and now they are as expected.

Issue #590, opened 2024-06-03
Redundant ADD after folding overflow


TerryGuo:
In order to fix problem in #521 , we will end up with two Nodes: one is to do the arithmetic operations (add/sub/mul) and another is to set the flags. If the result of arithmetic operation isn't used, then the arithmetic node will be optimized out, this is expected. If there is contradiction between glue and data dependency, then we do need those two nodes, this is expected as well. However when the result of arithmetic operation is needed and there is **NO** contradiction, we still get two nodes, the second one becomes redundant now. An example is as below, for input IR:
```
define i256 @add_test_neg_1(i256 %a, i256 %b, i256 %x, i256 %y) {
entry:
  %res1 = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 %y)
  %res2 = extractvalue {i256, i1} %res1, 0
  %overflow = extractvalue {i256, i1} %res1, 1
  %selected = select i1 %overflow, i256 %a, i256 %b
  %sum = add i256 %res2, %selected
  ret i256 %sum
}
``` 
The generated asm code with #521 will be:
```
     1          .text
     2          .file   "overflow-neg.ll"
     3          .globl  add_test_neg_1                  ; -- Begin function add_test_neg_1
     4  add_test_neg_1:                         ; @add_test_neg_1
     5  ; %bb.0:                                ; %entry
     6          add!    r3, r4, r5
     7          add.lt  r1, r0, r2
     8          add     r3, r4, r1
     9          add     r1, r2, r1
    10          ret
``` 
The `add` in `line 8` is redundant and can reuse `r5` from `line 6`.

TerryGuo:
Same problem happens to #581 as well, for input IR like below:
```
define void @add_branch_neg_1(i256 %x, i256 %y) {
entry:
  %res1 = call {i256, i1} @llvm.uadd.with.overflow.i256(i256 %x, i256 %y)
  %sum = extractvalue {i256, i1} %res1, 0
  %overflow = extractvalue {i256, i1} %res1, 1
  br i1 %overflow, label %overflow_detected, label %no_overflow_detected

overflow_detected:
  call void @has_overflow(i256 %sum)
  br label %exit

no_overflow_detected:
  call void @has_no_overflow(i256 %sum)
  br label %exit

exit:
  ret void
}
``` 

We will get below asm:
```
    13  add_branch_neg_1:                       ; @add_branch_neg_1
[...]

PR #612, opened 2024-06-10
[EraVM] Enable CSE between sub and cmp

vladimirradosavljevic:


vladimirradosavljevic:
This should resolve CPR-1404.

llvm/test/CodeGen/EraVM/rol.ll
```diff
@@ -1,5 +1,5 @@
 ; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py
-; RUN: llc --disable-eravm-scalar-opt-passes < %s | FileCheck %s
+; RUN: llc --disable-eravm-scalar-opt-passes -enable-eravm-combine-addressing-mode=false < %s | FileCheck %s
```
akiramenai, line 2: Could you add explanation why to disable combine addressing mode in `dbf11311f6fd6f5f2b31bda35d0d9b3cc53918a1` message?

llvm/lib/Target/EraVM/EraVMInstrInfo.td
```diff
@@ -549,15 +549,15 @@ multiclass ArithINonCommutable<EraVMOpcode opcode, SDPatternOperator node> : Ari
   }
 
   // non-commutative: irr_s -> crr_s
-  def : Pat<(node GR256:$rs1, large_imm:$imm),
+  def : Pat<(node large_imm:$imm, GR256:$rs1),
```
akiramenai, line 472: Commit message:
This patch fixes the order of input operands in crr_s and crs_s patterns. Thus it enables crr_s, crs_s forms for non-commutative instructions to be selected saving on materialisation of big integers.
Before the patch the incorrect patterns didn't cause bugs because yrr_s, yrs_s override them.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -71,16 +72,9 @@ class EraVMOptimizeSelectPreRA : public MachineFunctionPass {
...
-  /// register which we tie with the output register. In the end
-  /// copy flags and memory references from FlagMI.
-  MachineInstr *
-  getFoldedInst(MachineInstr &ToFold, const MachineInstr &FlagMI,
-                iterator_range<MachineInstr::const_mop_iterator> In0,
-                Register In1, unsigned NewOpcode, unsigned NewCC,
-                Register TieReg) const;
+  /// Return an out register of MI. In case MI has two outputs, we can
+  /// only fold instruction iff one output is used and the other is not.
+  Register getOutReg(const MachineInstr &MI) const;
```
akiramenai: Could you clarify the name and the documentation?
Most importantly, could we explain why 2 used outs is a show stopper? Is it a principal restriction or an implementation detail (TODO + issue)

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -98,9 +92,10 @@ class EraVMOptimizeSelectPreRA : public MachineFunctionPass {
   /// if z and y can be allocated to same reg. The tie is used to ensure this.
   bool tryFoldSelectZero(MachineBasicBlock &MBB);
 
-  /// In case ADD is fed into a SELECT, and one of the inputs of ADD matches
-  /// with the input of SELECT, then it is possible to fold ADD with SELECT
-  /// to result in just one folded instruction.
+  /// In case arithmetic or bitwise instruction is fed into a SELECT, and one of
```
akiramenai: an arithmetic or a bitwise.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -98,9 +92,10 @@ class EraVMOptimizeSelectPreRA : public MachineFunctionPass {
...
+  /// the inputs of these instructions matches with the input of SELECT, then it
+  /// is possible to fold them with SELECT to result in just one folded
```
akiramenai: them -> it

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -113,7 +108,7 @@ class EraVMOptimizeSelectPreRA : public MachineFunctionPass {
   ///
   /// if outSEL and x/y can be allocated to same reg.
   /// The tie is used to ensure this.
-  bool tryFoldAddToSelect(MachineBasicBlock &MBB);
+  bool tryFoldArithAndBitwiseToSelect(MachineBasicBlock &MBB);
```
akiramenai: nit: `tryFoldToSelect` is ok here.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -297,58 +301,93 @@ bool EraVMOptimizeSelectPreRA::tryFoldSelectZero(MachineBasicBlock &MBB) {
...
-    InAddRegs.insert(EraVM::in1Iterator(MI)->getReg());
+    SmallSet<Register, 2> InRegs;
+    InRegs.insert(EraVM::in1Iterator(MI)->getReg());
     if (EraVM::hasRRInAddressingMode(MI))
-      InAddRegs.insert(EraVM::in0Iterator(MI)->getReg());
+      InRegs.insert(EraVM::in0Iterator(MI)->getReg());
 
-    // In order to fold add to select, we expect that other input of a select
-    // instruction is matching with one of the add inputs.
+    // In order to do the folding, we expect that other input of a select
```
akiramenai: nit: in order to fold is ok.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -297,58 +301,93 @@ bool EraVMOptimizeSelectPreRA::tryFoldSelectZero(MachineBasicBlock &MBB) {
...
+    MachineInstrBuilder(MF, NewMI)
+        .addReg(TieReg, RegState::Implicit)
+        .addReg(EraVM::Flags, RegState::Implicit);
+
+    // Add tie to ensure those two operands will get same reg
+    // after RA pass. This is the key to make transformation in this pass
+    // correct.
+    NewMI->tieOperands(DefIdxToTie, NewMI->getNumOperands() - 2);
+
+    LLVM_DEBUG(dbgs() << "== Folding arith or bitwise:"; MI.dump();
```
akiramenai: nit: one can infer the class of instruction from MI.dump(), we don't need to duplicate.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
```diff
@@ -297,58 +301,93 @@ bool EraVMOptimizeSelectPreRA::tryFoldSelectZero(MachineBasicBlock &MBB) {
...
-    Register TieReg = OutAddIsIn1Use ? EraVM::in0Iterator(UseMI)->getReg()
-                                     : EraVM::in1Iterator(UseMI)->getReg();
-
-    [[maybe_unused]] auto *NewMI = getFoldedInst(
-        UseMI, MI, EraVM::in0Range(MI), EraVM::in1Iterator(MI)->getReg(),
-        MI.getOpcode(), CCNewMI, TieReg);
-    LLVM_DEBUG(dbgs() << "== Folding add:"; MI.dump();
-               dbgs() << "       and use:"; UseMI.dump();
-               dbgs() << "          into:"; NewMI->dump(););
+    // To fold the instruction into select, we need to update the original
```
akiramenai: I suggest to still do it in a separate function and put this documentation to its declaration.
I also suggest to add details (how we update CC, which registers we tie, and also explicitly mention that we add flags usage because we adding CC). An example is also of help here.

akiramenai (commented):
I haven't finished with the last patch yet. I need to think about the design because it looks very similar to the problem we have in #590. Perhaps it's better to teach CSE that flag setting counterparts have the same semantic bug flags.

vladimirradosavljevic:
> I haven't finished with the last patch yet. I need to think about the design because it looks very similar to the problem we have in https://github.com/matter-labs/era-compiler-llvm/issues/590. Perhaps it's better to teach CSE that flag setting counterparts have the same semantic bug flags.

I'm not sure how this is similar to #590, since in ISEL cmp and sub instructions are represented by different nodes + CSE is doing global transformation not local (in the BB scope). First step in CSE is to find identical instructions to eliminate, and to do so it is using lookup map. If we want to teach CSE that _s and _v are the same, we would always need to generate MI, to check the lookup table and to worry about deallocation of that MI (a lot of local changes + we would need to be very careful not to introduce some issues). After that, CSE is checking whether it can remove identical instructions and it is not the simple heuristic that is using.
Please note that I also thought about teaching CSE about this case, and then I saw this approach that AArch64 is using,  to generate fake cmp instructions from sub to leverage MachineCSE, and later on in the PeepholeOptimizer to convert these fake cmp into sub.

llvm/lib/Target/EraVM/EraVMInstrInfo.cpp
```diff
@@ -417,6 +417,71 @@ unsigned EraVMInstrInfo::insertBranch(
...
+  // In this case, `Out2 = SUB_v` that will be converted to _s in this function,
+  // prevents MachineCSE to remove `Out3 = SUB_v` in favor of `Out1 = SUB_v`,
+  // just because of `implicit-def dead $flags`, so try to find identical
+  // compare instruction and remove it here.
+  //
+  // Search backwards in order to find the nearest flag setting instruction that
+  // can be identical to this compare instruction.
+  auto From = std::next(MachineBasicBlock::reverse_iterator(CmpInstr));
+  for (MachineInstr &MI : make_range(From, CmpInstr.getParent()->rend())) {
+    if (isFlagSettingInstruction(MI)) {
```
akiramenai: Perhaps we should redefine `isFlagSettingInstruction` as
```
any_of(MI.implicit_operands(), [](const MachineOperand &MO) {
          return MO.isReg() && MO.isDef() && MO.getReg() == EraVM::Flags;
        }))
```
And then use `find_if` instead of loop.

Also from the logic it looks like that by the time we do this search, `Out2 = SUB_v ...` is already converted and the `impact-def` is dropped while the text about says that the instruction will be converted. So could you clarify on that?

akiramenai (commented):
> Please note that I also thought about teaching CSE about this case, and then I saw this approach that AArch64 is using, to generate fake cmp instructions from sub to leverage MachineCSE, and later on in the PeepholeOptimizer to convert these fake cmp into sub.

Could you kindly put all your findings relevant to the design solution used in this PR to the commit message? Also referencing this PR is very welcome too.

hedgar2017:
@vladimirradosavljevic @akiramenai could we finalize this?
Look like a good piece of code to include in the release.

llvm/test/CodeGen/EraVM/rol.ll
vladimirradosavljevic: Added explanation.

llvm/lib/Target/EraVM/EraVMInstrInfo.td
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Changed the name to be the same as in the `EraVMOptimizeSelectPostRA`. It is a principal restriction and added explanation for it.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMOptimizeSelectPreRA.cpp
vladimirradosavljevic: Moved this code to a separate function. Simplified code by using `find_if` to remove the loop, and reverted back `getFoldedInst` function.

llvm/lib/Target/EraVM/EraVMInstrInfo.cpp
vladimirradosavljevic: Nice catch. Added `find_if` to simplify the code and added clarification that we can now do the CSE since we dropped `impact-def` for `Out2 = SUB_v ...`.

vladimirradosavljevic:
> > Please note that I also thought about teaching CSE about this case, and then I saw this approach that AArch64 is using, to generate fake cmp instructions from sub to leverage MachineCSE, and later on in the PeepholeOptimizer to convert these fake cmp into sub.
> 
> Could you kindly put all your findings relevant to the design solution used in this PR to the commit message? Also referencing this PR is very welcome too.

Done. PTAL. Thanks for the review!

akiramenai (approved):

PR #615, opened 2024-06-11
[EraVM] Prepare for removing repeated operand definitions in *.td files

atrosinenko:
Perform a few preparations before removing repeated definitions of operands in `EraVMInstrInfo.td`:
* move the definitions of EraVM-specific `AsmOperandClass`es and `Operand`s, so that they can be used in `EraVMInstrFormats.td`
* define tablegen classes to be mixed into `Ixx_x` instruction classes

The main part of refactoring is #616.


llvm/lib/Target/EraVM/EraVMInstrFormats.td
```diff
@@ -346,21 +428,59 @@ class IBinaryS<EraVMOpcode opcode,
...
-class Irr_r<EraVMOpcode opcode,
-            mod_swap swap_operands, mod_set_flags silent,
-            dag outs, dag ins,
-            string asmstr,
-            list<dag> pattern>
-  : IBinaryR<opcode, SrcReg, swap_operands, silent, outs, ins, asmstr, pattern> {
+// Mix-in classes to hook standard fields for different combinations
+// of input operands.
+
+class SrcRRFields {
```
asl: I do not like `Fields` as naming. Maybe just `Operands`? or something like this? After all, the mixin is to indicate the particular instruction has said operands.

atrosinenko: Renamed `SrcXXFields` to `SrcOperandsXX` according to the intention instead of the implementation details, thanks.

asl (approved):

PR #616, opened 2024-06-11
[EraVM] Drop repeated instruction operands definitions from *.td files

atrosinenko:
Do not repeat in/out dags and operand strings in definition of each particular instruction. Most of the times, instruction class already defines fields such as `bits<4> rs0` and hooks them to encoded fields (such as `Src0`), and then dags and operand string mirror these definitions. In a few places definitions are custom enough (such as eight variants of jump instruction) - in that case a more flexible way to describe the operands is exposed.

Several `Ixx_x` classes are split into r- and p-versions of operands: both are register operands but the former is `GR256` register and the latter is `GRPTR`. For example, `ADDrrr_s` is an instance of `Irr_r` class and `PTR_ADDrrr_s` is an instance of `Ipr_p` class.

A number of shortcuts are defined for specifying an operand (except for the special `${cc}` predicate): `EraVMOperand` tablegen class incapsulates both operand string and in/out dags (actually, one of the two dags is always empty and the other defines exactly one operand). For example, the output register of `PTR_ADDrrr_s` is `rd0_ptr` defined as `EraVMOperand<"$rd0", (ins), (outs GRPTR:$rd0)>`.

### DIFF
--- a/llvm/lib/Target/EraVM/EraVMInstrFormats.td
+++ b/llvm/lib/Target/EraVM/EraVMInstrFormats.td
@@ -191,6 +191,44 @@ def stackout : Operand<i256> {
 }
 
 //===----------------------------------------------------------------------===//
+// EraVM Instruction Operands - helper definitions.
+//===----------------------------------------------------------------------===//
+
+// Description of EraVM instruction operand, except for ${cc}.
+class EraVMOperand<string str, dag ins, dag outs = (outs)> {
+  string AsmOps = str;
+  dag InOps = ins;
+  dag OutOps = outs;
+}
+
+def rs0 : EraVMOperand<"$rs0", (ins GR256:$rs0)>;
+def rs1 : EraVMOperand<"$rs1", (ins GR256:$rs1)>;
+def rd0 : EraVMOperand<"$rd0", (ins), (outs GR256:$rd0)>;
+def rd1 : EraVMOperand<"$rd1", (ins), (outs GR256:$rd1)>;
+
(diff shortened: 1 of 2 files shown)


atrosinenko:
To estimate the changes: flatten tablegen description:
```bash
cd /path/to/era-compiler-llvm/build
# checkout old version
./bin/llvm-tblgen \
    -I ../llvm/lib/Target/EraVM \
    -I ./include \
    -I ../llvm/include \
    -I ../llvm/lib/Target \
    ../llvm/lib/Target/EraVM/EraVM.td \
    -o /tmp/old.txt
# checkout new version
./bin/llvm-tblgen ... -o /tmp/new.txt
```
Drop several known changes and compare the remaining:
```bash
filter() {
  # 1) skip class definitions
  # 2) drop end-of-line comments
  # 3) ignore reordered fields like `bits<4> rs0;`
  sed --regexp-extended \
    -e '1,/--- Defs ---/ d' \
    -e '/^def/ s://.*$::' \
    -e 's/bits<[0-9]+> (rs|rd|src|imm)(0|1)? = \{ \?(, \?)+ \}/bits<N> FIELD = { ?, ...}/'
}

filter < /tmp/old.txt > /tmp/old-filtered.txt
# `string asmstr` is added for conveniance - ignore it
filter < /tmp/new.txt | sed '/string asmstr = / d' > /tmp/new-filtered.txt
```
Compared to `old-filtered.txt`, the `new-filtered.txt` only differs in the first operand of `NEAR_CALL` named `rs0` instead of `in1` and a bunch of instances of `EraVMOperand` being defined.

akiramenai (approved):

PR #626, opened 2024-06-18
[CGP] Fix matching of uadd overflow

vladimirradosavljevic:


llvm/test/CodeGen/EraVM/cgp-uadd-overflow-wrong-match.ll
```diff
@@ -0,0 +1,50 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
```
akiramenai: Perhaps it's better to stop after CGP and check for the intrinsic instead of running all codegen pipeline.

akiramenai (commented):
It should be upsteamable, shouldn't it? E.g. we don't need i256 target to test this code, a test can be put to generic codegen tests or something.

llvm/test/CodeGen/EraVM/cgp-uadd-overflow-wrong-match.ll
vladimirradosavljevic: Makes sense. Added test that runs only CGP and matched IR.

vladimirradosavljevic:
> It should be upsteamable, shouldn't it? E.g. we don't need i256 target to test this code, a test can be put to generic codegen tests or something.

Yes, this can be upstreamed. Unfortunately, there are only few generic CGP tests, others are in `llvm/test/Transforms/CodeGenPrepare/TargetName`, so we can add the same test for ARM target to show the problem and the fix.

vladimirradosavljevic:
Closing in favor of #630.

PR #628, opened 2024-06-18
[InstCombine] Fix miscompile on GEP+load to icmp fold

vladimirradosavljevic:


vladimirradosavljevic:
Closing in favor of #630.

PR #629, opened 2024-06-19
ci: enable check-lld in regression tests

antonbaliasnikov:
# Code Review Checklist

## Purpose

Enables missing `check-lld` tests by running `check-all` as part of `verify-llvm`, instead of `check-llvm` only.


PavelKopyl (approved):
Thank you!

PR #630, opened 2024-06-19
Cherry pick fixes for large signed constant truncation

vladimirradosavljevic:


vladimirradosavljevic:
Please note that in the last commit, just parts that fix large signed constant truncation of the upstream [PR](https://github.com/llvm/llvm-project/pull/80309) are taken.

vladimirradosavljevic:
Regressions tests are failing, because clang-tidy is complaining about the following:
```
/__w/era-compiler-llvm/era-compiler-llvm/llvm/llvm/lib/Transforms/Scalar/IndVarSimplify.cpp:362:3: error: use auto when initializing with new to avoid duplicating the type name [hicpp-use-auto,-warnings-as-errors]
  362 |   ICmpInst *NewCompare = new ICmpInst(
      |   ^~~~~~~~
Use -header-filter=.* to display errors from all non-system headers. Use -system-headers to display errors from system headers as well.
      |   auto
```
This is a common code, so I don't think we should address it, since rebase will be harder.

akiramenai (approved):
Thank you for cherry-picking.

PR #633, opened 2024-06-24
[EraVM] Change lowering of memcpy and memmove intrinsics

vladimirradosavljevic:


akiramenai (approved):
Could we rewrite commit messages to make them context free? Right now a reader must now what we expect from LSR and indexed mem ops prepare otherwise the message will not make any sense to him and it's simpler to read the patch in its entirety.
Could we just explain that we removed main IV from the loop as we prefer to not to call LSR in llc pipeline. We also put the loop into the form so that indexed memory operations are used.
Otherwise, LGTM.

vladimirradosavljevic:
> Could we just explain that we removed main IV from the loop as we prefer to not to call LSR in llc pipeline.

Sorry, but I don't follow this. Why we don't prefer to call LSR in llc pipeline?
Reason why I included LSR and EraVMIndexedMemOpsPrepare changes into this patch is to do hoisting of GEP instruction so we can reuse it in loop and remainder BBs (as it is written in the commit messages) and as tests show, with this we reduced some instructions.

akiramenai:
> Sorry, but I don't follow this. Why we don't prefer to call LSR in llc pipeline?

Indeed and it seems that we first unfold memcpy and then call LSR. Then I totally miss why LSR didn't do the work for us and why we had to rewrite the IR generated for memcpy manually.

vladimirradosavljevic:
> > Sorry, but I don't follow this. Why we don't prefer to call LSR in llc pipeline?
> 
> Indeed and it seems that we first unfold memcpy and then call LSR. Then I totally miss why LSR didn't do the work for us and why we had to rewrite the IR generated for memcpy manually.

Is it OK as is, or I should add that `LSR` and `EraVMIndexedMemOpsPrepare` are after lowering of memcpy and memmove, or something else?

akiramenai:
I got the idea now. I suggest the following structure of the commit message if I got everything right.

This patch removes a gep instruction to calculate the end dst address of memcpy in residual bb of memcpy expansion. [Main point comes first] To do so it removes the main IV in the expanded loop and replaces its condition with a comparison of dst pointer with end dst pointer. The same end dst pointer is reused in the residual bb. [How it was done]
Normally the loop transformation is to be done by Loop Strength Reduction pass but as we want to use dst pointer in residual bb, we rewrite the IR for lowering memory intrinsics beforehand. [Useful context / why this way]
[I would omit mentioning other minor changes as they aren't very relevant here]
[Link to the PR]

vladimirradosavljevic:
> I got the idea now. I suggest the following structure of the commit message if I got everything right.
> 
> This patch removes a gep instruction to calculate the end dst address of memcpy in residual bb of memcpy expansion. [Main point comes first] To do so it removes the main IV in the expanded loop and replaces its condition with a comparison of dst pointer with end dst pointer. The same end dst pointer is reused in the residual bb. [How it was done] Normally the loop transformation is to be done by Loop Strength Reduction pass but as we want to use dst pointer in residual bb, we rewrite the IR for lowering memory intrinsics beforehand. [Useful context / why this way] [I would omit mentioning other minor changes as they aren't very relevant here] [Link to the PR]

Thanks for the review. Please note that I slightly changed the first sentence.

PR #636, opened 2024-06-27
Cherry-pick [SimplifyCFG] Add support for sinking instructions with multiple uses

akiramenai:


akiramenai:
As we don't plan to address SAR regression now, closing the ticket. #644 is to address it when upgrading LLVM.

PR #641, opened 2024-07-01
ci: use era-compiler-llvm-test tag in integration tests if exists

antonbaliasnikov:
# Code Review Checklist

## Purpose

Use `era-compiler-llvm-test` tag in `era-compiler-tester` repository for candidate benchmarks and integration tests if this tag exists. If it's not, default to the default branch (`main`).


.github/workflows/benchmarks-integration-tests.yml
```diff
@@ -0,0 +1,110 @@
...
+    inputs:
+      llvm_build_type:
+        description: "LLVM build type: debug | release"
+        required: true
+        default: "release"
+      compiler_tester_reference_branch:
+        description: "compiler-tester branch to use as a benchmark reference"
+        required: true
+        default: "main"
+      compiler_tester_candidate_branch:
```
hedgar2017: It is a tag, not branch, isn't it?
Let's rephrase it to: defaults to branch `main` if the tag doesn't exist.

hedgar2017 (changes requested):

.github/workflows/benchmarks-integration-tests.yml
antonbaliasnikov: Fixed, please, take a look

hedgar2017 (approved):

hedgar2017 (changes requested):
Shall we not skip the CI here to check that the new scripts work?

hedgar2017 (approved):

PR #649, opened 2024-07-04
[EraVM] Add EraVM C-API to LLD

PavelKopyl:
# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion

### DIFF
--- a/lld/Common/DriverDispatcher.cpp
+++ b/lld/Common/DriverDispatcher.cpp
@@ -150,6 +150,22 @@ static Driver whichDriver(llvm::SmallVectorImpl<const char *> &argsV,
   return it->d;
 }
 
+// EraVM local begin
+static DriverMemBuf whichDriver(llvm::SmallVectorImpl<const char *> &argsV,
+                                llvm::ArrayRef<DriverDefMemBuf> drivers) {
+  Flavor f = parseFlavor(argsV);
+  const auto *const it =
+      llvm::find_if(drivers, [=](auto &driverdef) { return driverdef.f == f; });
+  if (it == drivers.end()) {
+    // Driver is invalid or not available in this build.
+    return [](llvm::ArrayRef<llvm::MemoryBufferRef>, llvm::raw_pwrite_stream *,
+              llvm::ArrayRef<const char *>, llvm::raw_ostream &,
+              llvm::raw_ostream &, bool, bool) { return false; };
+  }
+  return it->d;
+}
+// EraVM local end
+
 namespace lld {
 bool inTestOutputDisabled = false;
 
@@ -174,6 +190,35 @@ int unsafeLldMain(llvm::ArrayRef<const char *> args,
 
   return r;
 }
+
+// EraVM local begin
+/// Universal linker main(). This linker emulates the gnu, darwin, or
+/// windows linker based on the argv[0] or -flavor option.
+int unsafeLldMainMemBuf(llvm::ArrayRef<llvm::MemoryBufferRef> inBuffers,
+                        llvm::raw_pwrite_stream *outBuffer,
+                        llvm::ArrayRef<const char *> args,
+                        llvm::raw_ostream &stdoutOS,
+                        llvm::raw_ostream &stderrOS,
+                        llvm::ArrayRef<DriverDefMemBuf> drivers,
+                        bool exitEarly) {
+  SmallVector<const char *, 256> argsV(make_range(args.begin(), args.end()));
+  DriverMemBuf d = whichDriver(argsV, drivers);
+  // Run the driver. If an error occurs, false will be returned.
+  const int r = !d(inBuffers, outBuffer, argsV, stdoutOS, stderrOS, exitEarly,
+                   inTestOutputDisabled);
+  // At this point 'r' is either 1 for error, and 0 for no error.
+
+  // Call exit() if we can to avoid calling destructors.
+  if (exitEarly)
+    exitLld(r);
+
+  // Delete the global context and clear the global context pointer, so that it
+  // cannot be accessed anymore.
+  CommonLinkerContext::destroy();
+
+  return r;
+}
+// EraVM local end
 } // namespace lld
 
 Result lld::lldMain(llvm::ArrayRef<const char *> args,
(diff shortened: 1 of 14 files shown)


lld/Common/DriverDispatcher.cpp
```diff
@@ -201,3 +246,34 @@ Result lld::lldMain(llvm::ArrayRef<const char *> args,
...
+                          llvm::raw_ostream &stdoutOS,
+                          llvm::raw_ostream &stderrOS,
+                          llvm::ArrayRef<DriverDefMemBuf> drivers) {
+  int r = 0;
+  {
+    // The crash recovery is here only to be able to recover from arbitrary
+    // control flow when fatal() is called (through setjmp/longjmp or
+    // __try/__except).
+    llvm::CrashRecoveryContext crc;
+    if (!crc.RunSafely([&]() {
```
akiramenai, line 263: All pointers or references, so can be passed by value.

lld/Common/DriverDispatcher.cpp
```diff
@@ -201,3 +246,34 @@ Result lld::lldMain(llvm::ArrayRef<const char *> args,
...
+          r = unsafeLldMainMemBuf(inBuffers, outBuffer, args, stdoutOS,
+                                  stderrOS, drivers, /*exitEarly=*/false);
+        }))
+      return {crc.RetCode, /*canRunAgain=*/false};
+  }
+
+  // Cleanup memory and reset everything back in pristine condition. This path
+  // is only taken when LLD is in test, or when it is used as a library.
+  llvm::CrashRecoveryContext crc;
+  if (!crc.RunSafely([&]() { CommonLinkerContext::destroy(); })) {
```
akiramenai: no capture seems to be needed

lld/ELF/Driver.cpp
```diff
@@ -1750,7 +1797,14 @@ void LinkerDriver::createFiles(opt::InputArgList &args) {
       break;
     }
     case OPT_script:
-      if (std::optional<std::string> path = searchScript(arg->getValue())) {
+      // EraVM local begin
+      if (config->useIOMemoryBuffers) {
+        if (std::optional<MemoryBufferRef> mb = readFile(arg->getValue()))
```
akiramenai, line 1802: It looks like we want to avoid using `searchScript`. Could you explain the reason for that?

llvm/lib/Support/FileOutputBuffer.cpp
```diff
@@ -109,6 +109,43 @@ class InMemoryBuffer : public FileOutputBuffer {
...
+
+  Error commit() override {
+    OutStream << StringRef(static_cast<const char *>(Buffer.base()),
+                           BufferSize);
+    OutStream.flush();
+    return Error::success();
+  }
+
+private:
+  // Buffer may actually contain a larger memory block than BufferSize
```
akiramenai, line 143: Do you mean that owning memory block size >= BifferSize?

akiramenai (commented):
The patch is in a draft state but you mentioned it in the overview of the linker work, so I reviewed. Could you clarify what is left to be done?

PavelKopyl:
> The patch is in a draft state but you mentioned it in the overview of the linker work, so I reviewed. Could you clarify what is left to be done?

I've switched it for ready to review. I changed LLD C-API to support metadata of arbitrary length (8b2002fc159e937cddb8767c4def9ea5948ac91e).

lld/Common/DriverDispatcher.cpp
PavelKopyl: You are right, but answering to this and to the following questions, the functions added to DriverDispatcher.cpp are almost copies of existing functions. It seems reasonable for me to preserver original structure. What do you think?

lld/Common/DriverDispatcher.cpp
PavelKopyl: Please, check my comment above.

lld/ELF/Driver.cpp
PavelKopyl: You are right, we do not need `searchScript` when LLD In/Out is performed via memory buffers. Instead, we call   `readFile()` that contains a bit hackery code that instead of reading a file on OS just returns the reference to the passed memory buffer:
  ```
  ...
MemoryBufferRef mbref;
  if (config->useIOMemoryBuffers) {
    unsigned idx = 0;
    if (path.getAsInteger(10, idx)) {
      error(path + ": path should be an index");
      return std::nullopt;
    }
    if (idx >= config->inBuffers.size()) {
      error("memory buffer index is out of range");
      return std::nullopt;
    }
    mbref = config->inBuffers[idx];
    ...
```
where `arg->getValue()` is the memory buffer index, "0", or "1", so on.
Please, check the changes in `readFile()` in lld/ELF/InputFiles.cpp, line 234.

llvm/lib/Support/FileOutputBuffer.cpp
PavelKopyl: Yes. Please note this class is almost a copy of the `InMemoryBuffer` one that has the same comment. The meaning of this comment arises from the implementation of an interface from `llvm/include/llvm/Support/Memory.h`.  I guess the allocated memory region can be more than `BifferSize` because of alignment requirement of the starting buffer address.

lld/Common/DriverDispatcher.cpp
akiramenai: I leave it up to you.
I'm just more nervous when reading a lambda that captures by reference as it should put a C++ programmer in a state of higher awareness. The lambda doesn't escape the scope and actually captures nothing so this awareness returns back to the normal level almost immediately. Still I expect us to read our local changes more frequently than the rest of the lld code, so stylistic consistency wouldn't be a priority for me.

PavelKopyl: OK, agreed. Will fix it.

PavelKopyl: Sorry, I've just have looked again. At least `stdoutOS`, `stderrOS` and `r` cannot be captured by value. 
Do you mean to employ default-capture by value and these three objects by reference, i.e `[=, &stdoutOS, &stderrOS, &r]`?

lld/Common/DriverDispatcher.cpp
PavelKopyl: You are right not needed. Removed.

akiramenai:
Regression tests passed locally, so I merge.

PR #651, opened 2024-07-04
[EraVM] Adjust CMP to expose CSE opportunities with SUB

vladimirradosavljevic:


llvm/lib/Target/EraVM/EraVMConditionOptimizer.cpp
```diff
@@ -0,0 +1,308 @@
...
+  if (MI.getOpcode() != EraVM::SUBxrr_v)
+    return false;
+
+  const auto *FlagDef = MI.findRegisterDefOperand(EraVM::Flags);
+  if (!FlagDef || (FlagsDead != FlagDef->isDead()))
+    return false;
+
+  return EraVM::out0Iterator(MI)->getReg().isVirtual() &&
+         EraVM::in0Iterator(MI)->isCImm() &&
+         EraVM::in0Iterator(MI)->getCImm()->getBitWidth() == ImmBitWidth &&
```
akiramenai: Can't we assert it?

llvm/lib/Target/EraVM/EraVMTargetMachine.cpp
```diff
@@ -208,6 +209,7 @@ class EraVMPassConfig : public TargetPassConfig {
 
   void addIRPasses() override;
   bool addInstSelector() override;
+  bool addILPOpts() override;
```
akiramenai, line 212: Why do we add the pass as an ILP optimization?

llvm/lib/Target/EraVM/EraVMConditionOptimizer.cpp
```diff
@@ -0,0 +1,308 @@
...
+         EraVM::in1Iterator(MI)->getReg().isVirtual() &&
+         TII->getCCCode(MI) == EraVMCC::COND_NONE;
+}
+
+bool EraVMConditionOptimizer::isLargeImmSubSWithFlags(const MachineInstr &MI,
```
akiramenai: These two functions are always used in conjunction and have ~50% of common code. Perhaps it's worth merging.

akiramenai (approved):
Thank you! Documentation in code is really helpful!
One note. We started to use this big immediate thing which is wrong usage of the term, and it confuses. I don't ask to rename anything, but it's better to stop to refer to non-immediate constants as 'big imm' in future patches.

llvm/lib/Target/EraVM/EraVMConditionOptimizer.cpp
vladimirradosavljevic: Merged them into one function and added check that CPI operand should be without offest. PTAL.

llvm/lib/Target/EraVM/EraVMConditionOptimizer.cpp
vladimirradosavljevic: Added assert that we expect 256-bit `CImm` here.

llvm/lib/Target/EraVM/EraVMTargetMachine.cpp
vladimirradosavljevic: Since we need to run this pass before `MachineCSE`, we could either add it in `addMachineSSAOptimization` (and call base implementation after that) or in a `addILPOpts` hook that is called just before MachineLICM and MachineCSE.
The only reason why I decided to run it as an ILP optimization is because `EarlyTailDuplicate` is called before that, and if we want to implement same optimization as in `AArch64ConditionOptimizer` in a future, it will benefit from it since it can catch more cases.

PR #658, opened 2024-07-11
Cherry-pick [SelectionDAG] Use LAST_INTEGER_VALUETYPE instead of i64 (#98299)

akiramenai:
External review: https://github.com/llvm/llvm-project/pull/98299


vladimirradosavljevic (approved):
LGTM. Good catch!

PR #667, opened 2024-07-29
[BranchFolding] Add an option to tail merge only bbs without successors

vladimirradosavljevic:
One of the benefits of Tail Merge optimization is to reduce code size by merging identical tails of the BBs. In this case, if two BBs have the same ending instructions, one of the BBs will do the fallthrough to the tail BB and other will use jump instruction to do so. In case BB where jump is generated is frequently executed, at least for EraVM this can hurt performance. On the other hand, for returning BBs we want to do tail merge since after these BBs we will return from the function and they are not frequently executed. Because of this, `-tail-merge-only-bbs-without-succ` option is introduced to boost performance in such cases and to reduce code size by doing tail merging for returning BBs.


vladimirradosavljevic:
These changes in performance numbers are unrelated to this PR, since the option is turned off by default.

akiramenai (approved):
LGTM, but the performance numbers

PR #671, opened 2024-08-05
[EraVM] Only generate GEPs that dominate latch BB in EraVMIndexedMemOpsPrepare

vladimirradosavljevic:
…OpsPrepare

In case we generate GEP that doesn't dominate latch BB, we will run into an issue and verifier will complain:
```
Instruction does not dominate all uses!
  %3 = getelementptr inbounds i256, ptr addrspace(1) %0, i256 1
  %0 = phi ptr addrspace(1) [ %arg3, %bb1 ], [ %3, %bb5 ]
```
This patch fixes this issue.

### DIFF
--- a/llvm/test/CodeGen/EraVM/indexed-memops-gep-dominate-bug.ll
+++ b/llvm/test/CodeGen/EraVM/indexed-memops-gep-dominate-bug.ll
@@ -1,0 +1,43 @@
+; RUN: llc -O3 -stop-after=verify -compile-twice=false < %s | FileCheck %s
+
+target datalayout = "E-p:256:256-i256:256:256-S32-a:256:256"
+target triple = "eravm"
+
+define i256 @test(i256 %arg, ptr addrspace(1) %arg1, ptr addrspace(1) %arg2) {
+; CHECK-LABEL: @test(
+entry:
+  %icmp1 = icmp eq i256 %arg, 0
+  br i1 %icmp1, label %bb7, label %bb1
+
+bb1:
+  br label %bb2
+
+bb2:
+  %phi1 = phi i256 [ %add, %bb5 ], [ 0, %bb1 ]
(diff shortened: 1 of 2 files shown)


akiramenai (approved):

llvm/test/CodeGen/EraVM/indexed-memops-gep-dominate-bug.ll
```diff
@@ -0,0 +1,73 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
+; RUN: llc -O3 < %s | FileCheck %s
```
akiramenai: An IR level test that only runs EraVMIndexedMemOpsPrepare could be a better option here.

vladimirradosavljevic: I agree, but we also need to run verify pass, since there was an assert. Because of this, I have added support for new pass manager to `EraVMIndexedMemOpsPrepare` similar to [this](https://github.com/matter-labs/era-compiler-llvm/commit/5b885a6ece97b8df7cb8bdedda11e50e0cf00723) one. Unfortunately there is a catch, `EraVMIndexedMemOpsPrepare` is a loop pass and LLVM new pass manager machinery is adding `LoopSimplifyPass` and `LCSSAPass` before loop passes. This is a problem since if we want to run `EraVMIndexedMemOpsPrepare` with `opt` we would get different pipeline than we use. Because of this, I dropped new pass manager support for `EraVMIndexedMemOpsPrepare`.
Please note that I changed this test to stop after verify pass, so we make sure assert is not triggered.

vladimirradosavljevic:
@akiramenai I included two more patches in this PR, so please review this again.

akiramenai (approved):
LGTM, thank you.

PR #673, opened 2024-08-07
[EraVM] Disable generation of .note.GNU-stack

vladimirradosavljevic:
This section is not used, so disable generation of it.


PavelKopyl (approved):
LGTM.

PR #682, opened 2024-08-16
[EraVM] Adding support of library linkage

PavelKopyl:
# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


lld/include/lld-c/LLDAsLibraryC.h
```diff
@@ -23,7 +23,9 @@ LLVM_C_EXTERN_C_BEGIN
     Input/output files are transferred via memory buffers. */
 LLVMBool LLVMLinkEraVM(LLVMMemoryBufferRef inBuf, LLVMMemoryBufferRef *outBuf,
                        const char *metadataPtr, uint64_t metadataSize,
-                       char **ErrorMessage);
+                       const char *const *linkerSymbol,
+                       const char linkerSymbolValue[][20],
```
hedgar2017: Magic number?

hedgar2017 (changes requested):

lld/include/lld-c/LLDAsLibraryC.h
```diff
@@ -23,7 +23,9 @@ LLVM_C_EXTERN_C_BEGIN
     Input/output files are transferred via memory buffers. */
 LLVMBool LLVMLinkEraVM(LLVMMemoryBufferRef inBuf, LLVMMemoryBufferRef *outBuf,
                        const char *metadataPtr, uint64_t metadataSize,
-                       char **ErrorMessage);
+                       const char *const *linkerSymbol,
```
hedgar2017: => `linkerSymbols`

lld/include/lld-c/LLDAsLibraryC.h
hedgar2017: => `linkerSymbolValues`

hedgar2017 (changes requested):

lld/include/lld-c/LLDAsLibraryC.h
PavelKopyl: Thank you, I've defined it.

lld/include/lld-c/LLDAsLibraryC.h
PavelKopyl: Fixed.

lld/include/lld-c/LLDAsLibraryC.h
PavelKopyl: Thank you, fixed.

lld/ELF/Arch/EraVM.cpp
```diff
@@ -59,6 +59,9 @@ void EraVM::relocate(uint8_t *loc, const Relocation &rel, uint64_t val) const {
   case R_ERAVM_16_SCALE_8:
     add16scaled(val, /*scale=*/8);
     break;
+  case R_ERAVM_32:
+    write32be(loc, static_cast<uint32_t>(val & 0xFFFFFFFF));
```
akiramenai: I didn't get why and with the mask if you truncate it to 4 bytes anyway.

lld/include/lld-c/LLDAsLibraryC.h
```diff
@@ -19,11 +19,34 @@
 
 LLVM_C_EXTERN_C_BEGIN
 
+#define LINKER_SYMBOL_SIZE 20
```
akiramenai, line 74: Could we document what makes us confident that a symbol doesn't exceed this size limit?

lld/include/lld-c/LLDAsLibraryC.h
```diff
@@ -19,11 +19,34 @@
...
+
 /** Performs linkage operation of an object code via lld-as-a-library.
     Input/output files are transferred via memory buffers. */
-LLVMBool LLVMLinkEraVM(LLVMMemoryBufferRef inBuf, LLVMMemoryBufferRef *outBuf,
-                       const char *metadataPtr, uint64_t metadataSize,
-                       char **ErrorMessage);
+LLVMBool LLVMLinkEraVM(LLVMMemoryBufferRef inBuffer,
+                       LLVMMemoryBufferRef *outBuffer,
+                       const char *const *linkerSymbols,
```
akiramenai: Could we document the parameters? `linkerSymbols` seems to be connected with `linkerSymbolValues` but if there are constraints on where they should point to they aren't obvious from the interface.

lld/include/lld-c/LLDAsLibraryC.h
```diff
@@ -19,11 +19,34 @@
...
+LLVMBool LLVMIsELF(LLVMMemoryBufferRef inBuffer);
+
+/** Returns undefined linker symbol names of the ELF object file passed
+ *  in \p inBuffer. The names are returned via \p linkerSymbols as the
+ *  array of null-terminated strings. \p numLinkerSymbols will point to
+ *  the number of names. For example, if the file has undefined symbols of
+ *  the form "library_path_[0-4]", then this returns just "library_path".
+ *  Caller should dispose the memory allocated for the \p linkerSymbols
+ *  using LLVMDisposeUndefinedSymbolsEraVM */
+void LLVMGetUndefinedSymbolsEraVM(LLVMMemoryBufferRef inBuffer,
```
akiramenai: Is a triple point necessary here? We return `void` and we allocate memory so could we use the return type instead?

lld/lld-c/LLDAsLibraryC.cpp
```diff
@@ -4,50 +4,135 @@
...
-                  .str();
-  } else {
-    // There is no metadata, but anyway the final size should be the odd number
-    // of words. If the size is already an odd number, we are done.
-    // Otherwise add one zeroed word.
-    padding = ". = ((. >> 5) | 1) << 5;\n";
+/// Returns size of the section \p SectionName in the object file \p file.
+/// Returns zero, if there is no such section.
+static uint64_t getSectionSize(ObjectFile &file, StringRef sectionName) {
+  for (const SectionRef &sec : file.sections()) {
```
akiramenai: Nit: `find` will express the intent more clearly.
Overall it's usually preferable to use algorithms where applicable as when a reader sees an algorithm he can infer much more information straightaway without even reading the body. In this particular case the fact that it's a lookup and the fact that section is supposed to be uniquely named.

lld/lld-c/LLDAsLibraryC.cpp
```diff
@@ -4,50 +4,135 @@
...
+        (symFlags & object::SymbolRef::SF_Undefined)) {
+      StringRef symName = cantFail(sym.getName());
+      if (!symbolsToBeDefined.contains(symName))
+        return true;
+    }
+  }
+  return false;
+}
+
+/// Returns a string with the linker symbol definitions passed in
```
akiramenai, line 104: It explains `LINKER_SYMBOL_SIZE` and the logic of the previous function. I would extract the parts that describe the design part of the solution and put it into the header.

lld/include/lld-c/LLDAsLibraryC.h
akiramenai: After reading the definition: Maybe `linkerSymbolNames` and `linkerSymbolValues`. It makes clear how they are connected.

lld/lld-c/LLDAsLibraryC.cpp
```diff
@@ -4,50 +4,135 @@
...
+              .str();
+      symbolsStr += ";\n";
+    }
+  }
+  return symbolsStr;
+}
+
+/// Returns a linker script when linking executable file.
+static std::string
+creteEraVMLinkerScript(uint64_t metadataSize, const char *const *linkerSymbols,
```
akiramenai: typo

lld/lld-c/LLDAsLibraryC.cpp
```diff
@@ -76,23 +163,51 @@ SECTIONS {                                                                  \n\
...
+    StringRef inData = localInMemBufRefs[0].getBuffer();
+    *outBuffer = LLVMCreateMemoryBufferWithMemoryRangeCopy(
+        inData.data(), inData.size(), "result");
+    return false;
+  }
+
+  uint64_t MDSize = getSectionSize(*static_cast<ObjectFile *>(InBinary.get()),
+                                   ".eravm-metadata");
+
+  std::string linkerScript =
```
akiramenai, line 319: LinkerSymbolDefs isn't a linker script, is it?
Perhaps worth refactoring or documenting what it means and implies if
* we don't have undef symbols
* we have undefs but 0 linkerSymbols
* we have have both undefs and linker symbols.

Otherwise, the logic below seems convoluted.

akiramenai (changes requested):
Could we put some effort into improving readability/maintainability?

PavelKopyl:
> Could we put some effort into improving readability/maintainability?

Thank you very much for your comments! Sure, I'll address them. But I'd start with renaming "linker symbol" -> "library symbol" everywhere, but the intrinsic name. @hedgar2017, @akiramenai, are you OK with this? Original terminology "linker symbol" came from Solidity docs, but it is meaningless in LLVM-based tool chain. Essentially, these are usual symbols (in terminology of ELF format) that are resolved to the library addresses at a linking time. The only difference from the default workflow is that on MC level each library symbol is split into 5 sub-symbols because a library address is a 20-byte relocatable value, but our ELF format is ELF-32 one.

lld/ELF/Arch/EraVM.cpp
PavelKopyl: Sure, that's not needed!

lld/include/lld-c/LLDAsLibraryC.h
PavelKopyl: Added the description.

lld/include/lld-c/LLDAsLibraryC.h
PavelKopyl: I agree. I replaced `linkerSymbols`  -> `linkerSymbolNames` and documented them.

lld/include/lld-c/LLDAsLibraryC.h
PavelKopyl: I agree, but let's fix it after all other fixes, because we also need to update the FE part.

lld/lld-c/LLDAsLibraryC.cpp
PavelKopyl: Done.

lld/lld-c/LLDAsLibraryC.cpp
PavelKopyl: Done.

lld/lld-c/LLDAsLibraryC.cpp
PavelKopyl: Thanks.

lld/lld-c/LLDAsLibraryC.cpp
PavelKopyl: In both cases we generate linker script. In case we are going emit an ELF file, the script is trivial, as it just contains linker symbol definitions. I've added the comments.

akiramenai (commented):
Documentation and code improvements look great, thank you.
I agree that library symbols make more sense in terms of terminology. @hedgar2017 is it ok to rename?
Except for that and remained triple pointer, LGTM.

hedgar2017:
> Documentation and code improvements look great, thank you.
> 
> I agree that library symbols make more sense in terms of terminology. @hedgar2017 is it ok to rename?
> 
> Except for that and remained triple pointer, LGTM.

I don't think it makes sense to make linker know about Solidity libraries.

Even Yul is agnostic of libraries and only features `linkersymbol` with an address. In handwritten Yul it is possible to link any address, not just a library.

PavelKopyl:
> Documentation and code improvements look great, thank you. I agree that library symbols make more sense in terms of terminology. @hedgar2017 is it ok to rename? Except for that and remained triple pointer, LGTM.

Thank you for the review!
As for the usage of "library symbol" terminology, after discussing this with @hedgar2017 we decided to keep the "linker symbol".
This seems reasonable because this patch, in fact, implements a generalization of relocatable symbol. In the ELF format the max width of a symbol value is 64 bits. Here we extended it to 160 bits. Yes, currently the only application of this feature is the libraries support, but in future we can employ this for other things.

akiramenai (approved):
Please squash

hedgar2017 (approved):

hedgar2017:
You may just remove the lock, as it is not complete and FE is not going to be multi-threaded here anymore.

PavelKopyl:
> You may just remove the lock, as it is not complete and FE is not going to be multi-threaded here anymore.

Yep, will remove it.

PR #684, opened 2024-08-20
[WIP][EraVM] Set AllowRegisterRenaming to 1

vladimirradosavljevic:
This will allow MachineCopyPropagation queries
to work properly.

### DIFF
--- a/llvm/lib/Target/EraVM/EraVM.td
+++ b/llvm/lib/Target/EraVM/EraVM.td
@@ -62,4 +62,5 @@ def EraVMAsmParser : AsmParser {
 def EraVM : Target {
   let InstructionSet = EraVMInstrInfo;
   let AssemblyParsers = [EraVMAsmParser];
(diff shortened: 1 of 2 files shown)


akiramenai:
> This will allow MachineCopyPropagation queries to work properly.

Could 'properly' be elaborated?

vladimirradosavljevic:
> > This will allow MachineCopyPropagation queries to work properly.
> 
> Could 'properly' be elaborated?

Will do it (pass is skipping copy instructions since they are not marked as `Renamable` like for e.g. [here](https://github.com/matter-labs/era-compiler-llvm/blob/a8bfb396754c98356bb15bdf843c8f2f05c4e94e/llvm/lib/CodeGen/MachineCopyPropagation.cpp#L921)), but please note this is WIP (I updated it now), since I want to check if we need to add special handling of fatptr copies.

vladimirradosavljevic:
Closed in favor of #712.

PR #694, opened 2024-08-26
ci: add coverage measurements workflow

antonbaliasnikov:
# Code Review Checklist

* [x] Add code coverage measurements by compiler tester
* [x] Save HTML archive with coverage results 
* [x] Upload coverage results to codecov.io 
* [x] Trigger with each merge to main for now 

## Testing

All testing workflows are here:
https://github.com/matter-labs/era-compiler-llvm/actions/workflows/coverage.yml

[Example of coverage report](https://github.com/matter-labs/era-compiler-llvm/actions/runs/10575697276/artifacts/1858920725)

[Codecov link](https://app.codecov.io/gh/matter-labs/era-compiler-llvm)

## Purpose

To allow checking which parts of LLVM are covered by running `era-compiler-tester` with all available optimization levels.

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,0 +1,116 @@
+name: Code coverage
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Mode filter for era-compiler-tester"
+        required: true
+        default: "0.8"
+      path:
+        description: "Path filter for era-compiler-tester"
+        required: true
+        default: "tests/solidity/complex/defi"
+
+defaults:
+  run:
+    shell: bash -ex {0}
+
+concurrency:
+  group: ${{ github.repository_id }}-${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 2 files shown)


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,116 @@
...
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+    inputs:
+      mode:
+        description: "Mode filter for era-compiler-tester"
+        required: true
+        default: "0.8.26"
```
hedgar2017: This will probably stop working when we update to 0.8.27 ><

hedgar2017 (approved):

.github/workflows/coverage.yml
antonbaliasnikov: Yeah... it would be nice if we can support `latest` filter for `solc` in our tester.

antonbaliasnikov: Oh, actually `0.8` works nice. I can use that.

PR #696, opened 2024-08-29
[EraVM] Fix EraVMCombineAddressingMode::mergeSelect for SEL with stac…

vladimirradosavljevic:
…k as output

This patch fixes the issue in mergeSelect where Base is SEL with stack as output and that output operand is not copied to the newly created SEL. This happens when we are replacing second operand of SEL (IsIn0 is false) with In, and we are not copying rest of the operands, just conditional code.


akiramenai (approved):
LGTM, thank you

PR #699, opened 2024-09-04
[APInt] Replace an assertion in constructor with fatal error

akiramenai:
The patch replaces assertions in APInt constructor with `report_fatal_error` calls.
EraVM compiler is shipped to users w.o. assertions enabled, but zero extensions of negative numbers is a strong signal that a transformation which called the constructor has a bug. In that case we prefer to fail with diagnostic instead of miscompiling a module even if false positives could happen.

### DIFF
--- a/llvm/include/llvm/ADT/APInt.h
+++ b/llvm/include/llvm/ADT/APInt.h
@@ -16,6 +16,9 @@
 #define LLVM_ADT_APINT_H
 
 #include "llvm/Support/Compiler.h"
+// EraVM local begin
+#include "llvm/Support/ErrorHandling.h"
+// EraVM local end
 #include "llvm/Support/MathExtras.h"
 #include <cassert>
 #include <climits>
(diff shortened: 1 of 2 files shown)


llvm/include/llvm/ADT/APInt.h
```diff
@@ -113,13 +116,21 @@ class [[nodiscard]] APInt {
...
-               "Value is not an N-bit signed value");
+        // EraVM local begin
+        if (!llvm::isIntN(BitWidth, val))
+          report_fatal_error("APInt error: Value is not an N-bit signed value");
+        // EraVM local end
       } else {
-        assert(llvm::isUIntN(BitWidth, val) &&
-               "Value is not an N-bit unsigned value");
+        // EraVM local begin
+        if (!llvm::isUIntN(BitWidth, val))
```
vladimirradosavljevic, line 130: I'm not sure this will catch bitwidth larger than 64, since `isUIntN` and `isIntN` will return true for BitWidth >= 64. Please take a look [here](https://github.com/matter-labs/era-compiler-llvm/blob/0482c7a514b7e9dfec713edf5d1ea204249dbcd4/llvm/include/llvm/Support/MathExtras.h#L228).

akiramenai: It will not catch indeed as no truncation happens in that case. And there is no way to differentiate your intent but the correctly specified `isSignedFlag`. If you do something like `APInt(256, -1, false)` you might indeed meant `APInt(256, 0xFFFFFFFFFFFFFFFF, false)`. However, if you do `APInt(32, -1, false)` it really looks like a bug.
Note that this piece of logic was accepted upstream (with assertion, not fatal errors): https://github.com/llvm/llvm-project/pull/106524

vladimirradosavljevic: FWIW, I think we should make this change, since it can catch some miscompilation issues.

akiramenai: What change do you refer to?

vladimirradosavljevic: To change asserts with fatal errors, not the last change you introduced as experiment AFAIU.

vladimirradosavljevic (approved):
LGTM. Thanks!

PR #703, opened 2024-09-12
[EraVM] Add support for splitting live ranges of PHI nodes in loops

vladimirradosavljevic:


llvm/test/CodeGen/EraVM/split-loop-phi-live-ranges.ll
```diff
@@ -29,19 +29,19 @@ define i256 @test() {
 ; CHECK-NEXT:    [[ADD2:%.*]] = add nuw nsw i256 [[RESULT]], 64
 ; CHECK-NEXT:    [[INTTOPTR0:%.*]] = inttoptr i256 [[ADD2]] to ptr addrspace(1)
 ; CHECK-NEXT:    store i256 0, ptr addrspace(1) [[INTTOPTR0]], align 1
-; CHECK-NEXT:    [[ADD3:%.*]] = add nuw nsw i256 [[RESULT]], -64
+; CHECK-NEXT:    [[ADD3:%.*]] = add i256 [[RESULT]], -64
```
akiramenai, line 32: Shouldn't we keep flags here?
Saving them below would take more effort but given `ADD3 = add nuw, nsw` and original `ADD4 = add nuw, nsw` there must be no wrap even when substituting `RESULT` with `ADD3`.

akiramenai (changes requested):

llvm/test/CodeGen/EraVM/split-loop-phi-live-ranges.ll
vladimirradosavljevic: I found cases with alive2, where not removing flags from the NearestDominator is leading to issues, like [here](https://alive2.llvm.org/ce/z/-KVMhN).

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+/// operand. This is useful for loops with large switch statements where PHI
+/// nodes are used frequently, and we want to keep these variables in a single
+/// register.
+/// We want to avoid following case: | And ideally, we want to have:
+///   header:                        |   header:
+///     %r1 = phi                    |     %r1 = phi
+///     ...                          |     ...
+///   bb1:                           |   bb1:
+///     %r2 = add %r1, 1             |     %r1 = add %r1, 1
+///     bcc bb2                      |     bcc bb2
```
akiramenai: In the example there is a mix of MIR with physical registers and phis while we are doing transformation on LLVM IR level. Could it be better explained as I still can't comprehend what makes us to have a copy on the left and frees from it on the right?

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+///     ...                          |     ...
+///     %add3 = add %phi, -96        |     %add3 = add %add1, -32
+///     ...                          |     ...
+///     b latch                      |     b latch
+///   latch:                         |   latch:
+///     bcc header                   |     bcc header
+///
+/// To achieve this, we need to find all users of the PHI node in the loop
+/// that are add instructions with constant operands. We are saving the
+/// original immediate operand of the instructions so we can correctly
```
akiramenai: immediate -> constant.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+/// to split the live range in the dominatee blocks. Finally, we are finding
+/// for each block the nearest dominator block from which we can split the
+/// live range and updating the instructions in the block with the nearest
+/// dominator instruction.
+static bool splitLoopPHILiveRanges(Function &F, LoopInfo &LI,
+                                   DominatorTree &DT) {
+  if (!EnableSplitLoopPHILiveRanges)
+    return false;
+
+  // This structure is used to store information about instructions in a block
```
akiramenai: Better is to add documentation to the data fields.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+    // Get the last instruction that will be used to split the live range.
+    InstInfo &getLastSplitInst() {
+      assert(LastSplitIdx != UINT64_MAX && "Expected valid index");
+      return Insts[LastSplitIdx];
+    }
+  };
+
+  bool Changed = false;
+  for (auto *L : LI) {
+    for (auto &PN : L->getHeader()->phis()) {
```
akiramenai: Nit: I recommend renaming to `Phi`, `PhiInst` as `PN` is not a common abbreviation and in the middle of the loop one might need to go to the definition to recall what is `PN`.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+      MapVector<BasicBlock *, InstSplitInfo> Splits;
+      for (auto *U : PN.users()) {
+        // Only collect add instructions with constant operands in this loop.
+        auto *UI = cast<Instruction>(U);
+        if (!UI || UI->getOpcode() != Instruction::Add ||
+            !isa<ConstantInt>(UI->getOperand(1)) ||
+            LI.getLoopFor(UI->getParent()) != L)
+          continue;
+
+        auto [Itr, _] = Splits.insert({UI->getParent(), {}});
```
akiramenai: `Splits[UI->getParent()].addInst(UI);`?

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+        Itr->second.addInst(UI);
+      }
+
+      // Skip if there are no at least two blocks where we can split
```
akiramenai: It repeats the code below. Could we explain why we need at least 2 blocks instead?

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+          auto *I = Splits.find(Node->getBlock());
+          if (I != Splits.end()) {
+            // We found the nearest dominator block, so take the last
+            // instruction from it.
+            NearestDominator = &I->second.getLastSplitInst();
+            break;
+          }
+
+          // Bail out if we reached the start of the loop.
+          if (Node == RootNode)
```
akiramenai: as a paranoia check, we may add that Node belongs to the loop.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
...
+
+        // Update operands in this block with the nearest dominator.
+        // Basically, we are doing the following transformation:
+        //   %Dominator = add %In, DomImm
+        //   %I         = add %In, Imm
+        // to
+        //   %Dominator = add %In, DomImm
+        //   %I         = add %Dominator, Imm - DomImm
+        auto [DomInst, DomImm] = *NearestDominator;
+        for (auto &Info : Infos.Insts) {
```
akiramenai: Big mental complexity of the function (nested loops of depth 4). Perhaps it is worth extracting the rewrite logic and other parts into separate functions and calling them from the loop.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
```diff
@@ -326,11 +317,213 @@ bool EraVMPostCodegenPrepare::rearrangeOverflowHandlingBranches(Function &F) {
   return Changed;
 }
 
+/// This optimization finds PHI nodes in loops with a large number of users
+/// and splits their live ranges, if users are add instructions with constant
```
akiramenai: a constant operand

akiramenai (commented):
I appreciate the documentation in the code, it helped me a lot. Still, I can't fully comprehend the patch as logic seems complex to me.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Updated the documentation.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: `PN` is abbreviation of `PHINode` that is used in LLVM code base, but I agree that `Phi` is easier to read. Changed to `Phi`.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Done. Thanks for suggestion!

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Added explanation why we need at least 2 blocks.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Split this code into 3 functions.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: Done.

llvm/lib/Target/EraVM/EraVMPostCodegenPrepare.cpp
vladimirradosavljevic: We can't do this, since we could traverse over subloop BBs, like in this example:
```
bb1:
  dominator add
...
subloop bbs:
...
bb2:
  dominatee add
```
So when we search from bb2 to find bb1, we could traverse over subloop bbs. Please note that BBs in the map are guaranteed to be in a loop, since there is a check for that.

vladimirradosavljevic:
> I appreciate the documentation in the code, it helped me a lot. Still, I can't fully comprehend the patch as logic seems complex to me.

Idea to do this optimization came when I was investigating performance drop (caused by regalloc not being able to preserve variable in single register and added unnecessary copy instructions in all opcodes) after one of the changes in the EVMInterpreter. We had following case in one of the opcode in the EVMInterpreter:
```
bb1:
  %add1 = add %phi, 64
  bcc bb2
bb2:
  %add2 = add %phi, -64
  b latch
```
After I manually changed this to:
```
bb1:
  %add1 = add %phi, 64
  bcc bb2
bb2:
  %add2 = add %add1, -128
  b latch
```
regalloc managed to preserve variable in a single register and removed a lot of copy instructions.
This is far from ideal, since the ideal fix would be to change regalloc (I don't think loop with big switch statement is a common case + other architectures won't have these problems given the number of registers they have and compared to us, copy is not expensive since we pay gas for each instruction), but given the time we have, this seemed to me as a better idea to do.
This support is essential for the lazy stack implementation (you can see numbers [here](https://github.com/matter-labs/era-compiler-tester/pull/98)), since it removes a lot of copy instructions.
Improvements that we can see here are fortunate, since for the current implementation of the EVMInterpreter, this optimization removes 1 instruction since it does following:
Before this optimization:
```
... Loop header
	sub.s	31, r14, r1
	ldm.h	r1, r1

... ADD opcode
	add	1, r14, r14      (ip + 1)
```
After this optimization:
```
... Loop header
	sub.s	31, r14, r1     (ip - 31)
	ldmi.h	r1, r2, r14     (ip - 31 + 32 = ip + 1)

... ADD opcode (no ip + 1, since it is calculated by above ldmi.h instruction)
```
I will create a ticket for this kind of optimization, since we could benefit from indexed loads and stores (there are optimizations in DAGCombine for this) and we can do something similar in MIR to create more opportunities for post increment loads and stores.

Btw, my idea is to disable this optimization by default and to enable it only for EVMInterpreter. I don't think we should do this optimization for other contracts.

akiramenai (approved):
LGTM. Thank you for putting effort into refactoring.

PR #709, opened 2024-09-30
[EraVM] Disable splitting loop phi live ranges by default

vladimirradosavljevic:
This optimization should only be used for EVMInterpreter.


akiramenai:
Based on the benchmarks, shouldn't we keep it enabled by default?

vladimirradosavljevic:
> Based on the benchmarks, shouldn't we keep it enabled by default?

Currently, improvements comes from https://github.com/matter-labs/era-compiler-llvm/issues/710 optimization, since it is fortunate that by splitting loop phi live ranges we are generating add instructions from which we are generating indexed load and store instructions. Since this is only intended for EVMInterpreter, I think it is safer to disable it by default and to enable it only for EVMInterpreter. I expect for #710 to bring more improvements for other contracts than we are seeing here.

akiramenai:
It's still -0.2% cycles in real-life m3b3 in unintended but understandable manner. Indeed the change we done for EVMInterpreter could result in better use of incremented loads and stores though I agree if to implement #410 the expected effect should be greater. So I'd suggest to attach this patch to #410 and close it for now. When #410 is done, it's worth rechecking the effect of the default. wdyt?

vladimirradosavljevic:
> It's still -0.2% cycles in real-life m3b3 in unintended but understandable manner. Indeed the change we done for EVMInterpreter could result in better use of incremented loads and stores though I agree if to implement #410 the expected effect should be greater. So I'd suggest to attach this patch to #410 and close it for now. When #410 is done, it's worth rechecking the effect of the default. wdyt?

Even though it looks like a nice improvement for real-life m3b3, it is just for 1 test:
```
Group 'Real life M3B3' cycles (-%) worst 100 out of 1:
 -0.419   : E+M3B3 0.6.12            tests/solidity/complex/defi/starkex-verifier/test.json::dydx_proof_verification[#fallback:47]
```
I don't think it is worth to enable it by default, given improvements are only for one test for 0.6.12 + there is a possibility to break indexed loads and stores in other real world contracts.
Does this make sense?

akiramenai:
Ok, merging.

akiramenai (approved):

PR #727, opened 2024-10-29
[TailDuplication] Add option to enable tail duplication of fallthrough BBs

vladimirradosavljevic:


akiramenai (commented):
I assume it improved numbers on EvmEmulator. Could we provide clearer guidance on when this option is beneficial?
> There are cases where we want to duplicate fallthrough BBs as well (e.g. EvmEmulator)

This explanation is vague. Additionally, the test case removes an empty BB (which ideally shouldn’t appear post-optimization), it doesn’t provide guidance - though it works fine as a test.
Could this be upstreamed?

vladimirradosavljevic:
> I assume it improved numbers on EvmEmulator. 

Yes, you can see the numbers [here](https://github.com/matter-labs/era-compiler-tester/pull/116#issuecomment-2443997861) (please use edit button to check the numbers before this).
> This explanation is vague.

I have changed the explanation, so PTAL.
>  Additionally, the test case removes an empty BB (which ideally shouldn’t appear post-optimization), it doesn’t provide guidance - though it works fine as a test.

To provide guidance through the test case, we would need to create a test which contains a loop with a switch case and a latch BB that is a fallthrough, and I don't think it is worth it, since current test is testing the option we introduced.

> Could this be upstreamed?

I tried to enable this by default, but the numbers are not clear win. This option is to cover an edge case that is not common, so we wouldn't have anything to convince the community that this should be merged.

akiramenai (approved):
LGTM, thank you!

PR #734, opened 2024-11-06
[EraVM][EVM] Use std::string instead of Twine

vladimirradosavljevic:
Twine variables are prone to use-after-free bugs, as it is stated in the clang-tidy documentation:
https://clang.llvm.org/extra/clang-tidy/checks/llvm/twine-local.html


PavelKopyl (approved):
LGTM, thanks!

PR #755, opened 2025-01-14
[EraVM] Add support for sinking loads to their users

vladimirradosavljevic:


akiramenai:
Can we run tests and benchmarks with
```
    EnableSinkLoads("eravm-enable-load-sinking", cl::Hidden, cl::init(true),
                    cl::desc("Enable sinking loads to their users"));
```
to check if it could be universally useful and to improve testing coverage?

akiramenai:
1. Could we add the real rationale for adding this change? I suspect we got some improvements on EVM Interpreter. Could we post it to the PR and link the PR in the commit message?
2. Could we explain why a normal LLVM sinking pass is not applicable?
3. Could we explain the layering, doing it on IR rather than MIR?

Except for unclarity in the documentation, the patch looks good to me.

vladimirradosavljevic:
> 1. Could we add the real rationale for adding this change? I suspect we got some improvements on EVM Interpreter. Could we post it to the PR and link the PR in the commit message?
> 2. Could we explain why a normal LLVM sinking pass is not applicable?
> 3. Could we explain the layering, doing it on IR rather than MIR?
> 
> Except for unclarity in the documentation, the patch looks good to me.

1. Yes, this optimization is related to [this](https://github.com/matter-labs/era-contracts/pull/1183) change that is still not merged, so there are no CI runs where the improvements can be seen. When that PR is merged, I will get the numbers in era-compiler-tester.
2. LLVM sinking passes are not doing sinking + duplication, just sinking. Only similar optimization that I have found is `CodeGenPrepare::tryToSinkFreeOperands`, but it is not checking whether there are memory clobbers between instructions and its user.
3. I have implemented this on IR level since I find it easier to detect clobbers (`isModSet(BAA.getModRefInfo(&I, Loc))`).

akiramenai (approved):

PR #759, opened 2025-01-21
[EVM] Improvements

vladimirradosavljevic:


akiramenai:
@vladimirradosavljevic is there a reason to merge it to `ef-stackification` rather than `main`?

vladimirradosavljevic:
> @vladimirradosavljevic is there a reason to merge it to `ef-stackification` rather than `main`?

I tested this and regenerated tests against `ef-stackification`, so we don't need to do it when we merge `ef-stackification`. My suggestion is to merge this after `ef-stackification` is merged into main. Wdyt?

llvm/test/CodeGen/EVM/context-remat.ll
```diff
@@ -0,0 +1,293 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
+; RUN: llc -O3 < %s | FileCheck %s
+
+target datalayout = "E-p:256:256-i256:256:256-S256-a:256:256"
+target triple = "evm"
+
+declare void @use(i256)
```
PavelKopyl, line 7: Though compilation to asm works well here, attempt to emit an object file will cause a crash. That's because for EVM we do not support compilation units. 
I think it's better to define `@use` as an empty, no-inline function at the point of a future work on assembler.

vladimirradosavljevic: Ah, didn't know about that, thanks for this info. Even though we have this issue with emitting an object files, do you think it is better to change that over simplicity in tests? Do you think we will use some of these tests for future work on assembler?

PavelKopyl: I guess we may use these files for asm parser testing, but ok let's leave them as is. It's not a bit deal to change when need.

akiramenai (approved):
LGTM

PavelKopyl (approved):
LGTM.

PR #760, opened 2025-01-21
[EVM] Refactor StackSlot std::variant -> LLVM style RTTI

akiramenai:


llvm/lib/Target/EVM/EVMStackSlotBuilder.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifndef LLVM_LIB_TARGET_EVM_EVMSTACKSLOTBUILDER_H
+#define LLVM_LIB_TARGET_EVM_EVMSTACKSLOTBUILDER_H
+
+#include "EVMStackModel.h"
+#include "llvm/Support/raw_ostream.h"
+
+namespace llvm {
+
+class StackSlotBuilder {
```
vladimirradosavljevic: Since this class is a storage and a builder of StackSlots, what do you think about moving this class to `EVMStackModel.h` and making `StackSlotBuilder` a data member of `EVMStackModel`, and access it only from `EVMStackModel`?

llvm/lib/Target/EVM/EVMStackSlotBuilder.h
```diff
@@ -0,0 +1,79 @@
...
+    auto [It, Inserted] = TemporaryStorage.try_emplace(Key, MI, Idx);
+    return &It->second;
+  }
+  // Junk is always the same slot.
+  static JunkSlot *getJunk() {
+    static JunkSlot Junk;
+    return &Junk;
+  }
+
+  void clear() {
```
vladimirradosavljevic: If we move this class to `EVMStackModel`, this won't be needed since it will be freed with `EVMStackModel` after stackification. If we don't move this, we could also remove it if we remove `SlotBuilder` data member from `EVMBPStackification` and make it local in `EVMBPStackification::runOnMachineFunction`.

llvm/lib/Target/EVM/EVMStackShuffler.h
```diff
@@ -396,50 +397,50 @@ class Shuffler {
...
+    if (auto *p = dyn_cast<TemporarySlot>(Slot))
+      return TemporarySlotMultiplicity.at(p);
+
+    assert(isa<JunkSlot>(Slot));
     return JunkSlotMultiplicity;
   }
 
 private:
-  std::map<FunctionCallReturnLabelSlot, int>
+  std::map<const FunctionCallReturnLabelSlot *, int>
```
vladimirradosavljevic, line 435: Maybe change all `std::map` with `DenseMap` in this class?

llvm/lib/Target/EVM/EVMStackShuffler.h
```diff
@@ -519,9 +518,9 @@ void createStackLayout(Stack &CurrentStack, Stack const &TargetStack,
     }
 
     void pushOrDupTarget(size_t Offset) {
-      auto const &targetSlot = targetStack[Offset];
+      const auto *targetSlot = targetStack[Offset];
       pushOrDupCallback(targetSlot);
-      currentStack.push_back(targetSlot);
+      currentStack.push_back(const_cast<StackSlot *>(targetSlot));
```
vladimirradosavljevic: If we remove const from `const auto *targetSlot`, we can remove this `const_cast`.

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -34,6 +35,39 @@ namespace llvm {
...
+  const SlotKind KindID;
+
+protected:
+  StackSlot(SlotKind KindID) : KindID(KindID) {}
+
+public:
+  virtual ~StackSlot() = default;
+
+  unsigned getSlotKind() const { return KindID; }
+  virtual bool isRematerializable() const { return false; };
```
vladimirradosavljevic: Maybe make this as a pure virtual method?

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -34,6 +35,39 @@ namespace llvm {
...
+  virtual std::string toString() const { return std::string(); };
```
vladimirradosavljevic: Maybe make this as a pure virtual method?

llvm/lib/Target/EVM/EVMStackDebug.h
```diff
@@ -28,7 +28,6 @@ namespace llvm {
 class EVMStackLayout;
 
 const Function *getCalledFunction(const MachineInstr &MI);
```
vladimirradosavljevic: This is no longer needed.

vladimirradosavljevic (commented):
Could you please run clang-format?

llvm/lib/Target/EVM/EVMStackSlotBuilder.h
akiramenai: Moved `StackSlotBuilder`, but I left making access only from `EVMStackModel` for future patches as it expands the scope.

llvm/lib/Target/EVM/EVMStackSlotBuilder.h
akiramenai: Done

llvm/lib/Target/EVM/EVMStackShuffler.h
akiramenai: I agree, but I plan to do it in the scope of `Shuffler` rework.

akiramenai:
`clang-format` is done.

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -77,8 +77,8 @@ findStackTooDeep(Stack const &Source, Stack const &Target) {
           Errors.emplace_back(EVMStackLayoutGenerator::StackTooDeep{
               I - 16, getVariableChoices(take_back(CurrentStack, I + 1))});
       },
-      [&](StackSlot const &Slot) {
-        if (isRematerializable(Slot))
+      [&](StackSlot const *Slot) {
```
PavelKopyl: In other places you moved `const` to the beginning of a declaration. Let's fix this one also.

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -35,138 +36,175 @@ namespace llvm {
...
+  virtual bool isRematerializable() const = 0;
+  virtual std::string toString() const = 0;
+};
+
 /// The following structs describe different kinds of stack slots.
 /// Each stack slot is equality- and less-than-comparable and
 /// specifies an attribute 'isRematerializable' that is true,
 /// if a slot of this kind always has a known value at compile time and
 /// therefore can safely be removed from the stack at any time and then
 /// regenerated later.
```
PavelKopyl: Please, update the description.

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -193,10 +231,26 @@ struct Operation {
 };
 
 class EVMStackModel {
+  MachineFunction &MF;
+  const LiveIntervals &LIS;
+  DenseMap<const MachineBasicBlock *, SmallVector<Operation>> OperationsMap;
+
+  // Storage for stack slots.
+  mutable DenseMap<APInt, LiteralSlot> LiteralStorage;
```
vladimirradosavljevic: In all these maps, values are stored by value, and during rehash we will run into use after free issue:
1. `getLiteralSlot` and store that pointer somewhere.
2. `getLiteralSlot` (rehash of map) and store that pointer somewhere. 1. return pointer will be dangling.

I suggest to use `unique_ptr` for storing values to overcome this issue.

vladimirradosavljevic (changes requested):
I ran tests, and I saw some issues:
1. Use-after-free issue (explained in the comment).
2. `tests/solidity/simple/loop/complex/1.sol` is failing with `Variable not found on stack`, where previously was failing with `too deep in the stack`.
3. `tests/solidity/simple/algorithm/floating_point_simulation/geometry/lines.sol` is failing with `too deep in the stack`, where previously was passing.

llvm/lib/Target/EVM/EVMStackShuffler.h
```diff
@@ -0,0 +1,540 @@
...
+      currentStack.push_back(targetSlot);
+    }
+  };
+
+  Shuffler<ShuffleOperations>::shuffle(CurrentStack, TargetStack, Swap,
+                                       PushOrDup, Pop);
+
+  assert(CurrentStack.size() == TargetStack.size());
+  for (unsigned I = 0; I < CurrentStack.size(); ++I) {
+    auto *Current = CurrentStack[I];
```
PavelKopyl: I think `Current` should be defined as a reference, otherwise we assign JunkSlot address to the local variable `Current` on the line 533.

PavelKopyl:
As we have switched to pointer semantics, I think it's reasonable to add `const` qualifier to `StackSlot` pointer definitions where this semantically required and also removing `auto`. At least this will simplify the code reading.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -357,14 +336,12 @@ void EVMStackifyCodeEmitter::createStackLayout(const Stack &TargetStack) {
             Emitter.emitDUP(static_cast<unsigned>(Depth + 1));
             return;
           }
-          if (!isRematerializable(Slot)) {
-            std::string VarName = SlotVariableName(Slot);
+          if (Slot->isRematerializable()) {
```
PavelKopyl: It should be `!Slot->isRematerializable()` :)

PavelKopyl: This caused new test failures.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -357,14 +336,12 @@ void EVMStackifyCodeEmitter::createStackLayout(const Stack &TargetStack) {
...
             std::string Msg =
-                ((VarName.empty() ? "slot " + stackSlotToString(Slot)
-                                  : Twine("variable ") + VarName) +
-                 " is " + std::to_string(Depth - 15) +
-                 " too deep in the stack " + stackToString(CurrentStack))
-                    .str();
+                (isa<VariableSlot>(Slot) ? "variable "
+                                         : "slot " + Slot->toString()) +
+                " is " + std::to_string(Depth - 15) +
+                " too deep in the stack " + stackToString(CurrentStack);
```
PavelKopyl: Please, replace
```
(isa<VariableSlot>(Slot) ? "variable "
                                         : "slot " + Slot->toString())
```
with 
`(isa<VariableSlot>(Slot) ? "variable "    
                                         : "slot ") + Slot->toString()`
otherwise variable name will not be printed.

akiramenai:
The same tests are failing now in the base branch and the PR.
So the patch is ready for another review iteration.

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -205,14 +259,54 @@ class EVMStackModel {
     return OperationsMap.at(MBB);
   }
 
+  // Get or create a requested stack slot.
+  StackSlot *getStackSlot(const MachineOperand &MO) const;
+  LiteralSlot *getLiteralSlot(const APInt &V) const {
+    if (LiteralStorage.count(V) == 0)
+      LiteralStorage[V] = std::make_unique<LiteralSlot>(V);
+    return LiteralStorage[V].get();
```
vladimirradosavljevic, line 267:
```suggestion
    auto [It, Inserted] = LiteralStorage.try_emplace(V, nullptr);
    if (Inserted)
      It->second = std::make_unique<Literal>(V);
    return It->second.get();
```

vladimirradosavljevic: This also applies to methods below.

vladimirradosavljevic (approved):
LGTM, thanks! Just a comment to reduce repeated hash lookups (we have up to 3 per call currently).

PavelKopyl (approved):
LGTM, thanks!

PavelKopyl:
Please, rebase also both stackifcation branches into the main.

llvm/lib/Target/EVM/EVMStackModel.h
akiramenai: I prefer to keep the original implementation even if optimizer doesn't manage to remove the second lookup. It's shorter and better readable.

PR #768, opened 2025-01-28
[EVM][LLD] relocate: remove wrong numeric limit

PavelKopyl:
When applying a relocation in `EVM::relocate`, the `val` is actually a sign-extended 64-bit value (see `TargetInfo::relocateAlloc`), so the check `val > std::numeric_limits<uint32_t>::max()` is false.

# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion

### DIFF
--- a/lld/ELF/Arch/EVM.cpp
+++ b/lld/ELF/Arch/EVM.cpp
@@ -60,9 +60,7 @@ RelExpr EVM::getRelExpr(RelType type, const Symbol &s,
 void EVM::relocate(uint8_t *loc, const Relocation &rel, uint64_t val) const {
   switch (rel.type) {
   case R_EVM_DATA: {
-    if (val > std::numeric_limits<uint32_t>::max())
-      llvm_unreachable("R_EVM_DATA: to big relocation value");
-    write32be(loc, val);
+    write32be(loc, static_cast<uint32_t>(val));
     break;
   }
   default:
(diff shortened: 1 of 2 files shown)


lld/ELF/Arch/EVM.cpp
```diff
@@ -60,9 +60,7 @@ RelExpr EVM::getRelExpr(RelType type, const Symbol &s,
 void EVM::relocate(uint8_t *loc, const Relocation &rel, uint64_t val) const {
   switch (rel.type) {
   case R_EVM_DATA: {
-    if (val > std::numeric_limits<uint32_t>::max())
```
vladimirradosavljevic, line 63: Should we add here a check whether val is signed or unsigned 32-bit value by calling `checkIntUInt`?

PavelKopyl: I guess, `checkIntUInt` will also fail if passing something like `0xffffffff9f1ebbf1` in `val`. At the same time, original symbol value in linker script is 0x9f1ebbf1.

vladimirradosavljevic: Is there any chance we are seeing this behaviour because we set that EVM is a 32-bit architecture, and `TargetInfo::relocateAlloc` sign extend these values?

vladimirradosavljevic (approved):
LGTM. I'm wondering whether we can avoid this change if we set that EVM is a 64-bit architecture?

PavelKopyl:
You are right. I'll add a ticket to change arch type.

akiramenai (approved):

akiramenai:
Please rebase `ef-stackification`.

PR #772, opened 2025-01-31
Remove TemporarySlot

akiramenai:


vladimirradosavljevic:
This should be NFC, right? I'm asking because this PR changes assembly in `tests/solidity/simple/algorithm/tree/treap.sol`.
Previously for `TemporarySlot` we had keys as a MI + idx, so we could distinguish between multiple definitions of a same register (e.g. phi that was removed). Now, if we call `getRegisterSlot` for such case, we will get the same pointer, thus we can't distinguish between multiple definitions of a same register. Is that safe to do? Could that be a reason why we are seeing changes in `tests/solidity/simple/algorithm/tree/treap.sol`?

akiramenai:
@vladimirradosavljevic The reason for the changes was different. The register defined by MI is not necessarily used in the basic block, so it doesn't always come with a Post stack and must be added manually. TemporarySlot is only used within instruction propagation logic, so the MI context of the definition is already set, making `(MI, idx)` redundant.

llvm/lib/Target/EVM/EVMStackModel.cpp
```diff
@@ -155,73 +147,47 @@ void EVMStackModel::createOperation(MachineInstr &MI,
...
   } break;
   }
-  // We don't need an assignment part of the instructions that do not write
-  // results.
-  if (!Input.empty() || !Output.empty())
-    Ops.emplace_back(Operation::Assignment, std::move(Input), std::move(Output),
-                     &MI);
+
+  // Skip for the instructions that do not write results.
+  if (!Input.empty() || !MI.getNumExplicitDefs())
```
vladimirradosavljevic: With this change, we will produce assignments for instructions that doesn't have outputs, so should this be `if (MI.getNumExplicitDefs())`?

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -255,12 +226,18 @@ class EVMStackModel {
   EVMStackModel(MachineFunction &MF, const LiveIntervals &LIS);
   Stack getFunctionParameters() const;
   Stack getInstrInput(const MachineInstr &MI) const;
-  Stack getInstrOutput(const MachineInstr &MI) const;
   Stack getReturnArguments(const MachineInstr &MI) const;
   const SmallVector<Operation> &
   getOperations(const MachineBasicBlock *MBB) const {
     return OperationsMap.at(MBB);
   }
+  SmallVector<StackSlot *>
```
vladimirradosavljevic, line 234: Use `Stack` instead of `SmallVector<StackSlot *>`?

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -255,12 +226,18 @@ class EVMStackModel {
...
+  getSlotsForInstructionDefs(const MachineInstr *MI) const {
+    SmallVector<StackSlot *> Defs;
```
vladimirradosavljevic, line 236: Likewise.

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -91,14 +91,14 @@ findStackTooDeep(Stack const &Source, Stack const &Target) {
...
-/// generated on the fly during shuffling.
+/// Returns the ideal stack to have before executing the MachineInstr \p MI
+/// s.t. shuffling to \p Post is cheap (excluding the input of the operation
+/// itself). If \p GenerateSlotOnTheFly returns true for a slot, this slot
+/// should not occur in the ideal stack, but rather be generated on the fly
+/// during shuffling.
 template <typename Callable>
-Stack createIdealLayout(const Stack &OperationOutput, const Stack &Post,
-                        Callable GenerateSlotOnTheFly) {
+Stack createIdealLayout(const SmallVector<StackSlot *> &OpDefs,
```
vladimirradosavljevic, line 100: Use `Stack` instead of `SmallVector<StackSlot *>`?

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -272,18 +272,23 @@ Stack EVMStackLayoutGenerator::propagateStackThroughOperation(
     return AggressiveStackCompression && Slot->isRematerializable();
   };
 
+  SmallVector<StackSlot *> OpDefs;
```
vladimirradosavljevic: Use `Stack` instead of `SmallVector<StackSlot *>`?

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -272,18 +272,23 @@ Stack EVMStackLayoutGenerator::propagateStackThroughOperation(
...
+  for (const auto &MO : Op.getMachineInstr()->defs())
+    OpDefs.push_back(StackModel.getRegisterSlot(MO.getReg()));
```
vladimirradosavljevic: We can call `StackModel.getSlotsForInstructionDefs` instead of doing this loop?

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -226,11 +226,9 @@ void EVMStackifyCodeEmitter::adjustStackForInst(const MachineInstr *MI,
   CurrentStack.erase(CurrentStack.end() - NumArgs, CurrentStack.end());
 
   // Push return values to CurrentStack.
-  unsigned Idx = 0;
-  for (const auto &MO : MI->defs()) {
-    assert(MO.isReg());
-    CurrentStack.push_back(StackModel.getTemporarySlot(MI, Idx++));
-  }
+  for (const auto &MO : MI->defs())
+    CurrentStack.push_back(StackModel.getRegisterSlot(MO.getReg()));
```
vladimirradosavljevic: We can call `StackModel.getSlotsForInstructionDefs` instead of doing this loop?

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -273,16 +271,17 @@ void EVMStackifyCodeEmitter::processAssign(const Operation &Assignment) {
   assert(Assignment.isAssignment());
   assert(Emitter.stackHeight() == CurrentStack.size());
 
+  const auto *MI = Assignment.getMachineInstr();
   // Invalidate occurrences of the assigned variables.
   for (auto *&CurrentSlot : CurrentStack)
-    if (const auto *VarSlot = dyn_cast<VariableSlot>(CurrentSlot))
-      if (is_contained(Assignment.getOutput(), VarSlot))
+    if (const auto *RegSlot = dyn_cast<RegisterSlot>(CurrentSlot))
+      if (MI->definesRegister(RegSlot->getReg()))
```
vladimirradosavljevic, line 277: Paranoia nit: As far as I saw, `definesRegister` also checks implicit defs, so maybe it is better to implement a method in `Operation` to only check explicit defs and use it in all places?

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -272,18 +272,23 @@ Stack EVMStackLayoutGenerator::propagateStackThroughOperation(
...
+
   // Determine the ideal permutation of the slots in ExitLayout that are not
   // operation outputs (and not to be generated on the fly), s.t. shuffling the
   // 'IdealStack + Operation.output' to ExitLayout is cheap.
-  Stack IdealStack =
-      createIdealLayout(Op.getOutput(), ExitStack, generateSlotOnTheFly);
+  Stack IdealStack = createIdealLayout(OpDefs, ExitStack, generateSlotOnTheFly);
 
-  // Make sure the resulting previous slots do not overlap with any assignmed
+#ifndef NDEBUG
```
vladimirradosavljevic, line 283: I don't think this is needed.

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -110,15 +110,15 @@ Stack createIdealLayout(const Stack &OperationOutput, const Stack &Post,
...
+    if (is_contained(OpDefs, Slot) || GenerateSlotOnTheFly(Slot))
       --PreOperationLayoutSize;
 
   // The symbolic layout directly after the operation has the form
   // PreviousSlot{0}, ..., PreviousSlot{n}, [output<0>], ..., [output<m>]
   LayoutT Layout;
   for (size_t Index = 0; Index < PreOperationLayoutSize; ++Index)
     Layout.emplace_back(PreviousSlot{Index});
-  Layout.append(OperationOutput.begin(), OperationOutput.end());
+  Layout.append(OpDefs.begin(), OpDefs.end());
```
vladimirradosavljevic:
```suggestion
  append_range(Layout, OpDefs);
```

llvm/lib/Target/EVM/EVMStackModel.h
akiramenai: They are aliases, true. But `Stack` implies additional semantic, here we just return a collection of arguments which shouldn't be associated with stack by itself, you are to put it to a stack.
Similarly, I didn't put `Stack` where it's not supposed to be used as stack.

llvm/lib/Target/EVM/EVMStackModel.cpp
akiramenai: fixed, thanks for noticing

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
akiramenai: refactored

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
akiramenai: refactored

llvm/lib/Target/EVM/EVMStackModel.h
vladimirradosavljevic: Makes sense. Thanks for explanation.

llvm/lib/Target/EVM/EVMStackModel.cpp
vladimirradosavljevic: Do we also need to check `!Input.empty()`? It seems to me that checking number of explicit defs is suffice, since `Input` size and number of explicit defs are tightly coupled.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
akiramenai: Addressed in https://github.com/matter-labs/era-compiler-llvm/pull/772/commits/164a006b5cfb8132dff8235449d3df29e8cb4e0f

llvm/lib/Target/EVM/EVMStackModel.cpp
akiramenai: Actually the condition seems to be always true except for control flow instructions.
@PavelKopyl do you know if it's supposed to be so?

akiramenai: Given that we haven't observed regressions despite the incorrect condition, I'll investigate it further. The fact that we use the same data for Input and Output seems suspicious. Since the condition is no worse than before, I suggest proceeding as is and addressing it in future PRs.

vladimirradosavljevic: Sounds good. Let's do that.

llvm/lib/Target/EVM/EVMStackLayoutGenerator.cpp
```diff
@@ -272,18 +272,20 @@ Stack EVMStackLayoutGenerator::propagateStackThroughOperation(
...
   // Determine the ideal permutation of the slots in ExitLayout that are not
   // operation outputs (and not to be generated on the fly), s.t. shuffling the
   // 'IdealStack + Operation.output' to ExitLayout is cheap.
-  Stack IdealStack =
-      createIdealLayout(Op.getOutput(), ExitStack, generateSlotOnTheFly);
+  Stack IdealStack = createIdealLayout(OpDefs, ExitStack, generateSlotOnTheFly);
 
-  // Make sure the resulting previous slots do not overlap with any assignmed
+  // Make sure the resulting previous slots do not overlap with any assigned
   // variables.
```
vladimirradosavljevic, line 285: Locally when I build in release mode, there are no unused errors, but it seems that in CI there are. Could you please add `#ifndef NDEBUG`. Sorry for the noise.

vladimirradosavljevic (approved):
LGTM. Thanks!

PavelKopyl (approved):
LGTM, thanks!

PR #784, opened 2025-02-18
[EVM] Remove stack layout from StackSolver

akiramenai:


llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -68,7 +59,7 @@ findStackTooDeep(const Stack &Source, const Stack &Target,
     return Result;
   };
 
-  ::createStackLayout(
+  ::calculateStack(
```
PavelKopyl: By the way, here we actually do not need global scope operator `::`, as the call is not ambiguous.

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -165,130 +158,155 @@ Stack createIdealLayout(const SmallVector<StackSlot *> &OpDefs,
...
-  return IdealLayout;
+
+  return BeforeInst;
 }
 
 } // end anonymous namespace
 
+/// Returns the number of operations required to transform stack \p Source to
+/// \p Target.
+size_t llvm::calculateStackTransformOpNum(Stack Source, Stack const &Target,
```
PavelKopyl: This function evaluates transformation costs, so I would name it `calculateStackTransformCost`. What do you think?

PavelKopyl (approved):
LGTM, thanks.

PR #786, opened 2025-02-26
[EVM] Replace Operation with MachineInstruction

akiramenai:


llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -202,75 +202,73 @@ void EVMStackSolver::run() {
...
+  // MI defs, s.t. shuffling the 'BeforeMI + MIDefs' to AfterMI is cheap.
+  Stack BeforeMI = calculateStackBeforeInst(MIDefs, AfterMI, CompressStack,
+                                            StackModel.stackDepthLimit());
 
 #ifndef NDEBUG
   // Make sure the resulting previous slots do not overlap with any assigned
   // variables.
-  if (Op.isAssignment())
-    for (auto *StackSlot : BeforeInst)
+  if (isConstCopyOrLinkerMI(MI))
```
PavelKopyl: Could you, please clarify why do we need `isConstCopyOrLinkerMI(MI)` constrain? Before the change, `Op.isAssignment()` was created for each MI.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -417,46 +407,42 @@ void EVMStackifyCodeEmitter::run() {
...
 
     // Might set some slots to junk, if not required by the block.
-    CurrentStack = StackModel.getMBBEntryStack(Block);
-    Emitter.enterMBB(Block, CurrentStack.size());
+    CurrentStack = StackModel.getMBBEntryStack(MBB);
+    Emitter.enterMBB(MBB, CurrentStack.size());
+
+    SmallVector<const MachineInstr *> MIs;
+    for (const auto &MI : StackModel.instructionsToProcess(MBB))
+      MIs.push_back(&MI);
```
PavelKopyl, line 422: Could you, please clarify why do we first need to put instructions into the vector `MIs` and then iterate over it, instead of just calling `emitMIEntryStack(MI)` directly?

llvm/lib/Target/EVM/EVMStackSolver.cpp
akiramenai: Removed.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
akiramenai: Added an explanation three lines above.

PavelKopyl (approved):
LGTM, thanks!

PR #801, opened 2025-04-10
[EVM][BPS] Fix EVMStackSolver::compressStack to not removing inaccess…

akiramenai:
…ible copies


akiramenai:
Now a part of #806

Issue #799, opened 2025-04-09
[EVM] Hide optimizations that remove external calls behind a parameter


hedgar2017:
There are a lot of assumptions in EVM tooling about static calls kept alive even if their result is unused.

Foundry tests do not work at the moment, as test statements are implemented as static calls to magic addresses.

Whereas it makes sense for EraVM where EVM tooling doesn't work, we must disable this optimization by default and hide them behind a parameter, such as `-removed-unused-static-calls`.

akiramenai:
@hedgar2017 could tell more about these calls? May I assume that ideally we want to have them eliminated in production but kept for tests? How Foundry users work with this testing functionality now, is it removed from the code they actually deploy? If it is, could you tell us how they achieve so?
Perhaps we should think about the whole workflow first. And my fear is if we keep calls alive by default they might be deployed.
Also does it apply to users who does not use Foundry?

hedgar2017:
@akiramenai well, solc has never ever removed these calls as it would've broken their gas and a lot of other assumptions.

Thus, static calls have been always deployed by solc, even in production, so keeping them in tests and removing in -O3 is not a solution. The only safe way is to make the @vladimirradosavljevic's optimization optional, at least until we've started working on advanced optimizations adjacent to smart contract security.

Upd: got a consultation on smart contract security and EVM equivalence:

![Image](https://github.com/user-attachments/assets/c8eaa3ec-bdea-4449-ac3d-4c0556b60c61)

akiramenai:
Revert is a solid argument. But the overall design makes me cry.

Anyway, thank you kindly for education. We'll make it optional.
We'll keep inventing `#ifdef NDEBUG` for Solidity for other day.

akiramenai:
@hedgar2017 could you kindly give me a reproducer?

```
target datalayout = "E-p:256:256-i256:256:256-S256-a:256:256"
target triple = "evm"

define i256 @staticcall(i256 %gas, i256 %addr, ptr addrspace(1) %arg_off, i256 %arg_size, ptr addrspace(1) %ret_off, i256 %ret_size) nounwind {
  %ret = call i256 @llvm.evm.staticcall(i256 %gas, i256 %addr, ptr addrspace(1) %arg_off, i256 %arg_size, ptr addrspace(1) %ret_off, i256 %ret_size)
  ret i256 42
}

[...]

PR #803, opened 2025-04-11
[EVM] Update the intrinsics' attributes

### LINKED ISSUE #807: Refine intrinsics' attributes
**extcodecopy**
This should have 'read of inaccessible' memory along with read/write memory accessed through
its pointer-typed arguments. Since we can't express that precisely here, we represent
it as read/write access to arbitrary memory.

**log** & **call**
Log intrinsics should have 'read/write of inaccessible' memory along with read memory
accessed through its pointer-typed arguments. Since we can't express that precisely
here, we represent it as read/write access to arbitrary memory.
Note: Attempting to mark them with both IntrReadMem and IntrHasSideEffects won't work,
because IntrHasSideEffects does not override IntrReadMem. As a result, intrinsics will
lack the 'HasSideEffects' flag entirely, making them subject to DCE.
https://discourse.llvm.org/t/intrinsic-with-sideeffect-is-optimized-out/66053?page=2

**return/revert**
These should ideally have the IntrReadMem attribute, but doing so causes DSE
to incorrectly eliminate necessary stores:
```
arget datalayout = "E-p:256:256-i256:256:256-S256-a:256:256"
target triple = "evm-unknown-unknown"

; Function Attrs: mustprogress nofree nosync nounwind willreturn memory(none)
declare i256 @llvm.evm.calldatasize() #0

; Function Attrs: nofree noreturn nounwind memory(argmem: read)
declare void @llvm.evm.return(ptr addrspace(1), i256) #1

; Function Attrs: nofree noreturn null_pointer_is_valid memory(readwrite, inaccessiblemem: none)
define void @__entry() local_unnamed_addr #2 {
entry:
  %calldatasize = tail call i256 @llvm.evm.calldatasize()
  br label %if_join22

if_join22:                                        ; preds = %switch_case_branch_2_block
  store i256 1, ptr addrspace(5) null, align 4294967296
  %comparison_result.i = icmp ugt i256 %calldatasize, 4
  br i1 %comparison_result.i, label %fun_g.exit, label %if_join.i

if_join.i:                                        ; preds = %if_join22
  tail call void @llvm.evm.return(ptr addrspace(1) noalias nocapture nofree noundef nonnull align 32 null, i256 1)
  unreachable

fun_g.exit:                                       ; preds = %if_join22
  store i256 2, ptr addrspace(5) null, align 4294967296
  tail call void @llvm.evm.return(ptr addrspace(1) noalias nocapture nofree noundef nonnull align 32 null, i256 2)
  unreachable
}

attributes #0 = { mustprogress nofree nosync nounwind willreturn memory(none) }
attributes #1 = { nofree noreturn nounwind memory(argmem: read) }
attributes #2 = { nofree noreturn null_pointer_is_valid memory(readwrite, inaccessiblemem: none) }
```

`store i256 1, ptr addrspace(5) null, align 4294967296` will be removed which break the programm logic.

akiramenai:
Resolves https://github.com/matter-labs/era-compiler-llvm/issues/807.
Also adds more guardrails for https://github.com/matter-labs/era-compiler-llvm/issues/799.

### DIFF
--- a/llvm/include/llvm/IR/IntrinsicsEVM.td
+++ b/llvm/include/llvm/IR/IntrinsicsEVM.td
@@ -23,6 +23,37 @@ def AS {
 
 let TargetPrefix = "evm" in {
 
+// Ethereum Virtual Machine (EVM) opcodes and corresponding intrinsics can be grouped
+// by the runtime effects i.e., how their behavior or outputs are affected by state
+// changes during a transaction:
+//
+//   - 'Readnone'
+//     These do not read or write any persistent world-state once execution begins.
+//     Their results depend only on provided inputs or fixed environment data
+//     (constant during the transaction). They have no lasting side effects and
+//     return the same result given the same inputs.
+//
+//   - 'Volatile (State-Dependent)'
+//     These read data from the world state or execution environment that may change
+//     within the same transaction (for example, due to prior storage writes, value
+//     transfers, or external calls). Re-invoking them after a state mutation can
+//     yield a different result. They themselves do not permanently modify state.
+//     Such intrinsics have 'IntrInaccessibleMemOnly' attribute, or
+//     'IntrReadMem, IntrInaccessibleMemOnly' attributes to model dependency from
+//     the State.
+//     NOTE: we assume (to be confirmed) that the heap is not part of the State;
+//     therefore, it cannot be modified in a volatile manner.
+//
+//   - 'Side-Effecting (State-Changing)'
+//     These opcodes cause persistent changes in world state - writing to storage,
+//     emitting logs, creating or destroying contracts, or transferring Ether.
+//     Their execution produces lasting side effects recorded on-chain (storage
+//     updates, logs, new accounts, etc.).
+//     Such intrinsics have empty memory attributes list which means they can
+//     arbitrary change the State.
+
+// TODO: Further improvements of intrinsics' attributes, #807
+
 // Arithmetical operations.
 def int_evm_div
   : Intrinsic<[llvm_i256_ty], [llvm_i256_ty, llvm_i256_ty], [IntrNoMem, IntrWillReturn]>;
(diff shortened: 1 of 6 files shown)


llvm/include/llvm/IR/IntrinsicsEVM.td
```diff
@@ -136,23 +136,28 @@ def int_evm_basefee : Intrinsic<[llvm_i256_ty], [], [IntrNoMem, IntrWillReturn]>
 def int_evm_blobbasefee : Intrinsic<[llvm_i256_ty], [], [IntrNoMem, IntrWillReturn]>;
 
 // Logging.
-def int_evm_log0
-  : Intrinsic<[], [LLVMQualPointerType<AS.HEAP>, llvm_i256_ty], [IntrReadMem]>;
+def int_evm_log0 : Intrinsic<[], [LLVMQualPointerType<AS.HEAP>, llvm_i256_ty],
+                             [IntrReadMem, IntrHasSideEffects]>;
```
PavelKopyl: Probably we could specify these intrinsics further by adding: `IntrArgMemOnly` and `ReadOnly<ArgIndex<0>>`.
What do you think?

llvm/include/llvm/IR/IntrinsicsEVM.td
```diff
@@ -129,62 +165,73 @@ def int_evm_gaslimit : Intrinsic<[llvm_i256_ty], [], [IntrNoMem, IntrWillReturn]
...
                LLVMQualPointerType<AS.HEAP>, llvm_i256_ty,
-               LLVMQualPointerType<AS.HEAP>, llvm_i256_ty], []>;
+               LLVMQualPointerType<AS.HEAP>, llvm_i256_ty], [IntrWillReturn]>;
 
 def int_evm_delegatecall
   : Intrinsic<[llvm_i256_ty],
               [llvm_i256_ty, llvm_i256_ty,
                LLVMQualPointerType<AS.HEAP>, llvm_i256_ty,
-               LLVMQualPointerType<AS.HEAP>, llvm_i256_ty], []>;
+               LLVMQualPointerType<AS.HEAP>, llvm_i256_ty], [IntrWillReturn]>;
```
hedgar2017, line 222: From ChatGPT:
```
Recommendation
Do not tag generic EVM intrinsics with IntrWillReturn.
Limit the flag to helpers that are provably unable to trigger OOG (e.g. pseudo-intrinsics eliminated before codegen) or accept the optimisation opportunities you lose by keeping the safer default.
```
TLDR: we cannot use `willreturn` as anything can fail with out-of-gas on EVM.

hedgar2017 (changes requested):

hedgar2017 (approved):

hedgar2017:
Let's keep it this way to be on the safe side. We still have some strange failures in some projects.
Later we can reassess everything once again when we are back to add more EVM-specific customization.

llvm/include/llvm/IR/IntrinsicsEVM.td
hedgar2017: However, subsequent queries showed that it's not that straighforward, maybe ChatGPT is hallucinating.
Let me know if I shall bring this attribute back.

PR #820, opened 2025-05-23
[EVM] Disable tail duplication optimization

vladimirradosavljevic:
In some cases, tail duplication can combine a loop’s header with its latch. When this happens, MachineLoopInfo may fail to recognize the loop during stackification. This is a problem because, in EVMStackSolver::runPropagation, we add an empty exit stack for loop latches during the first run (if the header hasn’t been visited yet) to prevent infinite loops. If the loop isn’t detected, we will run into an infinite loop situation.

### DIFF
--- a/llvm/lib/Target/EVM/EVMTargetMachine.cpp
+++ b/llvm/lib/Target/EVM/EVMTargetMachine.cpp
@@ -164,7 +164,9 @@ namespace {
 class EVMPassConfig final : public TargetPassConfig {
 public:
   EVMPassConfig(EVMTargetMachine &TM, PassManagerBase &PM)
-      : TargetPassConfig(TM, PM) {}
+      : TargetPassConfig(TM, PM) {
+    disablePass(&EarlyTailDuplicateID);
+  }
 
   EVMTargetMachine &getEVMTargetMachine() const {
     return getTM<EVMTargetMachine>();
@@ -232,6 +234,7 @@ void EVMPassConfig::addPostRegAlloc() {
   disablePass(&LiveDebugValuesID);
   disablePass(&PatchableFunctionID);
   disablePass(&ShrinkWrapID);
+  disablePass(&TailDuplicateID);
 
   // TODO: This pass is disabled in WebAssembly, as it hurts code size because
   // it can generate irreducible control flow. Check if this also true for EVM?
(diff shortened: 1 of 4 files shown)


PavelKopyl (approved):
LGTM.

akiramenai:
Could you refine this part of the commit message:

> This could be a problem, since during
EVMStackSolver::runPropagation for loop latches we are
adding empty exit stack just to prevent this from
happening.

Should we still add empty exit stack after this patch?

vladimirradosavljevic:
> Could you refine this part of the commit message:
> 
> > This could be a problem, since during
> > EVMStackSolver::runPropagation for loop latches we are
> > adding empty exit stack just to prevent this from
> > happening.
> 
> Should we still add empty exit stack after this patch?

Yes we should, since it is a safeguard to prevent infinite loop from happening. Please find the updated commits message:
_In some cases, tail duplication can combine a loop’s header with its latch. When this happens, MachineLoopInfo may fail to recognize the loop during stackification. This is a problem because, in EVMStackSolver::runPropagation, we add an empty exit stack for loop latches during the first run (if the header hasn’t been visited yet) to prevent infinite loops. If the loop isn’t detected, we will run into an infinite loop situation._

akiramenai:
I now see what problem the patch addresses. Could we also document why this particular approach was chosen? Among the alternatives, it seems we could have:
* taught tail duplication to avoid breaking loops,
* extended MachineLoopInfo to recognize tail-duplicated loops, or
* adjusted runPropagation logic to detect such loops.

It would be helpful to clarify why this path was preferred over the others.
Also prettifying the commit message with an AI is always welcome.

vladimirradosavljevic:
> It would be helpful to clarify why this path was preferred over the others.

This is not the best solution, but rather workaround. I have [created](https://github.com/matter-labs/era-compiler-llvm/issues/822) the ticket for what I think is the best solution (to revert TailDuplicate optimization for loops). Please note that other backends, (webassembly and amdgpu) have various loop transformations (e.g `FixIrreducibleControlFlow`).  

>it seems we could have:
>
>taught tail duplication to avoid breaking loops,

This would require a lot of local changes (a lot of maintenance burden) , since `MachineLoopInfo` is not used in that pass.
>extended MachineLoopInfo to recognize tail-duplicated loops, or

Since LoopInfo base class is shared between IR and MIR analysis, I don't think this is a good idea, since we might break a lot of common passes by doing this.
>adjusted runPropagation logic to detect such loops.

I'm not sure how this would work since in this case we would have latch and header merged,  and stackification is expecting to have backedges between these two BBs. It seems to me that this would require a lot of changes in `EVMStackSolver::runPropagation` to support this case.

akiramenai (approved):

akiramenai:
Still worth mentioning the issue in the commit message or in the code, but that’s just a nit at this point.

PR #823, opened 2025-05-26
ci: run LNT benchmarks only for internal PRs

antonbaliasnikov:
## What?

Run LNT benchmarks only for internal PRs.

## Why?

To protect uploading secret key to improve security.

### DIFF
--- a/.github/workflows/lnt-benchmarks.yml
+++ b/.github/workflows/lnt-benchmarks.yml
@@ -14,6 +14,8 @@ concurrency:
 jobs:
 
   lnt-benchmarks:
+    # Run only for matter-labs pull requests, ignore for forks
+    if: ${{ github.event.pull_request.head.repo.full_name == github.repository || github.event_name == 'push' }}
     uses: matter-labs/era-compiler-ci/.github/workflows/lnt.yml@v1
     secrets: inherit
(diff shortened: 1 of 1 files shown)


.github/workflows/lnt-benchmarks.yml
```diff
@@ -6,6 +6,8 @@ on:
       - main
   workflow_dispatch:
   pull_request:
+    branches:
+      - main
```
hedgar2017: Should we allow running it between arbitrary branches?

hedgar2017 (changes requested):

.github/workflows/lnt-benchmarks.yml
antonbaliasnikov: I'm not sure. I thought if PR is not open to be merged to `main`, then we don't need so much info about performance yet. And the comparison will anyway always be VS the `main`.

antonbaliasnikov: But we can allow it if required, that's ok.

antonbaliasnikov: Allowed for now.

hedgar2017 (approved):

PR #828, opened 2025-05-29
[EVM] Add support for spills and reloads

vladimirradosavljevic:
In order to add spills and reloads support, following has implemented:

1. The stackification process is now iterative, with a configurable
   maximum number of iterations. Spill candidates are identified
   and spilled in subsequent iterations until stack-too-deep errors
   are resolved or the iteration limit is reached. When they are resolved
   only then we are emitting instructions.
2. The heuristic for selecting which registers to spill and reload
   is based on LLVM's weight calculation. When an unreachable slot
   is encountered, all registers from that point to the end are
   considered, and the register with the lowest weight is chosen.
   This approach has yielded the best results in our benchmarks so far.
3. To indicate that a register needs to be spilled, an IsSpill data member
   has been added to RegisterSlot, along with corresponding getter and
   setter methods.
4. For all spills and reloads, new PUSH_FRAME instruction is introduced which
   accepts frame index. This is needed because we need to allocate stack size
   for each function after stackification is done.
5. To minimize the number of stack slots, StackSlotColoring is invoked
   after stackification. This pass relies on LLVM analyses—also used
   during register allocation—to assign stack slots with the live
   intervals of their corresponding registers.
6. EVMFinalizeStackFrames is introduced to calculate stack sizes for each function
   and to replace frame indices with concrete stack offsets.

3 new options are added to control this support:
1. -evm-max-spill-iterations=uint32_t
   This option controls how many iterations we want to perform during stackification.
   If 0, spills and reloads support is disabled.
2. -evm-stack-region-size=uint64_t
   This option sets the size of the allocated stack region for the whole module.
   In case allocated stack region is less than the actual total size, error
   will be produced.
3. -evm-stack-region-offset=uint64_t
   This options sets the offset where the stack region starts.


vladimirradosavljevic:
Even though this is a WIP patch, comments are welcome.

abinavpp:
Only reviewed EVMStackifyCodeEmitter::CodeEmitter::emitReload and emitSpill and how EVM::PUSH_FRAME is lowered in EVMFinalizeStackFrames::finalizeStackFrames. All good!

Question: I just forced the round up to multiple of 32 for the spill area size in solc. Saw the `MF.getFrameInfo().CreateSpillStackObject(32, Align(32))`, but just double checking: Is the offset calculated in the PUSH_FRAME lowering (`CurrentStackRegionOffset + MFI.getObjectOffset(FrameIndexOp.getIndex())`) always a multiple of 32? Just concerned if the mload/mstore near the end of spill area might go beyond the word boundary.

vladimirradosavljevic:
> Only reviewed EVMStackifyCodeEmitter::CodeEmitter::emitReload and emitSpill and how EVM::PUSH_FRAME is lowered in EVMFinalizeStackFrames::finalizeStackFrames. All good!
> 
> Question: I just forced the round up to multiple of 32 for the spill area size in solc. Saw the `MF.getFrameInfo().CreateSpillStackObject(32, Align(32))`, but just double checking: Is the offset calculated in the PUSH_FRAME lowering (`CurrentStackRegionOffset + MFI.getObjectOffset(FrameIndexOp.getIndex())`) always a multiple of 32? Just concerned if the mload/mstore near the end of spill area might go beyond the word boundary.

Thanks for the comment!
As long as the `StackRegionOffset` is the multiple of 32, mload/mstore will do the word load/store. Good point, I will add an assert to always check that `StackRegionOffset` is the multiple of 32.
Please note that `MF.getFrameInfo().CreateSpillStackObject(32, Align(32))` will create slots of 32bytes.

llvm/lib/Target/EVM/EVMMachineFunctionInfo.h
```diff
@@ -36,6 +38,8 @@ struct EVMMachineFunctionInfo final : public yaml::MachineFunctionInfo {
 template <> struct MappingTraits<EVMMachineFunctionInfo> {
   static void mapping(IO &YamlIO, EVMMachineFunctionInfo &MFI) {
     YamlIO.mapOptional("isStackified", MFI.IsStackified, false);
+    YamlIO.mapOptional("NumberOfParameters", MFI.NumberOfParameters, 0);
```
akiramenai: Nit: The previous member 'isStackified' starts with the lowercase, should we name the new ones the same way?

llvm/lib/Target/EVM/EVMMachineFunctionInfo.cpp
```diff
@@ -26,7 +26,9 @@ yaml::EVMMachineFunctionInfo::~EVMMachineFunctionInfo() = default;
 
 yaml::EVMMachineFunctionInfo::EVMMachineFunctionInfo(
     const llvm::EVMMachineFunctionInfo &MFI)
-    : IsStackified(MFI.getIsStackified()) {}
```
akiramenai, line 29: Could we test this patch?

llvm/lib/Target/EVM/EVMFinalizeStackFrames.cpp
```diff
@@ -0,0 +1,182 @@
...
+  LLVM_DEBUG({ dbgs() << "Total stack size: " << TotalStackSize << "\n"; });
+
+  // Check if it is valid to replace frame indices.
+  if (TotalStackSize > 0) {
+    if (TotalStackSize > StackRegionSize)
+      report_fatal_error("Total stack size: " + Twine(TotalStackSize) +
+                         " is larger than the allocated stack region size: " +
+                         Twine(StackRegionSize));
+
+    if (StackRegionOffset == std::numeric_limits<uint64_t>::max())
```
akiramenai: Could we move this and below diagnostics to the beginning of the function and check together with StackRegionSize? I mean if StackRegionSize set, we expect StackRegionOffset to be correctly set as well and error early if it is not.

llvm/lib/Target/EVM/EVMFinalizeStackFrames.cpp
```diff
@@ -0,0 +1,182 @@
...
+      dbgs() << "Stack size for function " << MF->getName()
+             << " is: " << StackSize
+             << " and starting offset is: " << StackRegionStart << "\n";
+    });
+  }
+  LLVM_DEBUG({ dbgs() << "Total stack size: " << TotalStackSize << "\n"; });
+
+  // Check if it is valid to replace frame indices.
+  if (TotalStackSize > 0) {
+    if (TotalStackSize > StackRegionSize)
```
akiramenai, line 174: Could we also add a warning if StackRegionSize > TotalStackSize?

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -201,54 +259,155 @@ llvm::calculateStackTransformCost(Stack Source, const Stack &Target,
...
+            // Swap.
+            [&](unsigned I) {
+              assert(I > 0 && I < From.size());
+              if (I > StackDepthLimit)
+                UnreachableSlots.push_back({From, From.size() - I - 1});
+            },
+            // Push or dup.
+            [&](const StackSlot *Slot) {
+              // Dup the slot, if already on stack and reachable.
+              auto SlotIt = llvm::find(llvm::reverse(From), Slot);
```
akiramenai, line 302: Maybe we can do an early exit on rematerializable slots instead of in/else.

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -201,54 +259,155 @@ llvm::calculateStackTransformCost(Stack Source, const Stack &Target,
...
+    auto [BranchTy, TBB, FBB, BrInsts, Condition] = getBranchInfo(&MBB);
+
+    // Check the transformation for the exit stack. This mimics logic from
+    // the EVMStackifyCodeEmitter::run() method, as in some cases this is
+    // not needed.
+    switch (BranchTy) {
+    case EVMInstrInfo::BT_Uncond:
+    case EVMInstrInfo::BT_NoBranch:
+      if (!MBB.succ_empty())
+        checkStackTransformation(CurrentStack,
```
akiramenai, line 334: Could you please give some details why we need the special handling here?

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -195,6 +196,65 @@ void EVMStackifyCodeEmitter::CodeEmitter::emitCondJump(
...
+          .addFrameIndex(getStackSlot(Reg));
+  verify(NewMI);
+  NewMI = BuildMI(*CurMBB, CurMBB->end(), DebugLoc(), TII->get(EVM::MLOAD_S));
+  NewMI->setAsmPrinterFlag(MachineInstr::ReloadReuse);
+  verify(NewMI);
+}
+
+void EVMStackifyCodeEmitter::CodeEmitter::emitSpill(Register Reg,
+                                                    unsigned DupIdx) {
+  // Reduce height if we are not going to duplicate the register.
```
akiramenai: I'd squash lines 212-230 into a single if/else.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -266,6 +326,58 @@ void EVMStackifyCodeEmitter::emitMI(const MachineInstr &MI) {
...
+  if (find_if(Defs, [](const StackSlot *Slot) {
+        return isa<RegisterSlot>(Slot) && Slot->isRematerializable();
+      }) == Defs.end())
+    return;
+
+  // In case of a single definition, we can remove it from the stack
+  // if it is not used after spill.
+  if (Defs.size() == 1) {
+    // Find the first instruction from which we can get entry stack.
+    while (Start != MBB.end() && StackModel.skipMI(*Start))
```
akiramenai, line 342: StackModel.instructionsToProcess() ?

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -266,6 +326,58 @@ void EVMStackifyCodeEmitter::emitMI(const MachineInstr &MI) {
...
+      ++Start;
+
+    // Find the next target stack, as we need to check if the register
+    // is used after spill.
+    const Stack &NextTargetStack =
+        Start != MBB.end() && !EVMInstrInfo::isStack(&*Start)
+            ? StackModel.getInstEntryStack(&*Start)
+            : StackModel.getMBBExitStack(&MBB);
+    const auto *RegSlot = dyn_cast<RegisterSlot>(Defs[0]);
```
akiramenai: cast<>?

llvm/test/CodeGen/EVM/stack-too-deep-1.ll
```diff
@@ -1,29 +1,488 @@
...
-;  [ %82 %85 %7 %78 %70 %11 %64 %17 %19 %54 %46 %5 %2 %40 %9 %82 %85 %29 %33 %72 ]
-;
-;  [ %82 %85 %7 %78 %70 %11 %64 %17 %19 %54 %46 %5 %2 %40 %9 %82 %85 %29 %33 %72 ]
-
-; CHECK:      EVMStackSolver cannot transform
-; CHECK-SAME: [ %82 %85 %7 %78 %70 %11 %64 %17 %19 %54 %46 %5 %2 %40 %9 %29 %33 ]
-; CHECK-SAME: to
-; CHECK-SAME: [ %82 %85 %7 %78 %70 %11 %64 %17 %19 %54 %46 %5 %2 %40 %9 %82 %85 %29 %33 7 21 ]
-; CHECK-SAME: : stack too deep.
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
```
akiramenai: For this test and tests below can we add some checks for debug prints to be sure StackSolver still catch and generate spills where we expect? I'd also avoid checking the whole code emitted because it would be hard to analize changes in future. May be we can just check debug prints and leave a comment that we no longer expect stack-too-deep on these tests.

llvm/test/CodeGen/EVM/bps-spills-1.mir
```diff
@@ -0,0 +1,1072 @@
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
```
akiramenai, line 1: Could we also add tests for fatal errors added in this patch?

llvm/lib/Target/EVM/EVMStackModel.h
```diff
@@ -82,16 +83,22 @@ class LiteralSlot final : public StackSlot {
...
   Register Reg;
+  bool IsSpill = false;
 
 public:
   explicit RegisterSlot(const Register &R) : StackSlot(SK_Register), Reg(R) {}
   const Register &getReg() const { return Reg; }
+  void setIsSpill(bool Spill = true) { IsSpill = Spill; }
 
-  bool isRematerializable() const override { return false; }
+  bool isRematerializable() const override { return IsSpill; }
```
akiramenai: How about adding a dedicated method isSpill()? I'm not sure hidding it under isRematerializable() is clear enough.

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -358,6 +358,14 @@ void EVMStackSolver::run() {
     if (UnreachableSlots.empty())
       break;
 
+    // For recursive functions we can't use spills to fix the stack too deep
+    // errors, as we are using memory to spill and not real stack. Report an
+    // error if this function is recursive.
+    if (MF.getFunction().hasFnAttribute("evm-recursive"))
```
akiramenai: Could we add a test to check this error?

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -358,13 +358,26 @@ void EVMStackSolver::run() {
...
+        report_fatal_error(
+            "Stackification failed for '" + MF.getName() +
+            "' function. It is recursive and has stack too deep errors.");
+
+      LLVM_DEBUG({
+        dbgs()
+            << "EVMStackSolver: force CompressStack for recursive function.\n";
+      });
+
+      // TODO: This is a little bit agressive, since we can detect MBBs where
```
akiramenai, line 383: propagetThroughMBB() already does compression if detects stack-too-deep during propagation.
Could you give more details about the case here? Which transformation is not possible before the compression of the whole function and how the patch fixes it? Is this about MBB's entry stack ->first MI's entry stack tranformation which we cannot catch during propagation?

akiramenai (changes requested):
Can't we use `norecurse` LLVM attribute? In addition to carrying semantic similar to `!evm-recursive`, it seems it enables some optimizations.

llvm/lib/Target/EVM/EVMMachineFunctionInfo.h
vladimirradosavljevic: Will do.

llvm/lib/Target/EVM/EVMMachineFunctionInfo.cpp
vladimirradosavljevic: This patch is tested in the spills and reloads support in .mir tests. You can see:
```
machineFunctionInfo:
  isStackified:    false
  NumberOfParameters: 4
  HasPushDeployAddress: false
```
If you prefer, I can add standalone tests for this patch. Please let me know.

llvm/lib/Target/EVM/EVMFinalizeStackFrames.cpp
vladimirradosavljevic: Will do.

llvm/lib/Target/EVM/EVMFinalizeStackFrames.cpp
vladimirradosavljevic: Will do.

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: Makes sense, will do.

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: Before the switch, I have added following comment:
```
// Check the transformation for the exit stack. This mimics logic from
// the EVMStackifyCodeEmitter::run() method, as in some cases this is
// not needed.
```
As it is written in the comment, I implemented the same as in `EVMStackifyCodeEmitter::run()`, so we don't do additional checks, and maybe get false positives.

vladimirradosavljevic:
> Can't we use `norecurse` LLVM attribute? In addition to carrying semantic similar to `!evm-recursive`, it seems it enables some optimizations.

I was investigating this, and I found 2 issues with `norecurse`:
1. If function contains intrinsic that doesn't have `NoCallback` [attribute](https://github.com/matter-labs/era-compiler-llvm/blob/7212634e4f4d3dca4d84c6e544437345bdb006c8/llvm/lib/Transforms/IPO/FunctionAttrs.cpp#L1701), it won't mark it with `NoRecurse`. This is not a problem, as we can add this attribute to our intrinsics, but
2. In -O0 mode, this attribute is [not added](https://github.com/matter-labs/era-compiler-llvm/blob/7212634e4f4d3dca4d84c6e544437345bdb006c8/llvm/lib/Transforms/IPO/FunctionAttrs.cpp#L1809), and that could be a problem, as we might want to do spills and reloads in that case.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
vladimirradosavljevic: Will do.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
vladimirradosavljevic: I'm not sure how we can use that, since in some cases we are not starting from the beginning, but rather from the instruction that can be anywhere in the BB.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
vladimirradosavljevic: Will do.

llvm/test/CodeGen/EVM/stack-too-deep-1.ll
vladimirradosavljevic: You mean debug prints where we have spills? I will try to do that.

llvm/test/CodeGen/EVM/bps-spills-1.mir
vladimirradosavljevic: For all fatal errors, or? I can add for iteration count, but I don't think I can do that if there are no spillable regs found. I added that fatal error in case that happens as a paranoia check.

llvm/lib/Target/EVM/EVMStackModel.h
vladimirradosavljevic: I have implemented that in the beginning, and saw that spillable registers are handled as same as remat, so I gave up on doing that, since it adds unnecessary changes. I see one benefit if in a future we want to move instruction remat from `EVMSingleUseExpression` to here. If you want, I can add a dedicated method for spills now.

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: Do you have any suggestion in doing that? Since I don't have any real world example, the only way I see we can do this is to take existing test and to make function recursive. Wdyt?

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: Exactly, problem happens in MBB's entry stack ->first MI's entry stack transformation. In case you expect this to be frequent case (recursive function with stack-too-deep issues), I can implement this TODO, but currently I don't think it is worth it, as I saw this issue in tests of some project, not the source code.

llvm/lib/Target/EVM/EVMStackModel.h
akiramenai: It introduces terminological confusion. Rematerializable means you can create it from thin air, while here you need to keep the value in memory.
The alternative is to rename the function.

llvm/lib/Target/EVM/EVMMachineFunctionInfo.cpp
akiramenai: I think a standalone test is beneficial.

llvm/lib/Target/EVM/EVMStackSolver.cpp
akiramenai: Do I understand correctly that we have a single successor, yet the exit stack of the current basic block doesn’t match the entry stack of that successor? Can we pinpoint why this happens and explain it directly, rather than simply referring back to the code emitter, which exhibits the same discrepancy?

llvm/test/CodeGen/EVM/bps-spills-1.mir
akiramenai: I meant
1. To test that if the allocated region for spills is insufficient we emit fatal error.
2. To test if the region is greater than needed we emit a warning.
3. To test that if stack size is set, stack offset is set as well.
4. To test for stack offset non divisible by 32.

llvm/lib/Target/EVM/EVMStackSolver.cpp
akiramenai: Yes, taking existing test with STD and make a function recursive works.

llvm/lib/Target/EVM/EVMStackSolver.cpp
akiramenai: I meant we'd better to elaborate the comment. Something similar to the following:

```
propagateThroughMBB() can't detect stack-too-deep errors cased by transformation from BB entry to the first instruction entry. Since a recursive function doesn't support spills and reloads, try to compress it all.
// TODO: make it less aggressive.
```

Maybe a little more verbose.

akiramenai:
How many local changes are needed to make `norecurse` inference work in O0?
Is there a chance that NVPTX, SPIR-V would want it too and support upstreaming?

vladimirradosavljevic:
> How many local changes are needed to make `norecurse` inference work in O0? Is there a chance that NVPTX, SPIR-V would want it too and support upstreaming?

I'm not sure, but if we want to go this path, I would suggest to create a ticket for it, as trying to push this to community, will be slow + I'm not sure of historical reasons why this is disabled for O0.
The only place I see this is used for targets is for AMDGPU (`AMDGPUResourceUsageAnalysis`), but this is just usage analysis and it is not something that changes functionality of a module.

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: Added comment why we have to check entry stack of a successor, and not the exit stack of a current MBB.

llvm/test/CodeGen/EVM/stack-too-deep-1.ll
vladimirradosavljevic: In the end, added just some prints (iteration count, number of registers that we are spilling in each iteration and that we don't expect warning that we allocated more stack size than needed), since if we want to add stack prints where spillable registers are, it would be a lot of prints (e.g. for stack-too-deep-2.ll) and we would add a lot of burden for future changes (a lot more than just generate checks with a script).

llvm/test/CodeGen/EVM/bps-spills-1.mir
vladimirradosavljevic: Added these cases in finalize-stack-frames.mir.

llvm/lib/Target/EVM/EVMStackModel.h
vladimirradosavljevic: Added a new dedicated method isSpill(), as you suggested.

akiramenai:
Recursion is a correctness issue for most of GPUs, not only AMD. They diagnose it in the very beginning of the pipeline, in clang and in the very end (e.g. in PTX). Still vendors like ARM codegen directly from LLVM and keep their backends proprietary. So they might support inferring norecurse in O0. Maybe someone has some local changes already.
I think not maintaining a pass is better than maintaining it, even if slow. So we can go with a ticket, yet it might be better to not to postpone it and at least ask in the community what is the status.
Also, we can keep a local change if it's small. It might be a preferable option to the pass.

akiramenai (approved):
LGTM

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -201,54 +259,184 @@ llvm::calculateStackTransformCost(Stack Source, const Stack &Target,
...
+
+    if (MF.getFunction().hasFnAttribute("evm-recursive")) {
+      // For recursive functions we can't use spills to fix the stack too deep
+      // errors, as we are using memory to spill and not real stack. Report an
+      // error if this function is recursive and we forced compress stack
+      // before.
+      if (ForceCompressStack)
+        report_fatal_error(
+            "Stackification failed for '" + MF.getName() +
+            "' function. It is recursive and has stack too deep errors.");
```
PavelKopyl: Is it worth to add something like "Consider refactoring it to use a non-recursive approach." to the message?

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -305,21 +414,23 @@ void EVMStackifyCodeEmitter::emitStackPermutations(const Stack &TargetStack) {
             Emitter.emitDUP(static_cast<unsigned>(Depth + 1));
             return;
           }
-          if (!Slot->isRematerializable()) {
+          if (!Slot->isRematerializable() && !isSpillReg(Slot)) {
             std::string ErrMsg = getUnreachableStackSlotError(
                 MF, CurrentStack, Slot, Depth + 1, /* isSwap */ false);
             report_fatal_error(ErrMsg.c_str());
```
PavelKopyl, line 420: I wonder in which cases we may get into the stack-too-deep error here?

PavelKopyl (approved):
LGTM, thank!

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -139,6 +145,51 @@ Stack calculateStackBeforeInst(const Stack &InstDefs, const Stack &AfterInst,
...
+  const auto *BestInterval = &LIS.getInterval(SpillableRegs[0]);
+  for (auto Reg : drop_begin(SpillableRegs)) {
+    const auto *LI = &LIS.getInterval(Reg);
+
+    // Take this interval only if it has a non-zero weight and
+    // either BestInterval has zero weight or this interval has a lower
+    // weight than the current best.
+    if (LI->weight() != 0.0F && (BestInterval->weight() == 0.0F ||
+                                 LI->weight() < BestInterval->weight()))
+      BestInterval = LI;
```
PavelKopyl, line 163: I wonder if it would make sense to account for potential stack shuffling overhead? Spilling registers closer to the top of the stack might be more efficient—even if their live interval weight is higher—compared to spilling deeper in the stack at depths like 14 or 15.

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: Done, thanks for the suggestion!

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
vladimirradosavljevic: Mainly, these errors come from [here](https://github.com/matter-labs/era-compiler-llvm/blob/83092e8681d1f0ba1fb5dd31ba3b094863f44296/llvm/lib/Target/EVM/EVMStackShuffler.cpp#L150).

llvm/lib/Target/EVM/EVMStackSolver.cpp
vladimirradosavljevic: This is tricky, since we would also need to check in other stack layouts how this register is shuffled to make better heuristic. Maybe we can do this in a future.

PR #830, opened 2025-06-04
[EVM] Introduce APIs for retrieving spill area size

PavelKopyl:
This change introduces functionality to inform the front-end (FE)
about the spill area size determined by the stackifier. It includes:

I.
A new C-API for the FE to query the spill area size:

 ` uint64_t LLVMGetSpillAreaSizeEVM(LLVMContextRef C);`

New methods in LLVMContext for setting and retrieving the spill area
size calculated by the stackifier:

```
  void setSpillAreaSize(uint64_t Size);

  // Internal use only, utilized by LLVMGetSpillAreaSizeEVM
  uint64_t getSpillAreaSize() const;
```

Spill/Reload workflow:

 - The FE compiles LLVM IR normally without specifying any spill-related
   options

 - If no stackification (STD) errors occur, compilation completes successfully
   and emits object code

 - If a STD error occurs:
    - The `EVMFinalizeStackFrames` pass sets the required spill area size via
     ` LLVMContext::setSpillAreaSize`

    - Compilation fails gracefully (not via assertion) by aborting the compilation pipeline after `EVMFinalizeStackFrames`.
      `LLVMTargetMachineEmitToMemoryBuffer` returns true in this case.

    - The FE then invokes `LLVMGetSpillAreaSizeEVM` to retrieve the required
      spill area size

    - The FE allocates the appropriate spill area and recompiles the LLVM IR,
      passing the allocated size via the `StackRegionSize` option and the memory
      start address via `StackRegionOffset`.

II.
Terminate compilation pipeline after `EVMFinalizeStackFrames` if a spill area is required.
We need to terminate the compilation pipeline due to the presence of `PUSH_FRAME`
instructions, which are introduced by the spilling mechanism but should not appear in the MIR after stackification.

Note:
These changes could be simplified, or potentially avoided altogether if a callback-based mechanism were available for communicating with the front-end. The idea would look like this:

1. Defining a stack-too-deep handler in Rust:
```
void on_std(uint64_t SpillAreaSize) {
  ...
  exit(0);
}
```

2. Registering the callback:
`LLVMInstallEVMStdHandler(on_std);`

3. Invoking the handler from the EVM backend:
`report_evm_std(32);`

With this approach, there would be no need to modify LLVMContext or implement custom logic to interrupt the compilation pipeline. However, it introduces added complexity on the front-end side specifically, implementing the callback in Rust and ensuring clean termination of the current process or thread when a stack-too-deep (STD) error occurs. Given these trade-offs, this approach was ultimately not pursued.

# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion

### DIFF
--- a/llvm/lib/IR/LegacyPassManager.cpp
+++ b/llvm/lib/IR/LegacyPassManager.cpp
@@ -1584,6 +1584,14 @@ MPPassManager::runOnModule(Module &M) {
       removeNotPreservedAnalysis(MP);
     recordAvailableAnalysis(MP);
     removeDeadPasses(MP, M.getModuleIdentifier(), ON_MODULE_MSG);
+
+    // EVM local begin
+    // HACK: Interrupt the Module pass pipeline if the spill area size becomes
+    // non-zero after the preceding pass. This can occur only after the
+    // EVMFinalizeStackFrames pass.
+    if (M.getContext().getSpillAreaSize())
+      break;
+    // EVM local end
   }
 
   // Finalize module passes
--- a/llvm/include/llvm-c/Core.h
+++ b/llvm/include/llvm-c/Core.h
@@ -594,6 +594,13 @@ LLVMDiagnosticHandler LLVMContextGetDiagnosticHandler(LLVMContextRef C);
  */
 void *LLVMContextGetDiagnosticContext(LLVMContextRef C);
 
+// EVM local begin
+/**
+ * Retrieve the spill area size needed for EVM stackification.
+ */
+uint64_t LLVMGetSpillAreaSizeEVM(LLVMContextRef C);
+// EVM local end
+
 /**
  * Set the yield callback function for this context.
  *
--- a/llvm/include/llvm/IR/LLVMContext.h
+++ b/llvm/include/llvm/IR/LLVMContext.h
@@ -270,6 +270,12 @@ public:
   /// "warning: " for \a DS_Warning, and "note: " for \a DS_Note.
   void diagnose(const DiagnosticInfo &DI);
 
+  // EVM local begin
+  void setSpillAreaSize(uint64_t Size);
+
+  uint64_t getSpillAreaSize() const;
+  // EVM local end
+
   /// Registers a yield callback with the given context.
   ///
   /// The yield callback function may be called by LLVM to transfer control back
--- a/llvm/lib/IR/Core.cpp
+++ b/llvm/lib/IR/Core.cpp
@@ -118,6 +118,12 @@ void *LLVMContextGetDiagnosticContext(LLVMContextRef C) {
   return unwrap(C)->getDiagnosticContext();
 }
 
+// EVM local begin
+uint64_t LLVMGetSpillAreaSizeEVM(LLVMContextRef C) {
+  return unwrap(C)->getSpillAreaSize();
+}
+// EVM local end
+
 void LLVMContextSetYieldCallback(LLVMContextRef C, LLVMYieldCallback Callback,
                                  void *OpaqueHandle) {
   auto YieldCallback =
--- a/llvm/lib/IR/LLVMContext.cpp
+++ b/llvm/lib/IR/LLVMContext.cpp
@@ -283,6 +283,14 @@ void LLVMContext::diagnose(const DiagnosticInfo &DI) {
     exit(1);
 }
 
+// EVM local begin
+void LLVMContext::setSpillAreaSize(uint64_t Size) {
+  pImpl->SpillAreaSize = Size;
+}
+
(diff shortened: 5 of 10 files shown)


llvm/lib/IR/LegacyPassManager.cpp
```diff
@@ -1535,7 +1535,11 @@ MPPassManager::runOnModule(Module &M) {
   if (EmitICRemark)
     InstrCount = initSizeRemarkInfo(M, FunctionToInstrCount);
 
-  for (unsigned Index = 0; Index < getNumContainedPasses(); ++Index) {
+  // EVM local begin
+  for (unsigned Index = 0;
+       Index < getNumContainedPasses() && !M.getContext().getSpillAreaSize();
```
akiramenai: Do we need this condition to be a part of the loop?

akiramenai (changes requested):
Could we explain how it supposed to work?
We terminate unconditionally if the spill area is required, but we need to check against how much was actually allocated for spills.

PavelKopyl:
> Could we explain how it supposed to work? We terminate unconditionally if the spill area is required, but we need to check against how much was actually allocated for spills.

Note: This is a workaround due to the lack of a proper mechanism in LLVM to gracefully stop the pass manager mid-pipeline. Typically, if an unrecoverable error occurs during compilation, report_fatal_error is used, which forcibly terminates the process via exit or abort.

In our case, we interrupt the Module pass pipeline if the spill area size becomes non-zero after the previous pass. This situation can arise only after the EVMFinalizeStackFrames pass. Here’s how the spill area is set up in that pass:
```
--- a/llvm/lib/Target/EVM/EVMFinalizeStackFrames.cpp
+++ b/llvm/lib/Target/EVM/EVMFinalizeStackFrames.cpp
@@ -160,10 +160,10 @@ bool EVMFinalizeStackFrames::runOnModule(Module &M) {

   // Check if it is valid to replace frame indices.
   if (TotalStackSize > 0) {
-    if (TotalStackSize > StackRegionSize)
-      report_fatal_error("Total stack size: " + Twine(TotalStackSize) +
-                         " is larger than the allocated stack region size: " +
-                         Twine(StackRegionSize));
+    if (TotalStackSize > StackRegionSize) {
+      M.getContext().setSpillAreaSize(TotalStackSize);
+      return true;
+    }
```
This logic is not directly tied to the StackRegionSize option - it’s used solely to abort the pipeline. The effect of the abortion is that we emit empty assembly/objects elf file.

llvm/lib/IR/LegacyPassManager.cpp
PavelKopyl: Yes, because the return value of getSpillAreaSize() may change after the last pass has been executed.
But we can put it at the end of the loop body:
```
if (M.getContext().getSpillAreaSize())
      break;
```

akiramenai: Yes, it would be better to not to modify the condition and do as you suggested. Adding necessary documentation there.

akiramenai:
Could we put proper documentation to code and commit messages?
I also still don't understand why we terminate the pipeline if spill area size become non-zero. What is the meaning of spill area size? How do we handle the case where FE does allocate space for spills?

llvm/lib/IR/LegacyPassManager.cpp
PavelKopyl: Sure, done.

PavelKopyl:
> Could we put proper documentation to code and commit messages? I also still don't understand why we terminate the pipeline if spill area size become non-zero. What is the meaning of spill area size? How do we handle the case where FE does allocate space for spills?

I believe I understand your concern. I’ve included a detailed explanation of how the solution works and how it relates to the StackRegionSize and StackRegionOffset options in the commit message for the patch titled "[EVM] Introduce APIs for retrieving spill area size." Please let me know if any further clarification is needed.
I’d also like to point out that the patch modifying the Pass Manager is intended purely to reduce compilation time—since encountering a STD error means there's no reason to continue the compilation pipeline.

hedgar2017 (approved):

PavelKopyl:
Close in favor of #839

PR #833, opened 2025-06-05
[EVM] Implement large constants unfolding after stackification

PavelKopyl:
A large constant is replaced with a sequence of
operations to reduce a bytecode size.

For example,

  0xFFFFFFFF00000000000000000000000000000000000000000000000000000000

can be replaced with

  0xFFFFFFFF << 224
  
This pass runs after stackification to prevent additional stack-too-deep errors, which could otherwise arise from increased register pressure caused by a higher number of live values.

That said, there may be potential benefits to performing unfolding before stackification, since identical immediate values (if present) could be deduplicated during stackification. It would be reasonable to revisit this approach after resolving the stack-too-deep issues introduced by LiteralSlots.

Further possible enhancements concerning large constants (not limited to this patch):
 - Disable unfolding inside a loop when optimizing for speed
 - Spill frequently used constants to memory. This can be done at the module level, allowing all functions to access a shared "constant pool" in memory.
# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+  unsigned ValLen = Val.getActiveBits();
+  APInt AbsVal = Val.abs();
+  unsigned AbsValLen = AbsVal.getActiveBits();
+  bool IsMask = ((Ones + LeadZ + TrailZ) == Imm.getBitWidth());
+  assert(ValLen == (Imm.getBitWidth() - TrailZ - LeadZ));
+  assert(Val.isNegative());
+
+  // The upper limit for both size and instruction count.
+  constexpr unsigned MaxCost = /* PUSH32 size*/ 33 * 2;
+  SmallVector<std::pair<std::function<std::pair<unsigned, unsigned>(void)>,
```
akiramenai: Is it possible to name these structures (outer and inner) for better readability?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+#include "llvm/CodeGen/Passes.h"
+#include "llvm/CodeGen/TargetInstrInfo.h"
+#include "llvm/InitializePasses.h"
+#include "llvm/Support/Debug.h"
+#include "llvm/Support/raw_ostream.h"
+#include <functional>
+
+using namespace llvm;
+
+#define DEBUG_TYPE "evm-constant-unfolding"
```
akiramenai, line 42: Do we want to add statistics for this patch?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+        // Expressing total cost in terms of both size and number of operations
+        // enables more performance-friendly transformation choices. For
+        // example, consider the following two options:
+        //
+        //   [size: 7, OpNum: 5]
+        // vs
+        //   [size: 8, OpNum: 3]
+        //
+        // The second option is likely preferable, as the operation count is
+        // notably lower while the size remains nearly the same.
```
akiramenai: Do we have any perf numbers for this?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+      dbgs() << "Applying optimal transform : [size: " << Cost.first
+             << ", OpNum: " << Cost.second << "]\n";
+    });
+  } else {
+    unsigned OrigSize = (alignTo(Imm.getActiveBits(), 8) / 8) + 1;
+    // When optimizing for speed, only unfold constants if it reduces the size
+    // by a factor of at least 3 and keeps the instruction count to 3 or fewer.
+    // This seems to be an effective heuristic that avoids significantly
+    // impacting performance.
+    if (OrigSize / Cost.first >= 3 && Cost.second <= 3) {
```
akiramenai: How this 'magic number' was calculated? Could we extract it to a named variable for better readability?

llvm/test/CodeGen/EVM/constant-unfolding.ll
```diff
@@ -0,0 +1,267 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
```
akiramenai, line 1: Could we add a test that we do not try to unfold small constants?

llvm/test/CodeGen/EVM/constant-unfolding-Oz.ll
```diff
@@ -0,0 +1,304 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
+; RUN: llc -O3 < %s | FileCheck %s
+target datalayout = "E-p:256:256-i256:256:256-S256-a:256:256"
+target triple = "evm-unknown-unknown"
+
+declare void @llvm.evm.return(ptr addrspace(1), i256) noreturn
+
+; 0x4e487b7100000000000000000000000000000000000000000000000000000000
```
akiramenai, line 10: I'd add a short clarification why we expect here (and below) one or another transformation done.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+  }
+
+  if (Changed)
+    MI.eraseFromParent();
+
+  return Changed;
+}
+
+bool static isPUSH(const MachineInstr &MI) {
+  switch (MI.getOpcode()) {
```
akiramenai: Can we tablegen it?

akiramenai:
@kpv — sorry for the slow review. I’m looking at the patch now and should finish tonight or tomorrow.

I suggest two improvements:

1. **TableGen the `isPUSH` switch.** You can also TableGen aux functions from `EVMTargetInfo.cpp`.  
2. **Handle minsize functions and cold code.** For example:
   ```cpp
   auto &MBFI = getAnalysis<MachineBlockFrequencyInfo>();
   BlockFrequency Freq = MBFI.getBlockFreq(&MBB);
   if (Freq < SomeThreshold) {
     // treat MBB as cold
   }

akiramenai:
Also, do you know why do we have gas improvements in this patch? It's not what I would expect if we replace one instruction with multiple.

PavelKopyl:
> Also, do you know why do we have gas improvements in this patch? It's not what I would expect if we replace one instruction with multiple.

As I understand that's because we reduce the deployment bytecode size, for which we are charged less in gas at deploy time. It seems that, for CI tests, the savings from reduced deployment costs outweigh the increased gas usage at runtime.

akiramenai:
So we count deployment costs towards `Gas (-%)` category, don't we? But what is the assumption on how many time the contract will be run? Coupling deploy-time and run-time gas looks confusing to me.
@antonbaliasnikov are we correct? Is there a way we can decouple?
@PavelKopyl for this particular patch it would be great to understand what happened to runtime gas. Is there a way to get this data? Maybe with the performance stand if the CI can't help now.

akiramenai:
Also, do `cycles` reflect reality in some shape or form? I remember there were some concerns in how we measure cycles via the interpreter.
They went down as well.

PavelKopyl:
> So we count deployment costs towards `Gas (-%)` category, don't we? But what is the assumption on how many time the contract will be run? Coupling deploy-time and run-time gas looks confusing to me. @antonbaliasnikov are we correct? Is there a way we can decouple? @PavelKopyl for this particular patch it would be great to understand what happened to runtime gas. Is there a way to get this data? Maybe with the performance stand if the CI can't help now.

>So we count deployment costs towards `Gas (-%)` category, don't we? But what is the assumption on how many time the contract will be run? Coupling deploy-time and run-time gas looks confusing to me.

Yes, these are good questions. I hope @antonbaliasnikov will clarify how exactly that values are calculated. I'll also will check `benchmark-analyzer` tool.

 >it would be great to understand what happened to runtime gas. Is there a way to get this data? Maybe with the performance stand if the CI can't help now.

I think we can tweak `benchmark-analyzer` to just ignore lines containing #deployer to take into account only the runtime values.

PavelKopyl:
> Also, do `cycles` reflect reality in some shape or form? I remember there were some concerns in how we measure cycles via the interpreter. They went down as well.

Right, as @hedgar2017 explained to me, we should ignore `cycles`, as they do not corresponds to actually EVM cycles.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+  unsigned AbsValLen = AbsVal.getActiveBits();
+  bool IsMask = ((Ones + LeadZ + TrailZ) == Imm.getBitWidth());
+  assert(ValLen == (Imm.getBitWidth() - TrailZ - LeadZ));
+  assert(Val.isNegative());
+
+  // The upper limit for both size and instruction count.
+  constexpr unsigned MaxCost = /* PUSH32 size*/ 33 * 2;
+  SmallVector<std::pair<std::function<std::pair<unsigned, unsigned>(void)>,
+                        std::function<bool(void)>>>
+      Transformations = {
```
akiramenai: I would prefer extracting transformation as functions, name them properly and attach documentation. If lamas are needed to capture, we can capture and make a call.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -0,0 +1,377 @@
...
+    Changed = OptTransform->second();
+    LLVM_DEBUG({
+      dbgs() << "Applying optimal transform : [size: " << Cost.first
+             << ", OpNum: " << Cost.second << "]\n";
+    });
+  } else {
+    unsigned OrigSize = (alignTo(Imm.getActiveBits(), 8) / 8) + 1;
+    // When optimizing for speed, only unfold constants if it reduces the size
+    // by a factor of at least 3 and keeps the instruction count to 3 or fewer.
+    // This seems to be an effective heuristic that avoids significantly
```
akiramenai: > This seems to be an effective heuristic

Could you give some objective data in support?

akiramenai (changes requested):

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Done.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Done.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: I've just checked on small tests that the comparison in non-OptForSize mode:
`(CostA.first + CostA.second) < (CostB.first + CostB.second)`
is more adequate than the comparison for OptForSize mode.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: I got it experimentally on our CI tests, I've converted these two magic numbers into options.

llvm/test/CodeGen/EVM/constant-unfolding.ll
PavelKopyl: Done.

llvm/test/CodeGen/EVM/constant-unfolding-Oz.ll
PavelKopyl: We choose a concrete transformation based on the cost calculations and choosing the cheapest one. Do you mean we need to specify costs for each possible transformation for each test?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Yeap, done.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Sure, done.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Please, find below the aave-v3 gas/size changes after enabling this optimization.

[aave-const-unfolding.txt](https://github.com/user-attachments/files/20909219/aave-const-unfolding.txt)

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: I could check different metrics, but this should be done on real-life projects.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Please, also note that, as I understand, it's not possible to completely split size/gas costs, because when we execute CREATE opcode, the consumed amount of gas (runtime) depends on a size of a new contract.

akiramenai (approved):
LGTM, thank you

hedgar2017:
@akiramenai @PavelKopyl needs squashing?

PavelKopyl:
> @akiramenai @PavelKopyl needs squashing?

Done.

PR #834, opened 2025-06-10
[EVM] Update live interval during ADD remat in EVMSingleUseExpression

vladimirradosavljevic:
Since EVMSingleUseExpression preserves LiveIntervals analysis, it is important to ensure that the live intervals are correctly updated when rematerializing instructions. In this case, during the ADD rematerialization, new instruction is created and the live interval of base register is not updated. As a result, we are hitting assertion during stackification.


PavelKopyl (approved):
Thanks, LGTM.

PR #838, opened 2025-06-19
[EVM] Refine the memory effects of intrinsics for greater AA accuracy + custom extension of DSE

PavelKopyl:
EVM opcodes can be categorized based on how their behavior or output is influenced by the transaction state:

 - Readnone (Pure)
 - Volatile (State-Dependent)
 - Side-Effecting (State-Changing) (See: https://github.com/matter-labs/era-compiler-llvm/wiki/EVM-opcodes-categorization)

This patch refines the memory attributes of LLVM intrinsics corresponding to that opcodes. The transaction-scoped EVM state is represented at the LLVM IR level by reading/writing to the 'Inaccessible memory'. Notably, this EVM state does not include the heap, which is modeled separately through standard LLVM pointer parameters.
State-dependent intrinsics are now marked as reading from 'Inaccessible memory'. State-changing intrinsics are marked as both reading and writing to it.

To model memory dependencies between plain loads/stores to storage or transient storage and state-modifying intrinsics, the latter are transformed into extended versions of the same intrinsics. These extended versions include two additional pointer arguments:

 ` ptr addrspace(5), ptr addrspace(6)`

For example

 ` %ret = call i256 @llvm.evm.create(i256 %val, ptr addrspace(1) %off, i256 %size)`

transformed to

 ` %ret = call i256 @llvm.evm.create.sptr(i256 %val, ptr addrspace(1) %off, i256 %size, ptr addrspace(5) null, ptr addrspace(6) null)`

This conversion is handled by the EVMSystemIntrinsicsPromotion pass, which runs early in the LLVM IR pipeline.
It's possible to model these memory dependencies by custom changes in target-independent parts of LLVM’s Alias Analysis (AA), but this is undesirable.

Finally, note that 'revert' and 'return' intrinsics also have extended variants with two additional readonly pointer parameters. This prevents LLVM’s Dead Store Elimination (DSE) from incorrectly removing stores to storage or transient storage.

# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


llvm/test/CodeGen/EVM/stack-ops-commutable.ll
```diff
@@ -405,12 +405,12 @@ do:
...
+  call void @llvm.evm.return.sptr(ptr addrspace(1) null, i256 32, ptr addrspace(5) null, ptr addrspace(6) null)
   unreachable
 }
 
 declare i256 @llvm.evm.addmod(i256, i256, i256)
 declare i256 @llvm.evm.mulmod(i256, i256, i256)
 declare i256 @llvm.evm.calldataload(ptr addrspace(2))
-declare void @llvm.evm.return(ptr addrspace(1), i256)
-declare void @llvm.evm.revert(ptr addrspace(1), i256)
+declare void @llvm.evm.return.sptr.sptr(ptr addrspace(1), i256, ptr addrspace(5), ptr addrspace(6))
```
akiramenai: why double `sptr`?

llvm/lib/Transforms/Scalar/DeadStoreElimination.cpp
```diff
@@ -1828,8 +1857,17 @@ struct DSEState {
         // underlying objects is very uncommon. If it turns out to be important,
         // we can use getUnderlyingObjects here instead.
         const Value *UO = getUnderlyingObject(DefLoc->Ptr);
-        if (!isInvisibleToCallerAfterRet(UO))
-          continue;
+        if (!isInvisibleToCallerAfterRet(UO)) {
```
akiramenai, line 1860: why didn't we patch `isInvisibleToCallerAfterRet`?

llvm/lib/Transforms/Scalar/DeadStoreElimination.cpp
```diff
@@ -1828,8 +1857,17 @@ struct DSEState {
...
+          // EVM local begin
+          Triple TT(DefI->getFunction()->getParent()->getTargetTriple());
+          unsigned LocAS = DefLoc->Ptr->getType()->getPointerAddressSpace();
+          if (!TT.isEVM() || LocAS == 5 || LocAS == 6)
```
akiramenai: Can we use something instead of magic numbers?

llvm/lib/Transforms/Scalar/DeadStoreElimination.cpp
```diff
@@ -1828,8 +1857,17 @@ struct DSEState {
...
+            continue;
+
+          if (isOnPathToFunctionReturn(Def))
```
akiramenai: Can we document it? Is it upstreamable?

llvm/include/llvm/IR/IntrinsicsEVM.td
```diff
@@ -118,35 +119,43 @@ def int_evm_origin : Intrinsic<[llvm_i256_ty], [], [IntrNoMem, IntrWillReturn]>;
...
 def int_evm_calldataload
   : Intrinsic<[llvm_i256_ty], [LLVMQualPointerType<AS.CALL_DATA>],
-              [IntrReadMem, IntrArgMemOnly]>;
+              [IntrReadMem, IntrArgMemOnly, IntrNoCallback]>;
 
 def int_evm_codesize : Intrinsic<[llvm_i256_ty], [], [IntrNoMem, IntrWillReturn]>;
 
 def int_evm_gasprice : Intrinsic<[llvm_i256_ty], [], [IntrNoMem, IntrWillReturn]>;
 
+// This should have 'read of inaccessible' memory along with read/write memory
```
akiramenai: Could you clarify this documentation? Is it something left to do? Why can't we do it with this patch?

akiramenai (changes requested):

llvm/test/CodeGen/EVM/stack-ops-commutable.ll
PavelKopyl: Thank you, fixed!

llvm/lib/Transforms/Scalar/DeadStoreElimination.cpp
PavelKopyl: I don't fell that `isInvisibleToCallerAfterRet` is the appropriate place for EVM-specific logic, because it serves a fundamentally different purpose from the newly introduced isOnPathToFunctionReturn.
`isInvisibleToCallerAfterRet` determines whether the allocated (underlying) object (U0) has been captured or escaped from the current function.
In contrast, `isOnPathToFunctionReturn` checks whether the basic block containing a potentially dead store lies on a path leading to a block that ends with a ret instruction.
I also consider this as a temporary solution, as I'm going to move all the custom logic from DSE to an EVM pass.

llvm/lib/Transforms/Scalar/DeadStoreElimination.cpp
PavelKopyl: Thank you for pointing that out. I'll remove the address space check, as it's no longer necessary. The EVMSystemIntrinsicsPromotion pass now adds ptr addrspace(5) null, ptr addrspace(6) null parameters to system revert/return intrinsics, which ensures proper dependency tracking between stores and those intrinsics.

llvm/include/llvm/IR/IntrinsicsEVM.td
PavelKopyl: I've removed that comments.

llvm/lib/Transforms/Scalar/DeadStoreElimination.cpp
PavelKopyl: It reasonable to evaluate this transformation across all targets (i.e., without limiting it to EVM) to assess whether it could be upstreamed. At first glance, this appears to be EVM-specific, since in the general case it is not guaranteed that memory side effects preceding an unreachable can be safely ignored.

PavelKopyl:
Close it in favor of https://github.com/matter-labs/era-compiler-llvm/pull/865

PR #840, opened 2025-06-24
[EVM] Improvements for spill mechanism during stackification

vladimirradosavljevic:


akiramenai:
Do we have any performance data for this PR?

llvm/test/CodeGen/EVM/bps-spill-unreachable-arg.mir
```diff
@@ -1,15 +1,13 @@
-# RUN: not --crash llc -x mir -run-pass=evm-backward-propagation-stackification < %s 2>&1 | FileCheck %s
+# NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
+# RUN: llc -x mir -run-pass=evm-backward-propagation-stackification < %s | FileCheck %s
 
-# Test that stackification fails when unreachable register is spilled.
-
-# CHECK: Unexpected stack depth
+# Test that unreachable reg is not spilled.
```
akiramenai: Could you please add more details to the test's comment about which register and why we no longer spill? 
I'm also not sure that checking the whole asm output is useful. I realize that we previously agreed on pre-commit testing with auto-generated tests, but asm adds noice around stack manipulations and it's also problematic to do LLVM update as changes could become different from they were before. Maybe we should rediscuss.

llvm/test/CodeGen/EVM/bps-spill-unreachable-ret.mir
```diff
@@ -190,6 +187,228 @@ machineFunctionInfo:
   numberOfParameters: 0
   hasPushDeployAddress: false
 body:             |
+  ; CHECK-LABEL: name: main
```
akiramenai, line 193: Same as above.

vladimirradosavljevic:
> Do we have any performance data for this PR?

For that, we would need to have everything setup so we can run the contracts like [here](https://github.com/matter-labs/solx/pull/77#issuecomment-2961898496) + to get perf data from there.
I looked at some files in some projects and saw some small improvements like it is shown in the test cases added in this PR. Please note that these are not frequent cases.

akiramenai (approved):
If we can improve documentation around tests, it would be great.
We might also want to reconsider always autogenerating checks. But it's a different scope.

llvm/test/CodeGen/EVM/bps-spill-unreachable-arg.mir
vladimirradosavljevic: Added comment which registers have unreachable def and those are not taken into account for spilling.
Regarding auto-generated tests, during LLVM update you can just use script to update them, and you can spot on whether some of the tests are not valid anymore, since something changed between updates. Auto-generated tests are not just here to make it easier to write tests, they are also good to check whether some change that doesn't have to do with some test is actually change it. At this point, you know something is wrong with the implementation and you can easily fix it, and not wait CI to finish test and do deeper debug.
Please note that these are .mir tests that are only calling stackification, so these won't by affected by the update, only during development of stackification algorithm.

llvm/test/CodeGen/EVM/bps-spill-unreachable-ret.mir
vladimirradosavljevic: Done. Added comment which registers are not taken into account to spill.

PR #841, opened 2025-06-26
[EVM][BPS] Don't rematerialize large constants in optsize

vladimirradosavljevic:
In case we are optimizing for size, it is profitable to propagate large immediates, in case it will reduce code size (e.g. if the same large immediate is used in two different instructions in the same MBB, we can only once do PUSH instead of doing 2 PUSHes).
This support can be enabled when optimizing for speed (just different threshold for PUSH) and if second operand of commutable instruction is rematerializable. Idea of this PR is to check the approach, and later we can expand it in more places.


llvm/lib/Target/EVM/EVMStackSolver.h
```diff
@@ -111,6 +111,15 @@ class EVMStackSolver {
   /// for function arguments and return values.
   bool hasUnreachableDef(const Register &Reg) const;
 
+  /// Return true if \p Slot is a rematerializable slot and it can be removed.
+  /// In cases of LiteralSlot, it checks if the literal is used before \p MI,
+  /// and if it is not, it can be removed to reduce stack pressure. In case it
+  /// is used, we don't remove it to reduce size of the stack, as PUSH
+  /// instructions are large. Currently, this check whether to propagate large
+  /// literals is only enabled when optimizing for size.
+  bool shouldRemoveRematSlot(const StackSlot *Slot,
```
akiramenai: `shouldRemoveSlot`?
As it also checks for remateriabitily.

llvm/lib/Target/EVM/EVMStackSolver.cpp
```diff
@@ -483,6 +483,44 @@ static bool spillRegHasNoUseOrDefBefore(const StackSlot *Slot,
...
+bool EVMStackSolver::shouldRemoveRematSlot(const StackSlot *Slot,
+                                           const MachineInstr &MI) const {
+  const auto *Literal = dyn_cast<LiteralSlot>(Slot);
+  if (!Literal)
+    return Slot->isRematerializable();
+
+  // For now, remove all literals if we are not optimizing for size to reduce
+  // stack pressure.
+  if (!MI.getMF()->getFunction().hasOptSize())
+    return true;
```
PavelKopyl: I'm curious, did this change result in additional stack-too-deep (STD) errors in real-world cases when optimizing for size? And if it does what are the size changes in case of STD?

vladimirradosavljevic: I'm also curious about that 😄, so let's wait till we have all projects included in testing after we merge all stack-too-deep patches. Until then, does this approach make sense to you?
FWIW, I was surprised that with this change we don't have additional stack-too-deep issues in our CI.

akiramenai (approved):

akiramenai:
I would make the commit title more informative: Rematerialize small constants instead of keeping them on stack in optimize.

llvm/lib/Target/EVM/EVMStackSolver.cpp
akiramenai: Ideally we need a common heuristic for optimizing for size.
Which is not just optsize, but also could be cold code, low branch probability.
Maybe it makes sense to create an auxiliary predicate and use it for constant unfolding patch too.

PavelKopyl (approved):
LGTM

llvm/lib/Target/EVM/EVMStackSolver.h
vladimirradosavljevic: Done.

vladimirradosavljevic:
> I would make the commit title more informative: Rematerialize small constants instead of keeping them on stack in optimize.

Before this PR, we already did that, along with the large constants. I changed the title to:
`Don't rematerialize large constants in optsize`

vladimirradosavljevic:
Current numbers doesn't reflect this change only, since it contains fix for stack-too-deep errors, as `era-compiler-llvm-test` [tag](https://github.com/matter-labs/era-compiler-tester/tags) is used.

PR #844, opened 2025-07-08
[EVM] Clear transformation cache in EVMConstantUnfolding

vladimirradosavljevic:
In some cases, EVMConstantUnfolding is not destroyed after the run, and can be reused for other functions. In that case, we don't want to reuse the cache from the previous runs.


PavelKopyl (approved):
LGTM, thank you!

PR #852, opened 2025-07-15
[EVM] Support data sections in the code

PavelKopyl:
This is achieved by emitting global variables with constant
 initializers into the code section.
 Removed unused patters from TableGen.

# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


llvm/lib/Target/EVM/EVMAsmPrinter.cpp
```diff
@@ -323,8 +342,59 @@ void EVMAsmPrinter::emitWideRelocatableSymbol(const MachineInstr *MI) {
...
+  auto IsSubstring = [&UniqueInitializers](const StringRef &A) {
+    return any_of(UniqueInitializers,
+                  [&A](const StringRef &B) { return A != B && B.contains(A); });
+  };
 
+  UniqueInitializers.erase(remove_if(UniqueInitializers, IsSubstring),
+                           UniqueInitializers.end());
+
+  // Build the data section by concatenating unique initializers.
+  // NOTE: Rather than simply concatenating unique strings, we could attempt
```
hedgar2017, line 371: Certainly an overkill and not a real-life use case.
We can probably try it if it's requested by users one day, maybe some cryptographic libraries can have something alike.

I thought the substring thing was just very easy to do, as it should work naturally if you walk through the byte arrays from the longest to the shortest.

llvm/test/CodeGen/EVM/data-sections-in-code.ll
```diff
@@ -0,0 +1,45 @@
...
+; CHECK-NEXT:    PUSH1 0x5
+; CHECK-NEXT:    PUSH4 @code_data_section+6
+; CHECK-NEXT:    PUSH1 0x40
+; CHECK-NEXT:    CODECOPY
+; CHECK-NEXT:    PUSH1 0x80
+; CHECK-NEXT:    PUSH0
+; CHECK-NEXT:    REVERT
+; CHECK:         INVALID
+; CHECK-LABEL:  code_data_section:
+; CHECK-NEXT:    .byte	0x68, 0x65, 0x6c, 0x6c
```
hedgar2017: 🔥

hedgar2017 (changes requested):
Let's wait until my EVM interpreter PR is merged to check more accurate gas figures.

hedgar2017 (approved):
The [data](https://github.com/matter-labs/era-compiler-tester/pull/177#issuecomment-3091331652) looks good now!
Please fix the tests and let's merge.

llvm/lib/Target/EVM/EVMAsmPrinter.cpp
```diff
@@ -328,8 +347,49 @@ void EVMAsmPrinter::emitWideRelocatableSymbol(const MachineInstr *MI) {
...
+
+    const auto *CV = dyn_cast<ConstantDataSequential>(GV.getInitializer());
+    if (!CV)
+      continue;
+
+    Globals.emplace_back(&GV, CV->getRawDataValues());
+  }
+  // Sort global variables in descending order based on the size of their
+  // initializers.
+  sort(Globals, [](const auto &A, const auto &B) {
```
vladimirradosavljevic: I think we have to use stable_sort here, since we could get two different globals with the same size, and run into non-determinizm here. Could you also add test case for this?

llvm/lib/Target/EVM/EVMAsmPrinter.cpp
```diff
@@ -328,8 +347,49 @@ void EVMAsmPrinter::emitWideRelocatableSymbol(const MachineInstr *MI) {
...
 
-void EVMAsmPrinter::emitEndOfAsmFile(Module &) {
+void EVMAsmPrinter::createDataSectionBuffer(const Module &M) {
+  SmallVector<std::pair<const GlobalVariable *, StringRef>, 16> Globals;
+  for (const GlobalVariable &GV : M.globals()) {
+    if (GV.getAddressSpace() != EVMAS::AS_CODE || !GV.hasInitializer())
+      continue;
+
+    const auto *CV = dyn_cast<ConstantDataSequential>(GV.getInitializer());
+    if (!CV)
```
vladimirradosavljevic, line 357: In case of empty uses of CV, we could skip it, right?

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
```diff
@@ -117,12 +117,19 @@ void EVMStackifyCodeEmitter::CodeEmitter::emitConstant(uint64_t Val) {
 
 void EVMStackifyCodeEmitter::CodeEmitter::emitSymbol(const MachineInstr *MI,
                                                      MCSymbol *Symbol) {
-  assert(isLinkerPseudoMI(*MI) && "Unexpected symbol instruction");
```
vladimirradosavljevic, line 120: Maybe you can leave this assert and check also for `CODECOPY`. It will simplify handling below.

llvm/lib/Target/EVM/MCTargetDesc/EVMMCCodeEmitter.cpp
```diff
@@ -102,25 +102,40 @@ unsigned EVMMCCodeEmitter::getMachineOpValue(const MCInst &MI,
                                              const MCSubtargetInfo &STI) const {
   if (MO.isImm()) {
     Op = MO.getImm();
-  } else if (MO.isExpr()) {
-    auto Kind = MO.getExpr()->getKind();
+    return 0;
+  }
+
+  if (MO.isExpr()) {
```
vladimirradosavljevic: Maybe you can reverse this condition and add here llvm_unreachable.

vladimirradosavljevic (changes requested):
Overall looking good, just some comments to be addressed. Thanks!

llvm/test/MC/EVM/glob-symbol-with-offset.ll
```diff
@@ -0,0 +1,31 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 5
+; RUN: llc -filetype=obj %s -o - | llvm-objdump --disassemble - | FileCheck %s
+target datalayout = "E-p:256:256-i256:256:256-S256-a:256:256"
+target triple = "evm"
+
+@data1 = private unnamed_addr addrspace(4) constant [13 x i8] c"hello world\0A\00"
```
vladimirradosavljevic, line 6: Since unused private constants will be removed during OPT phase, maybe you can update this test as we won't see this.

PavelKopyl: I'm sorry, I didn't get why it's not used? Do you mean that identical global constants should be merged during OPT? Unfortunately, this is not true for non-default address space:  https://github.com/matter-labs/era-compiler-llvm/blob/e6dd841571f79edbd416433dbb3840848261375b/llvm/lib/Transforms/IPO/ConstantMerge.cpp#L96.

vladimirradosavljevic: I meant that `@data1` in this test is not used. Sorry for the confusion.

PavelKopyl: Ah, sorry I looked at wrong test) OK, I'll add use of the @data1. I just wanted to check how MC layer handles @symbol with non-zero immediate.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
PavelKopyl: Yes, we could treat `CODECOPY` as a special case here, but my intention was to distinguish instructions with symbol operands that are resolved at link time from those resolved at the MC layer. I wanted to support this distinction as a general mechanism rather than as an exception. (Sorry, I missed to remove the assert(`MI->getOpcode() == EVM::CODECOPY)`). To that end, I also modified `getSlotsForInstructionUses()` to handle MCSymbol operands as a general case (and also I wanted to support Imm operands the same way).
Nevertheless, `CODECOPY` is currently the only such case and likely will remain so.
What do you think?

PavelKopyl:
> The [data](https://github.com/matter-labs/era-compiler-tester/pull/177#issuecomment-3091331652) looks good now! Please fix the tests and let's merge.

The error comes from clang-tidy, but that case is typical for LLVM code, so we can ignore it.

llvm/lib/Target/EVM/EVMStackifyCodeEmitter.cpp
vladimirradosavljevic: Sorry if I'm missing something but if you do:
```
assert((isLinkerPseudoMI(*MI) || MI->getOpcode() == EVM::CODECOPY) && "Unexpected symbol instruction");
```
It will be seen that codecopy is not in the linker pseudo group.
I'm not saying that you should remove assert below, I'm saying that if you move codecopy opcode check inside the assert in the beginning, below code will just be if + BuildMI.

llvm/lib/Target/EVM/EVMAsmPrinter.cpp
PavelKopyl: Thank you, added.

vladimirradosavljevic (approved):
LGTM.

PR #854, opened 2025-07-16
[EVM] Constant unfolding -Oz fallback

PavelKopyl:
Key changes introduced:
    
 - Transitioned from a MachineFunction pass to a Module-level pass
 - Implemented module-wide code size estimation
    (not 100% exact but sufficiently accurate for our purpose)
 - Applied transformations by iterating over functions within the module.
    For each function:
      - Handled both loop and non-loop basic blocks using the default heuristic
         (i.e., based on the presence of the OptSize attribute)
      - If the module size exceeds the defined limit, switched to a fallback mode
         that treats all functions as if they have OptSize flag.
 - Repeatedly traversed all functions, targeting only basic blocks at a specific
    loop nesting depth (starting from depth 0 for non-loop blocks), and
    re-evaluated module size. The process stops once the size drops below the
    limit, or reaching the loop depth limit.
    
This enables a balanced distribution of runtime gas regressions across functions
 within the module, resulting from aggressive constant unfolding.

# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion


llvm/test/CodeGen/EVM/constant-unfolding-Oz-fallback.ll
```diff
@@ -0,0 +1,77 @@
+; RUN: llc -O3 --debug-only=evm-constant-unfolding -evm-bytecode-sizelimit=80 < %s 2>&1 | FileCheck %s
```
vladimirradosavljevic, line 1: You will need to add here `; REQUIRES: asserts`, since this test will fail without asserts.

llvm/test/CodeGen/EVM/constant-unfolding-Oz.ll
```diff
@@ -60,12 +57,9 @@ define void @test4() #1 {
 ; CHECK-LABEL: test4:
 ; CHECK:       ; %bb.0: ; %entry
 ; CHECK-NEXT:    JUMPDEST
-; CHECK-NEXT:    PUSH0
-; CHECK-NEXT:    NOT
-; CHECK-NEXT:    PUSH1 0xC8
+; CHECK-NEXT:    PUSH7 0xFFFFFFFFFFFFFF
```
vladimirradosavljevic, line 60: IIUC, for -Oz we want to have bytecode as small as possible. Is there a reason for this change?

llvm/lib/Target/EVM/EVMInstrFormats.td
```diff
@@ -36,32 +36,31 @@ class EVMInst<string asmstr, bit stack, int cost>
   bit IsPush = 0;
   bit StackBased = stack;
   string BaseName = NAME;
-  int GasCost     = cost;
+  bits<24> BaseGasCost = cost;
```
vladimirradosavljevic: Is there are reason why we need 24 bits? We could do it in 15 bits, as CREATE and CREATE2 are 32000.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -411,174 +427,313 @@ EVMConstantUnfolding::findOptimalTransfomtation(const APInt &Imm) {
...
+      TransformationCache.try_emplace({Imm, OptForSize}, std::move(*OptIt));
   assert(Res.second);
 
   return Tr;
 }
 
-bool EVMConstantUnfolding::tryUnfoldConstant(MachineInstr &MI) {
+static bool isProfitableToTranform(const APInt &Imm, const StackCostModel &Cost,
+                                   bool OptForSize) {
+  if (!OptForSize) {
```
vladimirradosavljevic: Please reverse condition and do early exit.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -411,174 +427,313 @@ EVMConstantUnfolding::findOptimalTransfomtation(const APInt &Imm) {
...
+  // TODO: remove this workaround once the EVM CodeGen pipeline adopts
+  // the new pass manager, #862.
+
+  // The order of object construction is important to avoid
+  // stack-use-after-scope error.
+  MachineFunctionAnalysisManager MFAM;
+  ModuleAnalysisManager MAM;
+
+  // Register module analyses
+  MAM.registerPass([&MMI] { return MachineModuleAnalysis(MMI); });
```
vladimirradosavljevic: This feels a little bit odd, to run passes inside a machine pass. Can we just do something like [here](https://github.com/matter-labs/era-compiler-llvm/blob/41693a3d04b33b51596613e07d14dc426ce0842d/llvm/lib/CodeGen/AsmPrinter/AsmPrinter.cpp#L1706) (to recalculate dominator tree and loop info for each function), since if I'm not mistaken under the hood these analysis will be recalculated before this pass.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -411,174 +427,313 @@ EVMConstantUnfolding::findOptimalTransfomtation(const APInt &Imm) {
...
+             << LoopDepth << " ***\n";
+    });
 
-      Changed |= tryUnfoldConstant(MI);
-    }
-  }
+    MinSizeAtLoopDepthOnly = LoopDepth;
+    Changed |= processModule(ModuleCodeSize, Unfolder, MinSizeAtLoopDepthOnly,
+                             M, MMI, MFAM);
+    CodeSizeReduction = Unfolder.getCodeSizeReduction();
```
vladimirradosavljevic: Can we use some other variable instead of `CodeSizeReduction`? Statistic variables are not cheap to use, since if stats are enabled they are using atomic operations under the hood.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -411,174 +427,313 @@ EVMConstantUnfolding::findOptimalTransfomtation(const APInt &Imm) {
...
+
+static bool runImpl(Module &M, ModuleAnalysisManager &AM) {
+  bool Changed = false;
+  MachineModuleInfo &MMI = AM.getResult<MachineModuleAnalysis>(M).getMMI();
+  MachineFunctionAnalysisManager &MFAM =
+      AM.getResult<MachineFunctionAnalysisManagerModuleProxy>(M).getManager();
+
+  const auto &EVMTM = static_cast<const EVMTargetMachine &>(MMI.getTarget());
+  ConstantUnfolder Unfolder(&M.getContext(),
+                            EVMTM.getSubtarget()->getInstrInfo());
```
vladimirradosavljevic: Usually in module passes, TTI is retrieved for each function like this `MF.getSubtarget<EVMSubtarget>().getInstrInfo()`.
Could we do the same?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -24,209 +24,243 @@
...
-                                   cl::desc("Gas weight used as a factor in the"
-                                            "transformation cost function"));
-
 static cl::opt<unsigned>
     LoopDepthLimit("evm-const-loop-depth-limit", cl::Hidden, cl::init(2),
                    cl::desc("The maximum loop depth at which constant"
                             "unfolding is still considered beneficial"));
 
+static cl::opt<unsigned>
+    CodeSizeLimit("evm-bytecode-sizelimit", cl::Hidden, cl::init(24576),
```
vladimirradosavljevic, line 71: This is true for runtime code, but for init code, limit is 48 KB. I think FE should set this accordingly if it is runtime or init code.

vladimirradosavljevic (dismissed):
I would suggest to do traverse over functions only once, and to collect everything you need to do transformation:
1. Move code size calculation in unfold pass, and during this, also collect PUSHes that we will transform. You can distinguish between PUSHes that will be transformed regardless of bytecode size and others if we are exceeding the limit.
2. With this approach, you won't need MLI analysis after you finish with that function.
3. In case we could end up with minsize functions during optimization for speed, you will need to distinguish between these two transformations. In case we don't, we can create a TODO.

PavelKopyl:
> I would suggest to do traverse over functions only once, and to collect everything you need to do transformation:
> 
> 1. Move code size calculation in unfold pass, and during this, also collect PUSHes that we will transform. You can distinguish between PUSHes that will be transformed regardless of bytecode size and others if we are exceeding the limit.
> 2. With this approach, you won't need MLI analysis after you finish with that function.
> 3. In case we could end up with minsize functions during optimization for speed, you will need to distinguish between these two transformations. In case we don't, we can create a TODO.

Thank you for the review!
One note regarding caching of instructions: the idea behind this check was to perform unfolding incrementally. Initially, it targets all blocks in OptForSpeed mode. Later, in OptForSize mode, I planned to expand it in stages, first for non-loop basic blocks, then for loops at depth 1, followed by loops at depth 2, and so on. Given this staged approach, maintaining a large number of caches would be both complex and potentially inefficient.

llvm/lib/Target/EVM/EVMInstrFormats.td
PavelKopyl: You are right. I allocated 3 bytes just to be prepared in case  of changes in base gas fees, but OK I'll shrink it to 2 bytes.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: Yes, this is a bit unconventional, I implemented it to cache loop analysis results, since we iterate over the machine function's basic blocks multiple times. Without caching, I would have to recompute this manually each time. It's also a step toward adopting the new pass manager for CodeGen.

vladimirradosavljevic:
> > I would suggest to do traverse over functions only once, and to collect everything you need to do transformation:
> > 
> > 1. Move code size calculation in unfold pass, and during this, also collect PUSHes that we will transform. You can distinguish between PUSHes that will be transformed regardless of bytecode size and others if we are exceeding the limit.
> > 2. With this approach, you won't need MLI analysis after you finish with that function.
> > 3. In case we could end up with minsize functions during optimization for speed, you will need to distinguish between these two transformations. In case we don't, we can create a TODO.
> 
> Thank you for the review! One note regarding caching of instructions: the idea behind this check was to perform unfolding incrementally. Initially, it targets all blocks in OptForSpeed mode. Later, in OptForSize mode, I planned to expand it in stages, first for non-loop basic blocks, then for loops at depth 1, followed by loops at depth 2, and so on. Given this staged approach, maintaining a large number of caches would be both complex and potentially inefficient.

IIRC, this PR increases compilation time, so traversing over functions only once should fix that.
Since you're already doing staging, you could maintain separate vectors for each stage and traverse them accordingly. Functionally, this isn’t much different from the current approach of manually switching stages, or I'm missing something?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
vladimirradosavljevic: We could cache analysis and reuse them when needed, if we are going to traverse over functions more than once.
Could you please explain how it will be different with the new pass manager for Module passes?

llvm/lib/Target/EVM/EVMCalculateModuleSize.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "MCTargetDesc/EVMMCTargetDesc.h"
+#include "llvm/CodeGen/MachineModuleInfo.h"
+#include "llvm/CodeGen/TargetInstrInfo.h"
+#include "llvm/CodeGen/TargetSubtargetInfo.h"
+#include "llvm/IR/Module.h"
+
+using namespace llvm;
+
+// This is the copied from AsmPrinter::isBlockOnlyReachableByFallthrough.
+static bool isBlockOnlyReachableByFallthrough(const MachineBasicBlock *MBB) {
```
akiramenai, line 25: Just to check: could't `analyzeBranch()` or `MachineBasicBlock::getFallThrough()` help here?

llvm/lib/Target/EVM/EVMCalculateModuleSize.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+    Size += std::accumulate(MBB.begin(), MBB.end(), 0,
+                            [&TII](unsigned Sum, const MachineInstr &MI) {
+                              return Sum + getInstSize(MI, TII);
+                            });
+  }
+  return Size;
+}
+
+uint64_t llvm::EVM::calculateModuleCodeSize(Module &M,
```
akiramenai, line 126: Dow we want a unit test for this?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -24,209 +24,243 @@
...
-                               "required for constant unfolding when"
-                               "optimizing for speed"));
+    SizeReductionThreshold("evm-const-unfolding-size-reduction-threshold",
+                           cl::Hidden, cl::init(2.5),
+                           cl::desc("Minimum original-to-unfolded size ratio"
+                                    "required for constant unfolding when"
+                                    "optimizing for speed"));
 
 static cl::opt<unsigned>
     InstrNumLimitUnfolInto("evm-const-unfolding-inst-num-limit", cl::Hidden,
```
akiramenai: typo: InstrNumLimitUnfol**d**Into

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -24,209 +24,243 @@
...
-    void insertInstr(MachineInstr &MI, unsigned Opcode) {
-      BuildMI(*MI.getParent(), MI, MI.getDebugLoc(), TII->get(Opcode));
-    }
+private:
+  LLVMContext *Context{};
+  const TargetInstrInfo *TII{};
+
+  // The 'second' field can be set to 0 or 1, indicating whether to
+  // optimize for performance or size.
+  using TransformationKey = std::pair<APInt, int>;
```
akiramenai, line 182: Do we need `int` if the second field can be only `0` or `1`?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -24,209 +24,243 @@
...
 
-FunctionPass *llvm::createEVMConstantUnfolding() {
+ModulePass *llvm::createEVMConstantUnfolding() {
   return new EVMConstantUnfolding();
 }
 
-const EVMConstantUnfolding::TransformationBuilder *
-EVMConstantUnfolding::findOptimalTransfomtation(const APInt &Imm) {
-  if (auto It = TransformationCache.find(Imm);
+static bool isBetterCandidate(const TransformationBuilder &A,
```
akiramenai: I'm thinking - could this be a `operator<` while renaming TransformationBuilder to TransformationCandidate? It will simplify minelement call.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
akiramenai: I think it should be module level metadata actually. A FE (think not just solx FE!) **must** specify what kind of module it is as limitations are different. Also execution profile is different too so we might want to make some alterations in optimizations as well.
If nothing is specified we can assume runtime but emit a warning. CC: @hedgar2017.

akiramenai (commented):
The overall idea look good to me. I have stylistic remarks only.
I still need to think about running passes inside a pass to suggest an alternative. It's better to be resolved.

llvm/test/CodeGen/EVM/constant-unfolding-Oz.ll
PavelKopyl: I've changed the transformation weight function to
```
static bool isBetterCandidate(const TransformationBuilder &A,
                                const TransformationBuilder &B, bool OptForSize) {
    static constexpr unsigned Weight = 2;
    const StackCostModel &CostA = A.getCost();
    const StackCostModel &CostB = B.getCost();
    unsigned ScoreA = 0;
    unsigned ScoreB = 0;
  
    if (OptForSize) {
      ScoreA = (CostA.getByteSize() * Weight) + CostA.getGas();
      ScoreB = (CostB.getByteSize() * Weight) + CostB.getGas();
    } else {
      ScoreA = CostA.getByteSize() + (Weight * CostA.getGas());
      ScoreB = CostB.getByteSize() + (Weight * CostB.getGas());
    }
    if (ScoreA != ScoreB)
      return ScoreA < ScoreB;
  
    return CostA.getInstrCount() < CostB.getInstrCount();
  }
```
It seems reasonable to consider the gas cost even in -Oz mode, since two transformations might differ by just one byte in size but have significantly different gas usage. Probably I'm wrong, so I'd like to compare the effect on Foundry projects.

PavelKopyl:
> > > I would suggest to do traverse over functions only once, and to collect everything you need to do transformation:
> > > 
> > > 1. Move code size calculation in unfold pass, and during this, also collect PUSHes that we will transform. You can distinguish between PUSHes that will be transformed regardless of bytecode size and others if we are exceeding the limit.
> > > 2. With this approach, you won't need MLI analysis after you finish with that function.
> > > 3. In case we could end up with minsize functions during optimization for speed, you will need to distinguish between these two transformations. In case we don't, we can create a TODO.
> > 
> > 
> > Thank you for the review! One note regarding caching of instructions: the idea behind this check was to perform unfolding incrementally. Initially, it targets all blocks in OptForSpeed mode. Later, in OptForSize mode, I planned to expand it in stages, first for non-loop basic blocks, then for loops at depth 1, followed by loops at depth 2, and so on. Given this staged approach, maintaining a large number of caches would be both complex and potentially inefficient.
> 
> IIRC, this PR increases compilation time, so traversing over functions only once should fix that. Since you're already doing staging, you could maintain separate vectors for each stage and traverse them accordingly. Functionally, this isn’t much different from the current approach of manually switching stages, or I'm missing something?

I agree, will do this, thank you.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: In this case, I'll need to provide a custom caching function for those keys, whereas a default implementation already exists for `int`.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: >We could cache analysis and reuse them when needed, if we are going to traverse over functions more than once.
Yeap, but this has to be done manually. 
>Could you please explain how it will be different with the new pass manager for Module passes?
In this case, it's enough to do following:
```
static bool runImpl(Module &M, ModuleAnalysisManager &AM) {
    MachineFunctionAnalysisManager &MFAM =
        AM.getResult<MachineFunctionAnalysisManagerModuleProxy>(M).getManager();
    MachineLoopInfo &MLI = MFAM.getResult<MachineLoopAnalysis>(*MF);
    ...
   }
```
and the new PM will run other analyses and cache their results. As I know, we cannot do this (i.e., access MF analyses from a Module pass level) with the legacy PM. Once we switch to the new PM, I'll just remove all the code from `runOnModule(Module &M)` that creates new PM instance, and that's it. In case of a manual caching, I'll need to do more deep clean up.

llvm/lib/Target/EVM/EVMCalculateModuleSize.cpp
PavelKopyl: I think this is the simplest implementation.
MachineBasicBlock::getFallThrough() returns the block that a given MBB falls through to, whereas here we need to solve the reverse problem.

llvm/lib/Target/EVM/EVMCalculateModuleSize.cpp
PavelKopyl: Sure, added: llvm/unittests/Target/EVM/BytecodeSize.cpp

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: I agree, but I'd make it as a separate task. For this specific pass it's enough just to know the size limit. Is that OK?

PavelKopyl:
> The overall idea look good to me. I have stylistic remarks only. I still need to think about running passes inside a pass to suggest an alternative. It's better to be resolved.

I removed the trick regarding new pass manager, so now I manually calculate MachineLoopInfo.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: I've renamed TransformationBuilder -> TransformationCandidate, yes this name looks better. But in the comparator I need third argument - OptForSize, so I don't think introducing `operator<` will simplify the code.

llvm/lib/Target/EVM/EVMCalculateModuleSize.cpp
```diff
@@ -0,0 +1,137 @@
...
+    break;
+  case EVM::PseudoCALL:
+    // In case that function call has a return label, we will emit JUMPDEST,
+    // so take it into account.
+    if (MI.getNumExplicitOperands() > 1)
+      Size = TII->get(EVM::JUMPDEST_S).getSize();
+    LLVM_FALLTHROUGH;
+  case EVM::PseudoJUMPI:
+  case EVM::PseudoJUMP:
+    // Even though we are emitting PUSH4_S, most probably this will be relaxed
```
akiramenai: // We emit PUSH4_S here. The linker usually relaxes it to PUSH2_S,
// since a 16-bit immediate covers the 24,576-byte EVM runtime code cap (EIP-170).
// If a wider immediate were ever required, the contract already exceeds the cap,
// so the push width is moot.

llvm/lib/Target/EVM/EVMTargetMachine.h
```diff
@@ -37,6 +37,8 @@ class EVMTargetMachine final : public LLVMTargetMachine {
     return &Subtarget;
   }
 
+  const EVMSubtarget *getSubtarget() const { return &Subtarget; }
```
akiramenai: Why is it needed?

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
```diff
@@ -24,209 +24,245 @@
...
+// allowing the TransformationCandidate object to be reused for repeated
+// constants. As new instructions are added, the StackCostModel is used
+// to track the accumulated cost.
+class TransformationCandidate {
+public:
+  TransformationCandidate(LLVMContext &Context, const TargetInstrInfo *TII)
+      : Context(Context), TII(TII) {}
 
-    void add() { accountInstr(EVMCOST::ADD); }
+  void addSub() {
```
akiramenai: Could you check if it's covered by regression / unit tests?
My coverage report complaints about this function.

akiramenai (approved):
LGTM

llvm/lib/Target/EVM/EVMTargetMachine.h
PavelKopyl: Thanks, missed to remove.

llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
PavelKopyl: This also needs to be removed.

PavelKopyl: Done.

llvm/lib/Target/EVM/EVMTargetMachine.h
PavelKopyl: Done.

PR #856, opened 2025-07-20
[EVM][DAGCombine] Expand SELECT to arithmetic when possible

akiramenai:
The EVM has no conditional move, so this patch replaces certain `select` nodes with straight‑line arithmetic:

```
| Source pattern  | Rewritten form   | Proof (IR equivalent)               |
|-----------------|------------------|-------------------------------------|
| select C, X, 0  | X * C            | https://alive2.llvm.org/ce/z/MqUacX |
| select C, 0, Y  | (1 − C) * Y      | https://alive2.llvm.org/ce/z/tB_ww7 |
| select C, X, 1  | X * C + (1 − C)  | https://alive2.llvm.org/ce/z/Uje2ts |
| select C, 1, Y  | C + (1 − C) * Y  | https://alive2.llvm.org/ce/z/XrBMpA |
| select C, X, ‑1 | (C - 1) | X      | https://alive2.llvm.org/ce/z/e5oNmW |
| select C, ‑1, Y | -C | Y           | https://alive2.llvm.org/ce/z/6-86BJ |
```

This patch also canonicalises

select C, (op Y, X), Y -> op Y, (select C, X, 0) -> op Y, C * X iff 0 is neutral element for op.

Alive2 proofs:
i256, sub, no undef inputs – https://alive2.llvm.org/ce/z/-BPfbO i4, sub, undef inputs allowed – https://alive2.llvm.org/ce/z/6fhhZx


PavelKopyl:
Great result, thank you! I'm curious, why wasn't these transformations handled by InstCombine?

vladimirradosavljevic:
> Great result, thank you! I'm curious, why wasn't these transformations handled by InstCombine?

My 2 cents it is just a matter of canonicalization, as I don't think you can do reverse transformation (from arithmetic to selects) and a lot of architectures have conditional move instruction, that is surely cheaper than mul. Also, during DAGCombine you have a lot of hooks that convert selects to arithmetic operations.

vladimirradosavljevic:
I saw one place where we can improve this a bit (we have iszero + iszero and we could just have iszero instead):
https://alive2.llvm.org/ce/z/KPecsZ

I would suggest to create a ticket for it, and do that later. @akiramenai  Wdyt?

akiramenai:
@vladimirradosavljevic 
Yes I think a separate ticket is the best option here.
Increasing the scope of the current PR doesn't seem right to me.

llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -758,3 +761,135 @@ MachineBasicBlock *EVMTargetLowering::emitSelect(MachineInstr &MI,
...
+  SDValue OtherOp = TrueV.getOperand(1 - OpToFold);
+  EVT OtherOpVT = OtherOp.getValueType();
+  SDValue IdentityOperand =
+      DAG.getNeutralElement(Opc, DL, OtherOpVT, N->getFlags());
+  if (!Commutative)
+    IdentityOperand = DAG.getConstant(0, DL, OtherOpVT);
+  assert(IdentityOperand && "No identity operand!");
+
+  if (Swapped)
+    std::swap(OtherOp, IdentityOperand);
```
vladimirradosavljevic, line 820: If I remove this line, all tests are passing. Please update checks.

llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -758,3 +761,135 @@ MachineBasicBlock *EVMTargetLowering::emitSelect(MachineInstr &MI,
...
+
+  EVT VT = N->getValueType(0);
+  SDLoc DL(N);
+  SDValue OtherOp = TrueV.getOperand(1 - OpToFold);
+  EVT OtherOpVT = OtherOp.getValueType();
+  SDValue IdentityOperand =
+      DAG.getNeutralElement(Opc, DL, OtherOpVT, N->getFlags());
+  if (!Commutative)
+    IdentityOperand = DAG.getConstant(0, DL, OtherOpVT);
+  assert(IdentityOperand && "No identity operand!");
```
vladimirradosavljevic:
```suggestion
  SDValue IdentityOperand = DAG.getConstant(0, DL, OtherOpVT);
```
No need to call `DAG.getNeutralElement` since 0 is neutral element for all above operations.

llvm/test/CodeGen/EVM/dag-combine-select.ll
```diff
@@ -0,0 +1,247 @@
+; NOTE: Assertions have been autogenerated by utils/update_mir_test_checks.py UTC_ARGS: --version 5
+; RUN: llc -stop-after=finalize-isel < %s | FileCheck %s
+
+target datalayout = "E-p:256:256-i256:256:256-S256-a:256:256"
+target triple = "evm"
+
+; CHECK-NOT: JUMPI
```
vladimirradosavljevic: Do we need this? We have negative tests where jump is created.

llvm/test/CodeGen/EVM/dag-combine-select.ll
```diff
@@ -0,0 +1,247 @@
...
+
+; select C, X, 0 -> X * C
+define i256 @sel_x_or_0(i1 %cond, i256 %x) {
+  ; CHECK-LABEL: name: sel_x_or_0
+  ; CHECK:       [[MUL:%[0-9]+]]:gpr = MUL
+  ; CHECK-NEXT:  RET killed [[MUL]]
+%val = select i1 %cond, i256 %x, i256 0
```
vladimirradosavljevic, line 12: Please fix indentation. Also in other tests.

llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -758,3 +761,135 @@ MachineBasicBlock *EVMTargetLowering::emitSelect(MachineInstr &MI,
   MI.eraseFromParent(); // The pseudo instruction is gone now.
   return BB;
 }
+
+// Fold
+//   (select C, (op Y, X), Y)  ->  (op Y, (select C, X, NEUTRAL)),
+//   when 0 is the neutral element for op.
+//
```
vladimirradosavljevic, line 778: This needs to be updated, since NEUTRAL element can also be in the TrueV, when Swapped is `true`.

vladimirradosavljevic (changes requested):
Could you also please add alive2 links for `tryFoldSelectIntoOp` only? Currently we have only 1 alive2 proof for final optimization.

llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -758,3 +761,135 @@ MachineBasicBlock *EVMTargetLowering::emitSelect(MachineInstr &MI,
...
+  case ISD::ADD:
+  case ISD::OR:
+  case ISD::XOR:
+    break;
+  }
+
+  if (!TrueV.hasOneUse() || isa<ConstantSDNode>(FalseV))
+    return SDValue();
+
+  unsigned OpToFold = 0;
```
vladimirradosavljevic: Maybe instead of using `OpToFold`, we can declare `OtherOp` here and set accordingly below?

vladimirradosavljevic:
> | select C, X, ‑1 | C * (X + 1) − 1  | https://alive2.llvm.org/ce/z/zMP3n4 |
| select C, ‑1, Y | −C + (1 − C) * Y | https://alive2.llvm.org/ce/z/_N56Af |

Have you tried to implement [these](https://github.com/matter-labs/era-compiler-llvm/blob/056fdd9fefeba6e29da29198a7d3ee4f1c294bf4/llvm/lib/Target/RISCV/RISCVISelLowering.cpp#L7567) transformation? It seems RISCV transformation should be more gas efficient than using mul.

llvm/test/CodeGen/EVM/dag-combine-select.ll
akiramenai: removed

llvm/lib/Target/EVM/EVMISelLowering.cpp
akiramenai: I updated, but imo too much details for a brief description. It more a distraction than help.

akiramenai:
> Could you also please add alive2 links for tryFoldSelectIntoOp only? Currently we have only 1 alive2 proof for final optimization.

We have a transformation of a kind A->B->C.
We've proven A->C aka final optimization, we've proven B->C aka select -> mul. Proving A->B looks redundant to me.

vladimirradosavljevic (approved):
LGTM. Thanks!
Could you please rebase this PR, as it should fix the failing tests.

llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -856,7 +856,7 @@ SDValue EVMTargetLowering::combineSELECT(SDNode *N,
 
   // fold (Cond ? X : -1) -> (Cond - 1) | X*)
   if (isAllOnesConstant(FalseV)) {
-    const SDValue Const1 = DAG.getConstant(1, DL, VT);
+    const SDValue Const1 = DAG.getConstant(APInt(256, 1, false), DL, VT);
```
vladimirradosavljevic: Why is this needed? VT should be 256bit, so this looks like it is not needed.

PR #858, opened 2025-07-22
[EVM] Don't emit JUMPDEST in entry point of bytecode

vladimirradosavljevic:
We don't need to emit JUMPDEST instruction in the first BB of the __entry function.


PavelKopyl (approved):
LGTM

PR #859, opened 2025-07-22
ci: add forge benchmarks support in llvm

antonbaliasnikov:
# Code Review Checklist

## Purpose

Add forge benchmarks support with each LLVM commit.

### DIFF
--- a/.github/workflows/forge-benchmarks.yaml
+++ b/.github/workflows/forge-benchmarks.yaml
@@ -1,0 +1,24 @@
+name: Forge benchmarks
+
+on:
+  workflow_dispatch:
+    inputs:
+      solx-version:
+        description: 'Prebuilt `solx` version to use. For example: `0.1.0` without `v` prefix. If empty, `solx` will be built.'
+        required: false
+        default: ''
+      llvm-ref:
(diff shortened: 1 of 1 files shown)


hedgar2017:
Hey @antonbaliasnikov, what is it with the LNT workflow?

hedgar2017 (changes requested):
I believe we should be publishing raw *.xslx files without archives, so it's easy to open, especially on mobile devices.
Please let me know if it's difficult to do now, so we'll finish it when my tool is ready.

antonbaliasnikov:
> I believe we should be publishing raw *.xslx files without archives, so it's easy to open, especially on mobile devices. Please let me know if it's difficult to do now, so we'll finish it when my tool is ready.

@hedgar2017 , unfortunately it's not possible with GitHub Actions. It's always .zip.

This is also the problem why we can't publish HTML reports just to open them in a new tabs, etc.
It's a stupid github actions restriction.

antonbaliasnikov:
> Hey @antonbaliasnikov, what is it with the LNT workflow?

Fixed today in CI repo, it should pass now.

hedgar2017 (approved):

PR #861, opened 2025-07-23
[EVM][DAGCombine] Expand SELECT of constants to arithmetic

vladimirradosavljevic:


llvm/lib/Target/EVM/EVMISelLowering.cpp
```diff
@@ -876,6 +880,37 @@ SDValue EVMTargetLowering::combineSELECT(SDNode *N,
     return DAG.getNode(ISD::OR, DL, VT, Neg, DAG.getFreeze(FalseV));
   }
 
+  // fold (Cond ? X : C) -> (C + (X* - C) * Cond)
+  // fold (Cond ? C : Y) -> (Y* + (C - Y*) * Cond)
+  //
```
akiramenai: The logic is hard to follow the way it implemented, I recommend to split handling of
```
  // fold (Cond ? X : C) -> (C + (X* - C) * Cond)
  // fold (Cond ? C : Y) -> (Y* + (C - Y*) * Cond)
``` 
from
```
 // fold (Cond ? C1 : C2) -> (C2 + (C1 - C2) * Cond) if (C1 > C2)
 // fold (Cond ? C1 : C2) -> (C2 - (C2 - C1) * Cond) if (C1 < C2)
```

akiramenai (approved):

llvm/lib/Target/EVM/EVMISelLowering.cpp
vladimirradosavljevic: Done. PTAL.

PavelKopyl (approved):
LGTM.

PR #863, opened 2025-07-23
[EVM] Remove EVMModuleLayout pass and use attribute to detect entry function

vladimirradosavljevic:
After FE changes, entry function has to be the first in the module, so `EVMModuleLayout` is no longer needed. Instead, add an assert in `EVMAsmPrinter` that entry function has to be the first function in the module.
Also, instead of checking function name to detect entry function, check for `"evm-entry-function"` attribute.


PavelKopyl:
I wonder why do we have any size/perf changes?

vladimirradosavljevic:
> I wonder why do we have any size/perf changes?

No idea. Maybe era-compiler-tester is not updated with the latest changes?

hedgar2017 (changes requested):
Phantom perf changes fixed!

@vladimirradosavljevic could you make the PR name more descriptive?

hedgar2017 (approved):

PavelKopyl (approved):
LGTM.

PR #867, opened 2025-07-29
[EVM][InlineCost] Set threshold to a small value when BB is unreachable-terminated

vladimirradosavljevic:
For EVM, to return from the contract we are using different instructions (e.g. return, revert) which are followed by unreachable. In the Inliner heuristic, if BB is unreachable-terminated threshold is set to 0 and these callsites are unlikely to be inlined. Instead, add small threshold and continue to calculate cost model, since in some cases we benefit from inlining.


akiramenai (changes requested):
Could we add a positive and a negative test case for inlining with unreachable?

akiramenai (approved):

PR #871, opened 2025-07-30
[EVM][DAGCombine] Legalize brcond (i1 anything)

akiramenai:


PavelKopyl:
I've analyzed several failures and I’m not sure this approach works in the general case. To insert an `any_extend` before type legalization and DAG optimizations, we’d need to analyze the entire instruction tree used to build the condition for `brcond` and ensure that only LBS bit is used. The problem, as I see it now, is that such analysis won’t remain valid after type legalization (i1 -> i256). ISel will optimize the DAG differently depending on whether `zext` or `anyext` is used, effectively changing the semantics of the original code.
The only viable option is to perform combining right before the instruction selection stage itself, that is, when `and i256 1, %x` has already been inserted, and we’re trying to remove it. However, this patch doesn’t yield any improvements in compiler-tester, with the only observed case being when a function has an argument of type `i1`. It seems to me that in all other cases the DAG either doesn’t insert an and for the branch condition, or those ands are actually required.

Here is an example when AND is redundant:
```
define void @dont_trunc_or_cond(i256 %a, i1 %conda) {
  %cond = icmp ne i256 %a, 0
  %condb = or i1 %cond, %conda
  br i1 %condb, label %true, label %false
false:
  unreachable
true:
  unreachable
}
```

```
dont_trunc_or_cond:                     ; @dont_trunc_or_cond
; %bb.0:
	JUMPDEST
	PUSH1           0x1
	SWAP3
	POP
	ISZERO
	ISZERO
	OR
	AND
	PUSH4           @.BB0_2
	JUMPI
; %bb.1:                                ; %false
.BB0_2:                                 ; %true
	JUMPDEST
                                        ; -- End function
	INVALID
```

This AND can be eliminated, for example, by following change in ISel:

```
SDValue EVMTargetLowering::combineBRCOND(SDNode *N,
                                         DAGCombinerInfo &DCI) const {
  // Perform combines only after DAG legalisation.
  if (!DCI.isAfterLegalizeDAG())
    return SDValue();

  SDLoc DL(N);
  SelectionDAG &DAG = DCI.DAG;
  assert(N->getOpcode() == ISD::BRCOND);
  SDValue Chain = N->getOperand(0);
  SDValue Cond = N->getOperand(1);
  SDValue Dest = N->getOperand(2);

  SDNode *CondNode = Cond.getNode();
  if (CondNode->getOpcode() != ISD::AND)
    return SDValue();

  if (!isa<ConstantSDNode>(CondNode->getOperand(0)) &&
      !isa<ConstantSDNode>(CondNode->getOperand(1)))
    return SDValue();

  unsigned AndConstOpIdx = isa<ConstantSDNode>(CondNode->getOperand(0)) ? 0 : 1;
  if (auto *C = cast<ConstantSDNode>(CondNode->getOperand(AndConstOpIdx));
      !C->getAPIntValue().isOne())
    return SDValue();

  SDValue AndNonConstOp = CondNode->getOperand(!AndConstOpIdx);
  SmallVector<SDNode *, 64> Worklist;
  Worklist.push_back(AndNonConstOp.getNode());

  const Function &F = DAG.getMachineFunction().getFunction();
  while (!Worklist.empty()) {
    SDNode *Node = Worklist.pop_back_val();
    unsigned Opcode = Node->getOpcode();

    if (Opcode == ISD::SETCC)
      continue;

    if (Opcode == EVMISD::ARGUMENT) {
      unsigned ArgIdx = Node->getConstantOperandVal(0);
      Type *ArgTy = F.getArg(ArgIdx)->getType();
      if (ArgTy->isIntegerTy() && ArgTy->getIntegerBitWidth() == 1)
        continue;
    }

    if (Opcode == ISD::AND) {
      for (const SDValue &Op : Node->op_values())
        if (auto *C = dyn_cast<ConstantSDNode>(Op); C->getAPIntValue().isOne())
          continue;
    }

    if (Opcode == ISD::Constant)
      if (!cast<ConstantSDNode>(Node)->getAPIntValue().isOne())
        continue;

    if (Opcode != ISD::AND && Opcode != ISD::OR && Opcode != ISD::XOR)
      return SDValue();

    for (const SDValue &Op : Node->op_values())
      Worklist.push_back(Op.getNode());
  }

  // Rebuild BRCOND as BRCOND.chain, Ext:i256, Dest
  return DAG.getNode(ISD::BRCOND, DL, MVT::Other, Chain, AndNonConstOp, Dest);
}
```

PR #872, opened 2025-07-30
[EVM][ISel] Select (setcc x, 0, ne) as ISZERO ISZERO

akiramenai:
Before the patch it was selected as `PUSH0 EQ ISZERO`.


vladimirradosavljevic (approved):
LGTM.

PR #873, opened 2025-07-31
[EVM] Fold signextend(b, signextend(b, x)) -> signextend(b, x)

vladimirradosavljevic:


akiramenai (changes requested):
Tests?

vladimirradosavljevic:
> Tests?

PR is now ready for the review. PTAL.

akiramenai (approved):
LGTM, let's ship it and then investigate if conversion to standard LLVM IR instructions is a better approach.

PR #902, opened 2025-09-22
[EVM] Set that staticcall can't change storage

vladimirradosavljevic:
From the description, staticcall can't change storage, so update the code to ensure no storage modifications occur during staticcall.

### DIFF
--- a/llvm/include/llvm/IR/IntrinsicsEVM.td
+++ b/llvm/include/llvm/IR/IntrinsicsEVM.td
@@ -108,6 +108,10 @@ def int_evm_msize : Intrinsic<[llvm_i256_ty], [], [IntrWillReturn]>;
 
 def int_evm_pc : Intrinsic<[llvm_i256_ty], [], [IntrWillReturn]>;
 
+// TODO: #904: Set IntrInaccessibleMemOnly attribute to gas intrinsic.
+// Currently, it is not set because benchmark numbers showed that it is not
+// profitable to do that for heap address space. Revisit this decision once
+// optimization for free ptr updates is implemented.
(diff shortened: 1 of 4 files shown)


PavelKopyl (approved):
LGTM, thanks!

akiramenai:
Can we improve `EVMTargetLowering::isLegalAddressingMode` coverage?

akiramenai:
Also, https://github.com/matter-labs/era-compiler-llvm/pull/899 seems to be a part of this PR. Should we close it or would you prefer 899 to be merged first?

vladimirradosavljevic:
@akiramenai Sorry for the confusion, but including https://github.com/matter-labs/era-compiler-llvm/pull/899 here was just an experiment to check whether degradations will be fixed when I set that gas doesn't read or write to heap address space. 
Since that wasn't the case, I disabled it for heap address space, and we can revisit that decision in a future when free ptr updates optimization is implemented.
PTAL now.

llvm/include/llvm/IR/IntrinsicsEVM.td
```diff
@@ -108,6 +108,10 @@ def int_evm_msize : Intrinsic<[llvm_i256_ty], [], [IntrWillReturn]>;
 
 def int_evm_pc : Intrinsic<[llvm_i256_ty], [], [IntrWillReturn]>;
 
+// TODO: Set IntrInaccessibleMemOnly attribute to gas intrinsic. Currently, it
```
akiramenai: Could we add a ticket and mark all relevant TODO with its number?

akiramenai (approved):

PR #903, opened 2025-09-24
[EVM] Enable Tail Duplication after stackification

vladimirradosavljevic:


akiramenai:
> Some of them can cause functional issues if duplicated (e.g. PUSH_LABEL) and some of them can increase code size (e.g. PUSH instructions).

Aren't any better solution to address size issues? For some instructions we also have Size property, can't we infer `isTailDuplicable` from it? Are there ways to reason about a BB as a whole in terms of size rather than on per-instruction basis?

vladimirradosavljevic:
> > Some of them can cause functional issues if duplicated (e.g. PUSH_LABEL) and some of them can increase code size (e.g. PUSH instructions).
> 
> Aren't any better solution to address size issues? For some instructions we also have Size property, can't we infer `isTailDuplicable` from it? Are there ways to reason about a BB as a whole in terms of size rather than on per-instruction basis?

Tail duplication implementation is based on instruction count, not on instruction size. This was the cleanest solution I found to tweak that pass, and not to introduce local changes.
Regarding Size property, that might work, but the only concern I have with that approach is that I would add `isNotDuplicable` to Size > 3, but that also includes `PseudoJUMPI` and `PseudoJUMP`, which we want to include in optimization, so I would leave it as is.

akiramenai:
Ok, then can we refactor PUSHN before adding isNotDuplicable. An AI suggests:
```
multiclass PushN<int Bytes, bits<8> Opc, string BaseName> {
  defvar ImmBits = !mul(Bytes, 8);
  defvar EncBits = !add(ImmBits, 8);

  def PUSH#Bytes#_S
    : PushBase<(outs), (ins i256imm:$imm), [],
               /*hasSideEffects*/ true,
               !strconcat(BaseName, "$imm"),
               Opc, 3> {

    bits<ImmBits> imm;
    bits<EncBits> Inst;

    let Inst{!add(ImmBits,7) - ImmBits} = Opc;
    let Inst{ImmBits-1 - 0}            = imm;

    let Size = !add(1, Bytes);
    let BaseName = BaseName;

    // Non-duplicable for PUSH3+ (Bytes >= 3)
    let isNotDuplicable = !ge(Bytes, 3);
    // or: let isNotDuplicable = !gt(Bytes, 2);
    // or from Size: let isNotDuplicable = !ge(Size, 4);
  }
}

defm : PushN<1, 0x60, "PUSH1">;
defm : PushN<2, 0x61, "PUSH2">;
defm : PushN<3, 0x62, "PUSH3">;
defm : PushN<4, 0x63, "PUSH4">;
...
```

vladimirradosavljevic:
> Ok, then can we refactor PUSHN before adding isNotDuplicable. An AI suggests:
> 
> ```
> multiclass PushN<int Bytes, bits<8> Opc, string BaseName> {
>   defvar ImmBits = !mul(Bytes, 8);
>   defvar EncBits = !add(ImmBits, 8);
> 
>   def PUSH#Bytes#_S
>     : PushBase<(outs), (ins i256imm:$imm), [],
>                /*hasSideEffects*/ true,
>                !strconcat(BaseName, "$imm"),
>                Opc, 3> {
> 
>     bits<ImmBits> imm;
>     bits<EncBits> Inst;
> 
>     let Inst{!add(ImmBits,7) - ImmBits} = Opc;
>     let Inst{ImmBits-1 - 0}            = imm;
> 
>     let Size = !add(1, Bytes);
>     let BaseName = BaseName;
> 
>     // Non-duplicable for PUSH3+ (Bytes >= 3)
>     let isNotDuplicable = !ge(Bytes, 3);
>     // or: let isNotDuplicable = !gt(Bytes, 2);
>     // or from Size: let isNotDuplicable = !ge(Size, 4);
>   }
> }
> 
> defm : PushN<1, 0x60, "PUSH1">;
> defm : PushN<2, 0x61, "PUSH2">;
> defm : PushN<3, 0x62, "PUSH3">;
> defm : PushN<4, 0x63, "PUSH4">;
> ...
> ```

This doesn't work, as `bits` and `Inst` only accept integer values, and you can't pass variables 😢
Only solution I see here is to generate custom encoding method, so we don't use `bits` and `Inst`, but that looks like an overkill to me, since we have common code that is working well.

PavelKopyl:
>This doesn't work, as bits and Inst only accept integer values, and you can't pass variables 😢
Yeah, I also spent quite a bit of time trying to work around this odd limitation. I’d say it’s a strange TableGen restriction that would be nice to fix. Once resolved, we could compress the PUSH1–PUSH32 declarations into a loop.

akiramenai (approved):
@vladimirradosavljevic  could you create a ticket then to address TableGen constraints in LLVM mainline? Given the uncertainty, it might be a good idea to fix it.

PR #914, opened 2025-10-08
[EVM] Clear the branch condition in TII::removeBranch for stackified code

PavelKopyl:
# Code Review Checklist

## Requirements

## Implementation

## Logic Errors and Bugs
that could break the code?

## Error Handling and Logging
be added or removed?
debugging?

## Maintainability

## Dependencies

## Security

## Performance
system performance?

## Testing and Testability
- [ ] Is the code testable?
	- [ ] Line Coverage
	- [ ] Region Coverage
	- [ ] Branch Coverage

## Readability
smaller methods?
file/folder/package?
flow?
better?
understandable?

## Documentation

## Best Practices
- [ ] Magic values avoided?
- [ ] Minimal nesting used?

## Experts' Opinion

### DIFF
--- a/llvm/lib/Target/EVM/EVMInstrInfo.cpp
+++ b/llvm/lib/Target/EVM/EVMInstrInfo.cpp
@@ -173,10 +173,18 @@ unsigned EVMInstrInfo::removeBranch(MachineBasicBlock &MBB,
   LLVM_DEBUG(dbgs() << "Removing branches out of " << printMBBReference(MBB)
                     << '\n');
 
-  unsigned Count = 0;
   // Only remove branches from the end of the MBB.
-  for (auto I = MBB.rbegin(); I != MBB.rend() && I->isBranch(); ++Count) {
+  unsigned Count = 0;
+  bool HasConditionalBranch = false;
+  MachineBasicBlock::iterator I = MBB.end();
+  while (I != MBB.begin()) {
+    --I;
+    if (I->isDebugInstr())
+      continue;
+    if (!I->isTerminator())
+      break;
 
+    HasConditionalBranch |= I->isConditionalBranch();
 #ifndef NDEBUG
     if (I->isUnconditionalBranch())
       assert(!Count && "Malformed basic block: unconditional branch is not the "
@@ -184,9 +192,39 @@ unsigned EVMInstrInfo::removeBranch(MachineBasicBlock &MBB,
 #endif // NDEBUG
 
     LLVM_DEBUG(dbgs() << "Removing branch: "; I->dump());
-    MBB.erase(&MBB.back());
-    I = MBB.rbegin();
+    I->eraseFromParent();
+    I = MBB.end();
+    ++Count;
   }
+
+  // Let's consider basic blocks that end with
+  //
+  //   JUMPI
+  //   JUMP
+  //
+  // or
+  //
+  //  JUMPI
+  //
+  // When removeBranch() is called, it removes all jump instructions.
+  // In stackified code, this can disrupt the stack layout by leaving an unused
+  // condition on top of the stack, a condition that would normally be consumed
+  // by the JUMPI instruction.
+  // To handle this, we add a POP_COND_S instruction at the end of the BB after
+  // removing the conditional branch, ensuring that the unused condition is
+  // properly consumed:
+  // In most cases, removeBranch() is followed by insertBranch(), which
+  // typically reinserts a JUMPI instruction, possibly with a different target.
+  // In such cases, the condition will again be consumed by the new JUMPI, so
+  // the POP_COND_S is no longer needed and is removed. Although this adds a
+  // small overhead, by creating and then removing POP_COND_S, the cost is
+  // negligible since it only affects stackified code.
+  //
+  const bool IsStackified =
+      MBB.getParent()->getInfo<EVMMachineFunctionInfo>()->getIsStackified();
+  if (IsStackified && HasConditionalBranch)
+    BuildMI(&MBB, DebugLoc(), get(EVM::POP_COND_S));
+
   return Count;
 }
 
@@ -214,6 +252,14 @@ unsigned EVMInstrInfo::insertBranch(MachineBasicBlock &MBB,
       CondOpc = Cond[0].getImm() ? EVM::JUMPI : EVM::JUMP_UNLESS;
     else
       CondOpc = Cond[0].getImm() ? EVM::PseudoJUMPI : EVM::PseudoJUMP_UNLESS;
+
+    // Remove the POP_COND_S instruction previously added by removeBranch,
+    // as the newly inserted conditional branch will now consume that
+    // condition.
+    auto Last = MBB.getLastNonDebugInstr();
+    if (IsStackified && Last != MBB.end() &&
+        Last->getOpcode() == EVM::POP_COND_S)
+      MBB.erase(*Last);
 
     auto NewMI = BuildMI(&MBB, DL, get(CondOpc)).addMBB(TBB);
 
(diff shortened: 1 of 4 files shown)


llvm/lib/Target/EVM/EVMInstrInfo.cpp
```diff
@@ -187,6 +188,17 @@ unsigned EVMInstrInfo::removeBranch(MachineBasicBlock &MBB,
...
+  const bool IsStackified =
+      MBB.getParent()->getInfo<EVMMachineFunctionInfo>()->getIsStackified();
+  if (IsStackified && HasConditionalBranch) {
+    auto NewMI = BuildMI(&MBB, DebugLoc(), get(EVM::POP_S));
+    // Mark this POP instruction with a special flag to distinguish it from other
+    // POP instruction in insertBranch that is not related to branch-condition
+    // clearing. It’s unlikely that such a case exists; this is mostly for
+    // paranoia check.
+    NewMI->setFlag(MachineInstr::MIFlag::ClearBranchCondition);
+    --Count;
```
vladimirradosavljevic: Do we need this?
Description of removeBranch says it returns the number of instructions that were removed.

llvm/test/CodeGen/EVM/branch-folding-clear-condition.ll
```diff
@@ -0,0 +1,47 @@
+; NOTE: Assertions have been autogenerated by utils/update_llc_test_checks.py UTC_ARGS: --version 2
```
vladimirradosavljevic: I think this should be a MIR test that is calling only branchfolding pass. Could you please convert it?

vladimirradosavljevic (changes requested):
Since we are removing EraVM target, we will remove big chunk of local changes, so I would recommend not to introduce new one, as we have other options (e.g. introduce new `POP_COND_S`).

llvm/lib/Target/EVM/EVMInstrInfo.cpp
PavelKopyl: Yeap, but we actually insert 1-byte instruction here, do you meant we shouldn't take it into account?

vladimirradosavljevic: I was looking at callers of `removeBranch` and in some cases they do something based on the return (e.g. `PeelSingleBlockLoop` in `MachineLoopUtils.cpp`). Currently in case we have only conditional branch and we add `POP`, we will return 0, but actually we need to return 1.

vladimirradosavljevic (approved):
LGTM. Thanks!

llvm/lib/Target/EVM/EVMInstrInfo.cpp
```diff
@@ -187,6 +188,11 @@ unsigned EVMInstrInfo::removeBranch(MachineBasicBlock &MBB,
     MBB.erase(&MBB.back());
     I = MBB.rbegin();
   }
+  const bool IsStackified =
+      MBB.getParent()->getInfo<EVMMachineFunctionInfo>()->getIsStackified();
+  if (IsStackified && HasConditionalBranch)
```
akiramenai, line 225: I think a part of commit message doc is to be put here.

llvm/lib/Target/EVM/EVMInstrInfo.cpp
```diff
@@ -174,9 +174,10 @@ unsigned EVMInstrInfo::removeBranch(MachineBasicBlock &MBB,
                     << '\n');
 
   unsigned Count = 0;
+  bool HasConditionalBranch = false;
   // Only remove branches from the end of the MBB.
   for (auto I = MBB.rbegin(); I != MBB.rend() && I->isBranch(); ++Count) {
```
akiramenai: getLastNonDebugInstr?

llvm/lib/Target/EVM/EVMInstrInfo.cpp
```diff
@@ -215,6 +221,13 @@ unsigned EVMInstrInfo::insertBranch(MachineBasicBlock &MBB,
     else
       CondOpc = Cond[0].getImm() ? EVM::PseudoJUMPI : EVM::PseudoJUMP_UNLESS;
 
+    // Remove the POP_COND_S instruction that was previously added in
+    // removeBranch to clear the jump condition from the stack.
```
akiramenai: Could you explain why?

akiramenai (changes requested):
Documentation:
1. Could you put some explanations to logic in code?
2. Can we document the rationale for the design choice in commit message?

llvm/lib/Target/EVM/EVMInstrInfo.cpp
PavelKopyl: Let's consider basic blocks (BBs) that end with
```
JUMPI
JUMP
```
or
```JUMPI```

When `removeBranch() `is called, it removes all jump instructions.
In stackified code, this can disrupt the stack layout by leaving an unused condition on top of the stack, i.e., condition that would normally be consumed by the `JUMPI` instruction.
To handle this, we add a `POP_COND_S` instruction at the end of the BB after removing the conditional branch, ensuring that the unused condition is properly consumed.
In most cases (I would say in 99%), `removeBranch()` is followed by `insertBranch()`, which typically reinserts a `JUMPI` instruction, possibly with a different target.
In such cases, the condition will again be consumed by the new `JUMPI`, so the `POP_COND_S` is no longer needed and is removed.
Although this adds a small overhead, by creating and then removing `POP_COND_S`, the cost is negligible since this only happens in stackified code.
In the rare cases where ` removeBranch()` is called without a subsequent insertBranch(), the `POP_COND_S` remains and will be emitted as a `POP` instruction.
For example, this happens here:
https://github.com/matter-labs/era-compiler-llvm/blob/8e7767fdb8a19cb687c78827907c10a757a78dc5/llvm/lib/CodeGen/BranchFolding.cpp#L1469

File: llvm/lib/Target/EVM/EVMInstrInfo.cpp (hunk 4 above)
Comment (PavelKopyl) [reply llvm/lib/Target/EVM/EVMInstrInfo.cpp]: We may have a BB ending with
```
JUMPI
JUMP 
```
so with `getLastNonDebugInstr` we will miss the presence of JUMPI.

File: llvm/lib/Target/EVM/EVMInstrInfo.cpp (hunk 3 above)
Comment (PavelKopyl) [reply llvm/lib/Target/EVM/EVMInstrInfo.cpp:225]: OK, added.

**Comment by PavelKopyl** [comment]
> Documentation:
> 
> 1. Could you put some explanations to logic in code?
> 2. Can we document the rationale for the design choice in commit message?

Sure, I've added this to the commit message.

File: llvm/lib/Target/EVM/EVMInstrInfo.cpp (hunk 4 above)
Comment (akiramenai) [reply llvm/lib/Target/EVM/EVMInstrInfo.cpp]: I mean with the loop we could hit a debug instruction and bail out.

**Review by akiramenai — APPROVED** [review]

File: llvm/lib/Target/EVM/EVMInstrInfo.cpp (hunk 4 above)
Comment (PavelKopyl) [reply llvm/lib/Target/EVM/EVMInstrInfo.cpp]: Thanks, got it. I've slightly rewrote the removeBranch to skip debug instructions.

File: llvm/lib/Target/EVM/EVMInstrInfo.cpp [hunk 6]
```diff
@@ -173,20 +173,58 @@ unsigned EVMInstrInfo::removeBranch(MachineBasicBlock &MBB,
...
-  for (auto I = MBB.rbegin(); I != MBB.rend() && I->isBranch(); ++Count) {
-
+  unsigned Count = 0;
+  bool HasConditionalBranch = false;
+  MachineBasicBlock::iterator I = MBB.end();
+  while (I != MBB.begin()) {
+    --I;
+    if (I->isDebugInstr())
+      continue;
+    if (!I->isTerminator())
```
vladimirradosavljevic, line 184: This should be `!I->isBranch()`, right?
Terminators are also RET, REVERT, RETURN, STOP and INVALID, and we don't want to remove them.

PavelKopyl: Right, thank you. Pushed a fix: https://github.com/matter-labs/era-compiler-llvm/pull/923

Issue #918, opened 2025-10-13
[EVM] Improve generation of CTZ instruction


vladimirradosavljevic:
CTZ can be implemented via CLZ, and LLVM is generating:
`256 - clz(~x & (x-1))`
If `x != 0` is proven, this can be optimized to:
`255 - clz(x & -x)`

During ISel, `SelectionDAG::isKnownNeverZero` can be used to check this.

PR #919, opened 2025-10-13
[EVM] Add support for CLZ instruction

vladimirradosavljevic:
CLZ will be introduced in the fusaka update:
https://eips.ethereum.org/EIPS/eip-7939

Beside adding support for CLZ instruction in this
patch, support for EVM versioning is added, so CLZ
instruction is generated for fusaka and later versions.
isCheapToSpeculateCttz and isCheapToSpeculateCtlz
are implemented so CGP doesn't generate check for zero
to avoid calling the intrinsic in despeculateCountZeros.
CTZ is generated via CLZ, and currently LLVM is
generating:
`256 - clz(~x & (x-1))`
If x != 0 is proven, this can be optimized to:
`255 - clz(x & -x)` (https://github.com/matter-labs/era-compiler-llvm/issues/918)


PavelKopyl (approved):
LGTM.

akiramenai (approved):

PR #926, opened 2025-10-30
[EVM] Fix handling of imprecise memory locations in EVMAliasAnalysis

vladimirradosavljevic:
This patch fixes following issues:
1. If both sizes are imprecise, return MayAlias instead of PartialAlias.
2. If both pointers have the same starting address and one of them has
   imprecise size, return MayAlias instead of PartialAlias. In this case
   we can't deduce between MustAlias or PartialAlias, so be conservative
   and return MayAlias.
3. If imprecise is lower than precise memory location, return MayAlias
   instead of PartialAlias. Imprecise memory location can be NoAlias or
   PartialAlias, but we don't know that in the compile time.

The only case where we can return PartialAlias is when imprecise memory
location is in the middle of precise location. This is not done since
there shouldn't be any optimization that could benefit from this, and
to keep the implementation simple.


PavelKopyl (approved):
LGTM, thank you!

PR #927, opened 2025-10-31
[EVM] Fix nondeterminism in constant unfolding

vladimirradosavljevic:
Previously, the code that collected PUSH instructions to process used a DenseMap with MachineFunction* as the key. This could lead to nondeterministic iteration order, causing inconsistent behavior across different runs. Fix this by changing DenseMap to SmallVector, since we are using lookups for asserts that can't fail, because we are iterating over all MachineFunctions only once.

### DIFF
--- a/llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
+++ b/llvm/lib/Target/EVM/EVMConstantUnfolding.cpp
@@ -635,7 +635,8 @@ static bool runImpl(Module &M, MachineModuleInfo &MMI) {
       EVM::calculateModuleCodeSize(M, MMI) + MetadataSize;
 
   // Collect PUSH instructions to process.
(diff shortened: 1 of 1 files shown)


PavelKopyl (approved):
LGTM

vladimirradosavljevic:
@PavelKopyl Could you please check the logic one more time, since I changed `DenseMap` to `SmallVector`, as I don't think we need a Map here.

PavelKopyl:
> @PavelKopyl Could you please check the logic one more time, since I changed `DenseMap` to `SmallVector`, as I don't think we need a Map here.

@vladimirradosavljevic Good point, the map isn’t necessary. A SmallVector of std::pair should be sufficient. It seems the map was unintentionally left over from previous versions of the patch.

PR #929, opened 2025-11-18
[EVM] Print large immediates in hex in disassembly

vladimirradosavljevic:
Large immediates were previously printed in decimal. Print them in hex instead.


PavelKopyl (approved):
LGTM

akiramenai (approved):

PR #933, opened 2025-12-02
[EVM] Force compress stack when there is no reg to spill

vladimirradosavljevic:
In cases where we are out of registers to spill, try to force compress stack to fix stack too deep errors. This can help if there are unreachable slots that only contain literals and other rematerializable slots, but no registers.


PavelKopyl (approved):
LGTM

akiramenai (approved):

PR #935, opened 2026-04-06
[EraVM] Disable NewGVN and GVNHoist passes by default

vladimirradosavljevic:
This patch disables NewGVN pass as in some cases can cause miscompilations. It also disables GVNHoist, as that pass is not turned on by default in the main line.


PavelKopyl (approved):
LGTM

PR #939, opened 2026-09-18
Fix BytesToCells for select stack and code inputs

akiramenai:
EraVMBytesToCells failed to convert stack and code-page inputs of
    selects that write their result to the stack. Select pseudos are not
    represented in the standard input addressing-mode maps, so these inputs
    are identified by explicit opcode tables. The tables only covered
    register-result forms, leaving the stack and code-page inputs of
    stack-result selects unconverted.

    Extend select address conversion to cover stack-result forms and all
    stack/code-page operands. Select pseudos are not covered by the generic
    addressing-mode machinery because select supports a non-register second
    input operand.

    Also restrict the code-address early-clobber handling to selects with a
    register output, since stack-result selects have no output register.

    Co-authored-by: Antonio Locascio <alo@matterlabs.dev>


AntonD3 (approved):

AntonD3:
Changes merged in https://github.com/matter-labs/era-compiler-llvm/pull/941

